PR #4, opened 2020-04-27
Minor English language improvements in docs

manning-ncsa:
These are trivial changes that only polish the existing content, which is already completely understandable and well-written. These updates make the text slightly more natural to a native English speaker.


cppforlife:
much appreciated!

PR #17, opened 2020-07-08
Support for password authentication when using git fetch

Haegi:
This PR adds username/password auth for accessing git repos over https.
Therefore the referenced secret needs to contain a `username` and `password` value.

### DIFF
--- a/pkg/fetch/git.go
+++ b/pkg/fetch/git.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@Haegi Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=k14s/kapp-controller&pullRequestId=17)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=k14s/kapp-controller&pullRequestId=17) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@Haegi Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=k14s/kapp-controller&pullRequestId=17)!

cppforlife:
i wonder if we can somehow pass username/password over stdin or env vars instead of being part of url

phil9909:
Do you have security concerns? The controller will have a service-account token [mounted as a file](https://github.com/kubernetes/client-go/blob/master/rest/config.go#L473), which allows to fetch the secret containing the password anyway.
I don't think using stdin or env-variables will increase the security here.

cppforlife:
> I don't think using stdin or env-variables will increase the security here.

typically stdin or envvars are preferred to cmdline args since args are visible to other users in the system afaik.

voor:
I'd recommend using the existing git credential helper if possible, this would achieve keeping the username and password away from environment variables or command line arguments.

https://stackoverflow.com/questions/35942754/how-to-save-username-and-password-in-git-gitextension

phil9909:
We switched to writing a temp-file with the credentials and set the `credential.helper` git config to use that file. The password will no longer be visible in the command / git-url.

pkg/fetch/git.go
```diff
@@ -80,9 +81,30 @@ func (t *Git) Retrieve(dstPath string) error {
...
+		err = ioutil.WriteFile(gitCredentialsPath, []byte(parsedUrl.String()+"\n"), 0600)
+		if err != nil {
+			return fmt.Errorf("writing .git-credentials: %s", err)
+		}
+	}
+
 	argss := [][]string{
 		{"init"},
-		{"remote", "add", "origin", t.opts.URL},
+		{"config", "credential.helper", "store --file " + gitCredentialsPath},
```
cppforlife, line 106: just to confirm if this command runs in parallel with others git "setup" for other apps, it should be ok because it's scoped to the git directory vs being global?

phil9909: Yes, otherwise the `--global` flag would be needed.

PR #27, opened 2020-08-12
Integrate SOPS as a templating function

Arttii:
Hi,

I was wondering how to add a few different types of templating functions to the kapp-controller, specifically I was looking at adding a [sops](https://github.com/mozilla/sops) templating function to decrypt files. While doing that I wondered why not have a templating function that just pipes your files through any arbitrary command. Most of the templating functions essentially do that anyway. 

I added a WIP version of this functionality and an example to illustrate the two usecases.I still need to add test for this as well.

 Any thoughts or comments are appreciated. If this does not fit the vision and is silly, please let me know.


examples/simple-app-command.yml
```diff
@@ -0,0 +1,49 @@
...
+  namespace: default
+spec:
+  serviceAccountName: default-ns-sa
+  fetch:
+    - image:
+        subPath: monitoring
+        url: arttii/dummydeploy
+  template:
+  - command: 
+      cmd: "sops"
```
loewenstein: Where is the binary coming from? Should this allow (or require) specifying the image to run? Would we implement the other templating logic (ytt, helm, jsonnet,...) using this approach?

Arttii: It should be present on the system the controller in running on, so in the Docker. So if you want to use custom commands like this, you need to build a custom Docker.
I would not implement all templating with this, I think it's better to have a more specific implementation for functions, that we know users need. This one is more or less for the case, when one is not available or does not make a lot of sense it integrate as an inbuilt templating function.

cari-lynn:
Thanks for the contribution, @Arttii!

Definitely, we would like to be able to support SOPS in `kapp-controller`.

We can totally see the thinking leading towards an arbitrary command. However, one of our goals with this tool is that it be relatively safe to use. Invoking the principle of least privilege.

So your initial tack — writing a templating adapter for SOPS — is on-point.

Thinking through what's needed, each step does just enough to ultimately support that deployment.
As you've been integrating with SOPS, what features do you need for your workflow?


As far as the `sed` case, we're wondering what kinds of edits you might be performing there. Is
it possible to achieve the desired result with another templating language?

cari-lynn:
To summarize my last comment, it would be an excellent addition to have a PR for a SOPS specific templater. We appreciate all the work you have put into this so far and are looking forward to that PR.

Arttii:
Thanks for the feedback. It makes sense. For SOPS our usecase is to create pregenerated yaml specs and push them as artifacts to a docker registry.  The tricky thing here are the secrets that would end up unencrypted in the artifact. Idea would be preencrypt them and then decrypt via SOPS on template. 

I do not see another usecase of it here really, but I guess we can make it quite generic, with one being able to pass arbitrary extra args to sops, and that command just being applied on all matching files. Same way its being done in this PR. Does this make sense?

cari-lynn:
Would you be open to changing this PR to support only SOPS as a first class templating feature? 

We want to keep the api surface small and only support a narrow use case so that we can reduce the possibility of security risks. Opening up the api to be more generic is something that has come up in the past, see [this issue](https://github.com/k14s/kapp-controller/issues/13) on using a docker image as a template engine. It’s something we are interested in as a potential future state. 

For this feature, supporting SOPS decryption with the flags necessary for your use case is where we would like to start. In the future as more use cases emerge, then we can build upon this implementation.

Please let us know if you have any thoughts on this.

Arttii:
Yes, we need this feature ourselves anyway, so I will work on the rework. Thanks for the input

Arttii:
I added a very simple sops implementation. This will decrypt by default as I don't know if it there is a use case to encrypt at all. I went for the lazy approach and did not make explicit args. I think ill add the following args explicitely as a next step, if the general design is acceptable:
```

   --kms value, -k value                    comma separated list of KMS ARNs [$SOPS_KMS_ARN]
   --aws-profile value                      The AWS profile to use for requests to AWS
   --gcp-kms value                          comma separated list of GCP KMS resource IDs [$SOPS_GCP_KMS_IDS]
   --azure-kv value                         comma separated list of Azure Key Vault URLs [$SOPS_AZURE_KEYVAULT_URLS]
   --pgp value, -p value  
   --input-type value                       currently json, yaml, dotenv and binary are supported. If not set, sops will use the file's extension to determine the type
   --output-type value                      currently json, yaml, dotenv and binary are supported. If not set, sops will use the input file's extension to determine the output format
  --ignore-mac
 

```

thoughts?

pivotaljohn:
Looking good, @Arttii.

Thanks for being flexible with your approach.

Yes: spelling out the parameters we'll pass to `sops` as fields on `AppTemplateSops` provides a way to allow-list only the required features/inputs; all in the name of limiting the surface area of the invocation. 👍🏼 

Thinking further on this part, given that we're deploying Kubernetes resources, seems safe to assume/hardcode `input-type` and `output-type` are both `yaml`... can you think of a reason to allow that to be configured within this integration? 🤔 

Other than that, LGTM.

Arttii:
Hey @pivotaljohn 
I added the parameters explicitly and also removed the input-output handling. We can add that later if anybody needs it.

pkg/apis/kappctrl/v1alpha1/types_template.go
```diff
@@ -21,6 +22,21 @@ type AppTemplateYtt struct {
...
+	AzureKV    []string `json:"azureKV,omitempty"`
+	PGP        []string `json:"pgp,omitempty"`
+	AWSProifle string   `json:"awsProfile,omitempty"`
+	IgnoreMac  bool     `json:"ignoreMac,omitempty"`
+}
+
+type AppTemplateSops struct {
+	Match      string               `json:"match,omitempty"`
+	MergeFiles bool                 `json:"mergeFiles,omitempty"`
+	Args       *AppTemplateSopsArgs `json:"args,omitempty"`
```
cppforlife: i think we should inline various AppTemplateSopsArgs fields into AppTemplateSops.

pkg/apis/kappctrl/v1alpha1/types_template.go
```diff
@@ -21,6 +22,21 @@ type AppTemplateYtt struct {
...
+	GCPKms     []string `json:"gcpKms,omitempty"`
+	AzureKV    []string `json:"azureKV,omitempty"`
+	PGP        []string `json:"pgp,omitempty"`
+	AWSProifle string   `json:"awsProfile,omitempty"`
+	IgnoreMac  bool     `json:"ignoreMac,omitempty"`
+}
+
+type AppTemplateSops struct {
+	Match      string               `json:"match,omitempty"`
+	MergeFiles bool                 `json:"mergeFiles,omitempty"`
```
cppforlife: do we need to have Match and MergeFiles functionality? can we simplify this integration to having sops just rewrite files on disk that are encrypted by it? i m guessing sops "templater" will always be used as a first templater so we probably shouldnt worry about merging of files.

cppforlife:
@Arttii apologies for the delay. ive left two comments on the files.

Arttii:
I have added the suggested changes. Please take a look. I am unsure why overwriting the files in place is a better approach in your opinion as the Deploy function still expects a string, so the overwritten file needs to be read again. Or I am missing something?

cppforlife:
> I am unsure why overwriting the files in place is a better approach in your opinion as the Deploy function still expects a string

oops, i was a bit ahead of where the code is. what i was imagining is that sops would change necessary files on fs, and subsequent step would pick them up (instead of expecting is as a stream). more specifically i was thinking about cases like sops + ytt where second step (ytt) needs fs structure to distinguish certain properties of given files (like what it is, etc.)

let me think for a sec how to best deal with that.

cppforlife:
it seems that we need a mechanism that somehow indicates to kapp-controller that sops will keep the directory structure instead of sending stream to subsequent templaters. it's probably best to hard code that here (https://github.com/k14s/kapp-controller/blob/develop/pkg/app/app_template.go#L23-L46) for now that sops will always do templatedir and subsequent templaters should also do template dir.

Arttii:
@cppforlife did you have a think on how you want to proceed with this?

cppforlife:
change to support multiple directory processing ended up pretty simple: https://github.com/k14s/kapp-controller/commit/dac29df1d3273d43f56dd265c1755a4cc9f42660

i've actually went ahead and implemented basic sops integration for PGP here: https://github.com/k14s/kapp-controller/commit/2588faa6b68d2be5359af31eabd05efaeb517cc1, while playing around with sops to understand how to best integrate.

cppforlife:
PGP support included in v0.11.0+: https://github.com/k14s/kapp-controller/releases/tag/v0.11.0. let's continue working together on bringing other sops strategies.

i was chatting with @voor about how would AWS KMS look like here. it would be ideal if somehow user would be able to specify how to auth to amazon. typically it's done thru some kind of Pod specific binding (or service accounts), but are there other ways of doing it? ideally we would be able to use app's service account as a way to connect to IaaS specific provider APIs.

Arttii:
Generally we do not have to do much, the aws ask handles most scenarios for doing this automatically. We just need to init it [properly](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html) and have a config to set the region and key id/alias

cppforlife:
> the aws ask handles most scenarios for doing this automatically

will have to look into this a bit closer but i have a feeling that the default setup is "pod-centric" vs being able to use particular service account per App.

cppforlife:
since we only are left with KMS specific functionality of this PR, ill close it in favor of discussion in https://github.com/k14s/kapp-controller/issues/37.

PR #45, opened 2020-11-13
Update Hardcoded Repo Name in Workflow

danielhelfand:
Updating checkout step to use `${{ github.repository }}` for its path to help with transitioning to new org. Should be aliased, but might be nice to avoid confusion if someone is reviewing CI process.


danielhelfand:
Possibly related to Helm 2 being unsupported for test failure? https://helm.sh/blog/helm-2-becomes-unsupported/

PR #58, opened 2020-12-10
Label automation

joaopapereira:
Automatically adds the label `carvel-triage` to issues that are reopened and to issues that are closed but a new comment is added.
Automatically marks the issue with a `stale` label 40 days of inactivity and will close that issue after 5 more days


cari-lynn:
Question for packaging team @ewrenn8 - do we want to triage issues separately? If do we want to optimize for easy searching by using two separate labels, `carvel-triage` and `carvel-packaging-triage`?

PR #59, opened 2020-12-10
Vendir fetch

ewrenn8:
Let me know if we should sqaush before merging


pkg/apis/kappctrl/v1alpha1/types.go
```diff
@@ -76,6 +76,7 @@ type AppStatus struct {
 
 type AppStatusFetch struct {
 	Stderr    string      `json:"stderr,omitempty"`
+	Stdout    string      `json:"stdout,omitempty"`
```
cppforlife, line 79: this might get pretty wordy especially for output vendir produces. i wonder if we need a separate issue to truncate it or something?

PR #63, opened 2020-12-18
Configurable timeout for http

achimweigel:
We had the situation that all threads of the reconciler hang in http requests. I looks as if there is no timeout set for the corresponding clients. With this commit it is possible to configure an http timeout. Nothing changes if no timeout is configured.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -8,6 +8,7 @@ package main
 import (
 	"flag"
 	"os"
+	"time"
 
 	// Pprof related
 	"net/http"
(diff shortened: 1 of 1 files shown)


cppforlife:
squashed and merged manually since merge conflicts were too great: https://github.com/vmware-tanzu/carvel-kapp-controller/commit/34cc94900ad44b5cadb05d5d428e79353ab32db2

cppforlife:
flag is `--api-request-timeout=<golang duration format>`

PR #65, opened 2021-01-05
Add configuration for http(s) proxy

ewrenn8:


cmd/controller/run.go
```diff
@@ -44,6 +45,14 @@ func Run(opts Options, runLog logr.Logger) {
 		os.Exit(1)
 	}
 
+	if proxyVal := os.Getenv("http_proxy"); proxyVal != "" {
```
cppforlife: should deal with no_proxy as well

pkg/config/global/configurer.go
```diff
@@ -81,6 +89,16 @@ func (gc *GlobalConfigurer) addTrustedCerts(certChain string) (err error) {
 	return file.Close()
 }
 
+func (gc *GlobalConfigurer) configureProxies(httpProxy, httpsProxy string) {
+	if httpProxy != "" {
```
cppforlife, line 96: + no_proxy

should we be setting also uppercase versions?

PR #72, opened 2021-01-21
Adding golint and goheader linters via golangci-lint

danielhelfand:
Adding go-lint/goheader linters to CI. Also fixing violations flagged by go-lint/goheader. To review those changes, see commit [`ae955d4`](https://github.com/vmware-tanzu/carvel-kapp-controller/commit/ae955d46bc526c165098f8e20dbf965f0908ffc7).

### DIFF
--- a/hack/build-and-lint.sh
+++ b/hack/build-and-lint.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+set -e -x -u
+
+./hack/build.sh
+./hack/linter.sh
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,18 @@
+name: golangci-lint
+
+on:
+  push:
+  pull_request:
+    types: [opened, reopened]
+
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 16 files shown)


hack/build-and-lint.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -e -x -u
+
+./hack/build.sh
+./hack/linter.sh
```
ewrenn8, line 6: I wonder if this should be added to test-all.sh. I rarely build because the test scripts have a build in them, so will likely miss any linting errors until they hit github.

danielhelfand: The intention in doing so is to keep the test-all script separate in CI. It's nice to run the linting/build and test in parallel and report results separately. 

One thing that we could do is have a script that runs everything (i.e., check.sh) that runs to determine whether or not everything is ready to be pushed up. Then the test all script can be run on its own in CI.

Not strongly opinionated though. Am happy to just load everything into one CI job if that's easier too.

ewrenn8: Alright, that makes sense. I dont think there is a need for another script yet. We can add one in the future if develop starts to get a lot of commits that just fix linting errors.

PR #76, opened 2021-01-27
WIP: Add support for fetching imgpkg bundles

alexbrand:
This PR introduces the ability to fetch imgpkg bundles when installing an App.

I have marked it as WIP because:

* It requires an un-released version of vendir. Once a new version of vendir is available, the Dockerfile must be updated to point to the new carvel-vendir release.
* It is missing an e2e test

To put together a testable container image, I created a custom build of vendir and copied it into the kapp-controller container:
1. Clone the vendir repo
2. Build a vendir linux binary
3. Copy the binary into the kapp-controller repo
3. Modify the dockerfile to include the vendir binary in the container
4. profit!


pkg/apis/kappctrl/v1alpha1/types_fetch.go
```diff
@@ -86,3 +87,14 @@ type AppFetchLocalRef struct {
 	// Object is expected to be within same namespace
 	corev1.LocalObjectReference `json:",inline" protobuf:"bytes,1,opt,name=localObjectReference"`
 }
+
+type AppFetchImgpkgBundle struct {
+	Image string `json:"image,omitempty"`
```
danielhelfand, line 92: Might be worthwhile looking into `imgpkg`'s Bundle definition here:

https://github.com/vmware-tanzu/carvel-imgpkg/blob/76031cd3e5dd4366b3d3163a8b24fcfc1f4c8f94/pkg/imgpkg/bundle/bundle.go

danielhelfand: A bit new to the project here and wasn't completely aware about using go modules versus calling the binary directly in this case. Disregard my comment.

cppforlife:
https://github.com/vmware-tanzu/carvel-vendir/releases/tag/v0.16.0 shipped.

danielhelfand:
@alexbrand Would you like to make final changes, or would you like us to finish the last bits of the pr?

danielhelfand:
Added in #93

PR #78, opened 2021-01-27
only inspect if deployment attempted

### LINKED ISSUE #77: Don't run kapp inspect if we haven't successfully deployed
**What steps did you take:**
1. Deploy an app that has a failing fetch stage:
```yaml
apiVersion: kappctrl.k14s.io/v1alpha1
kind: App
metadata:
  name: simple-app
  namespace: default
spec:
  serviceAccountName: default-ns-sa
  fetch:
  - http:
      url: i-dont-exist
  template:
  - ytt: {}
  deploy:
  - kapp: {}
```
2. `kubectl get apps -n <namespace you deployed to> -oyaml`
3. see that status has an inspect stage showing an error

**What happened:**
Seeing the inspect error, which isnt the actual error causing reconciliation to fail, is misleading
```yaml
    fetch:
      error: 'Fetching resources: exit status 1'
      exitCode: 1
      startedAt: "2021-01-27T19:33:30Z"
      stderr: |
        Error: Syncing directory '0': Syncing directory '.' with HTTP contents: Downloading URL: Initiating URL download: Get i-dont-exist: unsupported protocol scheme ""
      updatedAt: "2021-01-27T19:33:30Z"
    friendlyDescription: 'Reconcile failed: Fetching resources: exit status 1'
    inspect:
      error: 'Inspecting: exit status 1'
      exitCode: 1
      stderr: 'kapp: Error: App ''simple-app-ctrl'' (namespace: default) does not
        exist: configmaps "simple-app-ctrl" not found'
      stdout: Target cluster 'https://10.96.0.1:443'
      updatedAt: "2021-01-27T19:33:30Z"
```

**What did you expect:**
To just see the error that relates to why the app is failing (in this case the fetch)

We should not run inspect if the app has never been deployed.

danielhelfand:
Closes #77

This pull requests adds skipping the inspect reconciliation in the event an app deployment is never attempted. Since the deployment is not attempted, adding inspect to the app status might distract users from the true error they are experiencing.

### DIFF
--- a/pkg/app/app_reconcile.go
+++ b/pkg/app/app_reconcile.go
@@ -81,7 +81,10 @@ func (a *App) reconcileDeploy() error {
 	a.setReconcileCompleted(result)
 
 	// Reconcile inspect regardless of deploy success
-	_ = a.reconcileInspect()
+	// but don't inspect if deploy never attempted
+	if a.app.Status.Deploy != nil {
+		_ = a.reconcileInspect()
+	}
 
 	return a.updateStatus("marking reconcile completed")
 }
--- a/test/e2e/fetch_test.go
+++ b/test/e2e/fetch_test.go
@@ -1,0 +1,92 @@
+// Copyright 2021 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package e2e
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/ghodss/yaml"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func Test_NoKappInspect_IfNoDeployAttempted_AfterFetchFailure(t *testing.T) {
+	env := BuildEnv(t)
+	logger := Logger{}
+	kapp := Kapp{t, env.Namespace, logger}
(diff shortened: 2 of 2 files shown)


pkg/app/app_reconcile.go
```diff
@@ -80,8 +80,11 @@ func (a *App) reconcileDeploy() error {
 	result := a.reconcileFetchTemplateDeploy()
 	a.setReconcileCompleted(result)
 
-	// Reconcile inspect regardless of deploy success
-	_ = a.reconcileInspect()
+	deployAttempted := a.Status().Deploy == nil
+	// Reconcile inspect regardless of deploy success but don't inspect if deploy never attempted
+	if !deployAttempted {
```
cppforlife: other methods in this class use `a.app.Status...` so for consistency we should switch to `a.app.Status.Deploy`.

`a.app.Status.Deploy != nil` is descriptive enough imho for the conditional so var can be inlined.

test/e2e/fetch_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
```
cppforlife: nit: extra newline

pkg/app/app_reconcile.go
danielhelfand: Addressed

test/e2e/fetch_test.go
danielhelfand: Addressed

PR #105, opened 2021-02-19
Use Carvel Setup Action

danielhelfand:
Helps to align versions of tools used in kapp-controller image with versions of tools used in non e2e tests. This also helps to dogfood our own GitHub Action.

### DIFF
--- a/.github/workflows/test-gh.yml
+++ b/.github/workflows/test-gh.yml
@@ -23,6 +23,16 @@ jobs:
       uses: actions/checkout@v1
       with:
         path: src/github.com/${{ github.repository }}
+    - name: Install Carvel Tools
+      uses: vmware-tanzu/carvel-setup-action@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        only: ytt, kapp, kbld, imgpkg, vendir
+        ytt: v0.31.0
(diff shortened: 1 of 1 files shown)


.github/workflows/test-gh.yml
```diff
@@ -23,6 +23,16 @@ jobs:
       uses: actions/checkout@v1
       with:
         path: src/github.com/${{ github.repository }}
+    - name: Install Carvel Tools
+      uses: vmware-tanzu/carvel-setup-action@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        only: ytt, kapp, kbld, imgpkg, vendir
```
cppforlife, line 30: dont think we need imgpkg or vendir as binaries.

danielhelfand: There is a non e2e test for a failed fetch that relies on vendir being installed: https://github.com/vmware-tanzu/carvel-kapp-controller/blob/53cbfee531155c40268ff8914f62becd15a94836/pkg/app/app_reconcile_test.go#L21

I could also see imgpkg being used at some point in our testing, but I can remove for now as it's not currently used.

cppforlife: ah sounds good.

PR #127, opened 2021-03-05
Add App Reoncile Upon Update of Secrets/ConfigMaps

### LINKED ISSUE #26: watch for changes to configmap/secret and trigger App reconcile
this would make syncPeriod configuration more useful (e.g. it could be set higher, and App can try to react to "input" changes proactively). i'd like to have this mechanims pave the way for "watching" of other resources used by App.

danielhelfand:
Closes #26 

This pull request adds immediate app reconciliation when associated ConfigMaps/Secrets are updated. This helps to assure Apps are always using the most up to date version of ConfigMaps and Secrets.

This pull request adds two watchers to kapp-controller (one for each configmaps and secrets) and associated handlers that enqueue app reconcile requests upon updates to configmaps/secrets. 

It also adds an `AppRefTracker` struct which uses maps that allow communication between the App reconciler and handlers to know when to force an App to be reconciled in the event of an update to configmaps/secrets.


test/e2e/git_test.go
```diff
@@ -82,8 +82,8 @@ spec:
 			Template: &v1alpha1.AppStatusTemplate{
 				ExitCode: 0,
 			},
-			ConsecutiveReconcileSuccesses: 1,
-			ObservedGeneration:            1,
+			ConsecutiveReconcileSuccesses: 2,
```
danielhelfand: Mainly highlighting how behavior has changed with updates. This behavior should not change for this feature.

cmd/controller/apps_reconciler.go
```diff
@@ -35,5 +43,53 @@ func (r *AppsReconciler) Reconcile(request reconcile.Request) (reconcile.Result,
...
+	result, err := crdApp.Reconcile(force)
+	if err != nil {
+		return result, err
+	}
+
+	// Only update if reconcile is successful.
+	// Leave as not updated if error occurs
+	// in case follow up reconcile request may
+	// address issue for App.
+	r.appRefTracker.MarkAppUpdated(existingApp.Name, existingApp.Namespace)
```
cppforlife: dont think we need this. if things have failed then app will be re-enqueued because of that and Reconcile() will try to reconcile as well.

cppforlife: > dont think we need this.

let me rephrase, i dont think we need to do this only on success. i think it's fine doing it always.

cmd/controller/apps_reconciler.go
```diff
@@ -35,5 +43,53 @@ func (r *AppsReconciler) Reconcile(request reconcile.Request) (reconcile.Result,
...
+
+	return result, err
+}
+
+func (r *AppsReconciler) areAppRefsUpToDate(refNames map[string]struct{}, kind string, app *v1alpha1.App) bool {
+	// No updates if paused or cancelled
+	if app.Spec.Canceled || app.Spec.Paused {
```
cppforlife: areAppRefsUpToDate shouldnt concern itself if app is paused since that's something that Reconcile() accounts for that.

cmd/controller/run.go
```diff
@@ -73,12 +77,14 @@ func Run(opts Options, runLog logr.Logger) {
 	}
 
 	{ // add controller for apps
+		appRefTracker := reftracker.NewAppRefTracker()
```
cppforlife, line 79: instead of using & in following uses of this var, let's have NewAppRefTracker return a pointer to AppRefTracker

pkg/reftracker/ref_tracker.go
```diff
@@ -0,0 +1,91 @@
...
+
+var appExists struct{}
+
+type AppRefTracker struct {
+	refsToApps         map[string]map[string]struct{}
+	appsToUpdateStatus map[string]struct{}
+}
+
+func NewAppRefTracker() AppRefTracker {
+	refsToApps := make(map[string]map[string]struct{})
```
cppforlife: `make(map[string]map[string]struct{})` -> `map[string]map[string]struct{}{}`

test/e2e/app_secret_configmap_reconcile_test.go
```diff
@@ -0,0 +1,256 @@
...
+			t.Fatalf("Failed to unmarshal: %s", err)
+		}
+
+		// check for successful deployment
+		if cr.Status.Deploy == nil || !cr.Status.Deploy.Finished || cr.Status.Deploy.ExitCode != 0 {
+			t.Fatalf("Expected %s deployment to succeed but got:\n%s", name, cr.Status.Deploy.Stdout)
+		}
+	})
+
+	logger.Section("update secret", func() {
```
cppforlife: update configmap?

test/e2e/app_secret_configmap_reconcile_test.go
```diff
@@ -0,0 +1,256 @@
...
+			out := kubectl.Run([]string{"get", "configmap/configmap", "-o", "yaml"})
+
+			var cm corev1.ConfigMap
+			err := yaml.Unmarshal([]byte(out), &cm)
+			if err != nil {
+				return fmt.Errorf("failed to unmarshal: %s", err)
+			}
+
+			if cm.Data["hello_msg"] != "updated" {
+				return fmt.Errorf("\nconfigmap message was not updated to \"updated\"\nGot: %s", cm.Data["hello_msg"])
```
cppforlife: seems a bit weird to prefix error msg with newline.

cmd/controller/apps_reconciler.go
```diff
@@ -5,16 +5,24 @@ package controller
...
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
 	kcclient "github.com/vmware-tanzu/carvel-kapp-controller/pkg/client/clientset/versioned"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/reftracker"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	secret    = "secret"
```
cppforlife: seems weird to have these as constants.

cmd/controller/apps_reconciler.go
```diff
@@ -35,5 +43,53 @@ func (r *AppsReconciler) Reconcile(request reconcile.Request) (reconcile.Result,
...
+	}
+
+	// Make sure refs for App are always up to date
+	// in appRefTracker.
+	for refName := range refNames {
+		if !r.appRefTracker.CheckAppExistsForRef(kind, refName, app.Namespace, app.Name) {
+			r.appRefTracker.AddAppToRefMap(kind, refName, app.Namespace, app.Name)
+		}
+	}
+
```
cppforlife: this method is called areAppRefsUpToDate(). to me that reads that this method just checks on something. in highlighted code, it's doing some modification which i think is surprising for a checking method. is there a better place to put it (at the least separate function).

pkg/reftracker/ref_tracker.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+var appExists struct{}
+
+type AppRefTracker struct {
+	refsToApps         map[string]map[string]struct{}
+	appsToUpdateStatus map[string]struct{}
```
cppforlife: seems that this object is doing two fairly distinct things: keeping track of app refs and also keeping track if app should be forced to update. let's split them up into two separate objects.

pkg/reftracker/ref_tracker.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func (a AppRefTracker) GetAppUpdateStatus(appName, namespace string) bool {
+	appKey := strings.ToLower(fmt.Sprintf(`%s:%s`, appName, namespace))
+	_, keyExists := a.appsToUpdateStatus[appKey]
+	return keyExists
+}
+
+func (a AppRefTracker) RemoveAppFromUpdateMap(appName, namespace string) {
+	appKey := strings.ToLower(fmt.Sprintf(`%s:%s`, appName, namespace))
```
cppforlife: lots of ToLower going on. why do we need it?

pkg/reftracker/ref_tracker.go
```diff
@@ -0,0 +1,91 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package reftracker
+
+import (
+	"fmt"
+	"strings"
+)
+
+var appExists struct{}
```
cppforlife: since it's conventional to use `struct{}` as a sentinel value it's a bit weird to see as variable. in places where we use it i would recommend to do `struct{}{}`. (less indirection).

cmd/controller/apps_reconciler.go
danielhelfand: Intention here was to make sure consistent kind is being used for keys to avoid using strings directly. But can remove and use regular strings instead.

cmd/controller/apps_reconciler.go
danielhelfand: Sure, there can be an `updateAppRefs()` method that runs every time with these aspects of updating the AppRefTracker and then the check method of `appRefTracker.GetAppUpdateStatus(app.Name, app.Namespace)` can be used directly to determine whether to set force.

pkg/reftracker/ref_tracker.go
```diff
@@ -0,0 +1,91 @@
...
+
+	appKey := strings.ToLower(appName)
+	delete(apps, appKey)
+	refKey := strings.ToLower(fmt.Sprintf(`%s:%s:%s`, resourceKind, resourceName, namespace))
+	a.refsToApps[refKey] = apps
+
+	return nil
+}
+
+func (a AppRefTracker) MarkAppForUpdate(appName, namespace string) {
```
cppforlife: MarkAppForUpdate -> MarkNeedsUpdate
MarkAppUpdated -> MarkUpdated
GetAppUpdateStatus -> IsUpdateNeeded
(no need for RemoveAppFromUpdateMap, MarkUpdated does necessary things already).

pkg/reftracker/ref_tracker.go
```diff
@@ -0,0 +1,91 @@
...
+package reftracker
+
+import (
+	"fmt"
+	"strings"
+)
+
+var appExists struct{}
+
+type AppRefTracker struct {
```
cppforlife, line 11: we got maps that are accessed in concurrent so need locks.

pkg/reftracker/ref_tracker.go
danielhelfand: I am wondering if it actually would make sense to just have a single map and change `refToApps` to `map[string]map[string]bool`. The main reason for using `appsToUpdateStatus` was for supporting the following that is going to be removed:

```
// Leave as not updated if error occurs
// in case follow up reconcile request may
// address issue for App.
r.appRefTracker.MarkAppUpdated(existingApp.Name, existingApp.Namespace)
```

pkg/reftracker/ref_tracker.go
danielhelfand: Standardize the keys in case someone using this function makes mistake of how key is formatted.

pkg/app/crd_app_test.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	k8scs := k8sfake.NewSimpleClientset()
+	kappcs := fake.NewSimpleClientset()
+	fetchFac := fetch.NewFactory(k8scs)
+	tmpFac := template.NewFactory(k8scs, fetchFac)
+	deployFac := deploy.NewFactory(k8scs)
+
+	crdApp := NewCRDApp(&app, log, kappcs, fetchFac, tmpFac, deployFac)
+	_, err := crdApp.Reconcile(false)
```
cppforlife: why do we need to reconcile here?

danielhelfand: We don't. Copy paste error from another test.

cmd/controller/run.go
danielhelfand: Added

pkg/reftracker/ref_tracker.go
danielhelfand: Added lock

pkg/reftracker/ref_tracker.go
danielhelfand: Changed

pkg/reftracker/ref_tracker.go
danielhelfand: Added

pkg/app/crd_app_test.go
danielhelfand: Removed

pkg/reftracker/ref_tracker.go
danielhelfand: Ended up going with your advice to split to separate struct.

cmd/controller/apps_reconciler.go
danielhelfand: Updated func name and have it solely focus on updating app refs.

cmd/controller/apps_reconciler.go
danielhelfand: Removed

cmd/controller/apps_reconciler.go
danielhelfand: Removed and now always marking app has been updated.

pkg/app/crd_app.go
```diff
@@ -119,10 +119,100 @@ func (a *CRDApp) updateApp(updateFunc func(*kcv1alpha1.App)) error {
...
+func (a *CRDApp) Reconcile(force bool) (reconcile.Result, error) {
+	return a.app.Reconcile(force)
 }
 
 func (a *CRDApp) watchChanges(callback func(kcv1alpha1.App), cancelCh chan struct{}) error {
 	return NewCRDAppWatcher(*a.appModel, a.appClient).Watch(callback, cancelCh)
 }
+
+// Get all SecretRefs from App spec
+func (a *CRDApp) GetSecretRefs() map[string]struct{} {
```
cppforlife: we can probably push this down to App CR, but still expose it on CRDApp. tests could go against App CR's methods.

danielhelfand: Moved to App

cmd/controller/apps_reconciler.go
cppforlife: it seems like it's only being used in one place.

cmd/controller/run.go
```diff
@@ -10,6 +10,10 @@ import (
 	"os"
 	"time"
 
+	"github.com/vmware-tanzu/carvel-kapp-controller/cmd/controller/handlers"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/reftracker"
+	v1 "k8s.io/api/core/v1"
+
```
cppforlife: extra new line

pkg/app/app_test.go
```diff
@@ -0,0 +1,153 @@
+// Copyright 2021 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package app
```
cppforlife: this should be in app_test package.

pkg/reftracker/ref_tracker.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+type AppRefTracker struct {
+	lock       sync.Mutex
+	refsToApps map[string]map[string]struct{}
+}
+
+func NewAppRefTracker() *AppRefTracker {
+	refsToApps := map[string]map[string]struct{}{}
+	lock := sync.Mutex{}
```
cppforlife: dont need to initialize it.

pkg/reftracker/ref_tracker.go
```diff
@@ -0,0 +1,127 @@
...
+	return &AppRefTracker{refsToApps: refsToApps, lock: lock}
```
cppforlife: can inline refsToApps

cmd/controller/apps_reconciler.go
danielhelfand: Yeah, I could move it somewhere more central, but maybe let's just remove for now and use strings.

pkg/reftracker/ref_tracker.go
cppforlife: to me this seems trying to "fix" problems that should be fixed elsewhere. since we control all invocations this seems to be too defensive.

pkg/reftracker/ref_tracker.go
```diff
@@ -0,0 +1,127 @@
...
+		return fmt.Errorf("could not find App ref %s/%s", resourceKind, resourceName)
+	}
+
+	appKey := strings.ToLower(appName)
+	delete(apps, appKey)
+	a.refsToApps[refKey] = apps
+	return nil
+}
+
+type AppUpdateStatus struct {
```
cppforlife: let's split this off into a separate file.

cmd/controller/apps_reconciler.go
```diff
@@ -35,5 +39,31 @@ func (r *AppsReconciler) Reconcile(request reconcile.Request) (reconcile.Result,
 		return reconcile.Result{}, err
 	}
 
-	return r.appFactory.NewCRDApp(existingApp, log).Reconcile()
+	force := false
+	crdApp := r.appFactory.NewCRDApp(existingApp, log)
+	r.updateAppRefs(crdApp.GetSecretRefs(), "secret", existingApp)
```
cppforlife: nit: newline before.

cmd/controller/apps_reconciler.go
```diff
@@ -35,5 +39,31 @@ func (r *AppsReconciler) Reconcile(request reconcile.Request) (reconcile.Result,
...
+	r.updateAppRefs(crdApp.GetConfigMapRefs(), "configmap", existingApp)
+	if r.appUpdateStatus.IsUpdateNeeded(existingApp.Name, existingApp.Namespace) {
```
cppforlife: nit: newline before

cmd/controller/apps_reconciler.go
```diff
@@ -35,5 +39,31 @@ func (r *AppsReconciler) Reconcile(request reconcile.Request) (reconcile.Result,
 		return reconcile.Result{}, err
 	}
 
-	return r.appFactory.NewCRDApp(existingApp, log).Reconcile()
+	force := false
```
cppforlife: nit: i would move variable as close as possible to its use. so right before if.

cmd/controller/apps_reconciler.go
```diff
@@ -35,5 +39,31 @@ func (r *AppsReconciler) Reconcile(request reconcile.Request) (reconcile.Result,
...
+	if app.DeletionTimestamp != nil {
+		for refName := range refNames {
+			r.appRefTracker.RemoveAppFromRefMap(kind, refName, app.Namespace, app.Name)
+		}
+	}
+
+	// Make sure refs for App are always up to date
+	// in appRefTracker.
+	for refName := range refNames {
+		if !r.appRefTracker.CheckAppExistsForRef(kind, refName, app.Namespace, app.Name) {
```
cppforlife: let's not bother checking. just always mark it.

cmd/controller/apps_reconciler.go
```diff
@@ -35,5 +39,31 @@ func (r *AppsReconciler) Reconcile(request reconcile.Request) (reconcile.Result,
...
+	// If App is being deleted, remove App from appRefTracker.
+	if app.DeletionTimestamp != nil {
+		for refName := range refNames {
+			r.appRefTracker.RemoveAppFromRefMap(kind, refName, app.Namespace, app.Name)
+		}
+	}
+
+	// Make sure refs for App are always up to date
+	// in appRefTracker.
+	for refName := range refNames {
```
cppforlife: this i guess doesnt track when refs are deleted. not sure if we need to deal with that...

pkg/reftracker/ref_tracker.go
```diff
@@ -0,0 +1,127 @@
...
+	refsToApps map[string]map[string]struct{}
+}
+
+func NewAppRefTracker() *AppRefTracker {
+	refsToApps := map[string]map[string]struct{}{}
+	lock := sync.Mutex{}
+	return &AppRefTracker{refsToApps: refsToApps, lock: lock}
+}
+
+func (a AppRefTracker) AddAppToRefMap(resourceKind, resourceName, namespace, appName string) {
```
cppforlife: AddAppToRefMap -> AddAppForRef
GetAppsForRef -> AppsForRef
RemoveRefFromMap -> RemoveRef
RemoveAppFromRefMap -> RemoveApp
CheckAppExistsForRef (remove)

test/e2e/app_secret_configmap_reconcile_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+func Test_AppReconcileOccurs_WhenSecretUpdated(t *testing.T) {
+	env := BuildEnv(t)
+	logger := Logger{}
+	kapp := Kapp{t, env.Namespace, logger}
+	kubectl := Kubectl{t, env.Namespace, logger}
+	sas := ServiceAccounts{env.Namespace}
+
+	name := "configmap-with-secret"
+	// syncPeriod set to 5 minutes so that test
```
cppforlife: do you want to crank it up to something high like 60min?

pkg/app/app.go
```diff
@@ -114,3 +114,93 @@ func (a *App) newCancelCh() (chan struct{}, func()) {
...
+			}
+		default:
+		}
+	}
+
+	return secrets
+}
+
+// Get all ConfigMapRefs from App spec
+func (a *App) getConfigMapRefs() map[string]struct{} {
```
cppforlife: let's make getConfigMapRefs and getSecretRefs public and test them directly instead of going thru CRDApp. also rename getSecretRefs -> SecretRefs, ...

cmd/controller/apps_reconciler.go
```diff
@@ -35,5 +39,31 @@ func (r *AppsReconciler) Reconcile(request reconcile.Request) (reconcile.Result,
...
+	}
+
+	return crdApp.Reconcile(force)
+}
+
+func (r *AppsReconciler) updateAppRefs(refNames map[string]struct{}, kind string, app *v1alpha1.App) {
+	// If App is being deleted, remove App from appRefTracker.
+	if app.DeletionTimestamp != nil {
+		for refName := range refNames {
+			r.appRefTracker.RemoveAppFromRefMap(kind, refName, app.Namespace, app.Name)
```
cppforlife: we should probably turn this into removing app from all refs. search could be limited by namespace in theory.

cmd/controller/apps_reconciler.go
danielhelfand: Refs are removed upon deletion via the ConfigMap/Secret handlers. When a Secret/ConfigMap is deleted, the key is removed using the `RemoveRef` func.

pkg/reftracker/ref_tracker.go
danielhelfand: That's fair. I have gone ahead and removed.

pkg/reftracker/ref_tracker.go
danielhelfand: All renamed. I instead named `RemoveAppFromRefMap -> RemoveApp` to `RemoveAppFromAllRefs` as it now handles deleting App from all associated refs.

test/e2e/app_secret_configmap_reconcile_test.go
danielhelfand: Sure, I added.

pkg/app/app.go
danielhelfand: Addressed

cmd/controller/apps_reconciler.go
danielhelfand: I changed the func to remove an App from all ref keys by passing in ref names.

cmd/controller/apps_reconciler.go
cppforlife: i meant when ref is not longer used by an app but it still exists within cluster.

danielhelfand: Okay, makes sense. Maybe what could be done is the AppRefTracker stores `map[refs]map[app]map[kind]ref`. During the updateAppRefs method in apps_reconciler, if the refs don't match, some kind of prune method could be called to update.

danielhelfand: I added a `PruneAppFromRefs` func to help address this. It requires the `AppRefTracker` to also carry the state what refs an App has.

pkg/reftracker/ref_tracker_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+package reftracker
+
+import "testing"
+
+func Test_AddAppForRef_AddsApp_WhenRefNotInMap(t *testing.T) {
+	appRefTracker := NewAppRefTracker()
+	appRefTracker.AddAppForRef("secret", "secretName", "default", "app")
+
+	if _, ok := appRefTracker.refsToApps["secret:secretName:default"]["app"]; !ok {
```
cppforlife: we shouldnt access private fields (refsToApps) in tests since that's couples tests with impl (also doesnt prove that class external behaviour is correct).

pkg/reftracker/ref_tracker.go
```diff
@@ -0,0 +1,130 @@
...
+	appKey := fmt.Sprintf(`%s:%s`, appName, namespace)
+	refsInState := a.appsToRefs[appKey]
+
+	// Compare current state against App's
+	// previous refs.
+	var diff []string
+	for ref := range refsInState {
+		// The format of ref is refName:kind
+		// so we need to get the refName portion.
+		index := strings.Index(ref, ":")
```
cppforlife: it seems that in quite a few places we use string concatenation to build a key. typically that's fine but as soon as we start unpacking that string, we should switch to a struct to hold this info. additionally all of the methods on this class take a lot of strings which is pretty error-prone to ordering mistakes. let's have a few tiny structs that represent what we need to key off things. we can also use them as args to functions to make it easier to distinguish fields.

cmd/controller/run.go
```diff
@@ -73,12 +76,16 @@ func Run(opts Options, runLog logr.Logger) {
...
 		ctrlAppOpts := controller.Options{
 			Reconciler: NewUniqueReconciler(&ErrReconciler{
 				delegate: &AppsReconciler{
-					appClient:  appClient,
-					appFactory: appFactory,
-					log:        runLog.WithName("ar"),
+					appClient:       appClient,
+					appFactory:      appFactory,
+					log:             runLog.WithName("ar"),
+					AppRefTracker:   appRefTracker,
```
cppforlife: this field should not be public.

cmd/controller/handlers/secret_handler.go
```diff
@@ -0,0 +1,56 @@
...
+	appUpdateStatus *reftracker.AppUpdateStatus
+}
+
+var _ handler.EventHandler = &SecretHandler{}
+
+func NewSecretHandler(log logr.Logger, as *reftracker.AppRefTracker, aus *reftracker.AppUpdateStatus) *SecretHandler {
+	return &SecretHandler{log, as, aus}
+}
+
+func (sch *SecretHandler) Create(evt event.CreateEvent, q workqueue.RateLimitingInterface) {}
```
cppforlife: wouldnt we want to enqueue app for updates if secret is created?

cmd/controller/handlers/secret_handler.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (sch *SecretHandler) Update(evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
+	sch.enqueueAppsForUpdate(evt.MetaNew.GetName(), evt.MetaNew.GetNamespace(), q)
+}
+
+func (sch *SecretHandler) Delete(evt event.DeleteEvent, q workqueue.RateLimitingInterface) {
+	sch.appRefTacker.RemoveRef("secret", evt.Meta.GetName(), evt.Meta.GetNamespace())
```
cppforlife: wouldnt we want to enqueue app for updates if secret is deleted?

cmd/controller/handlers/secret_handler.go
danielhelfand: If a secret is created, can we not assume that it will be included as part of the App reconcile process? The App should be using that version of the created secret/it will update the AppRefTracker.

danielhelfand: I guess there is the scenario of it the App is created before the secret. So maybe it does make sense to add.

cmd/controller/handlers/secret_handler.go
danielhelfand: Added

cmd/controller/handlers/secret_handler.go
danielhelfand: Added

cmd/controller/run.go
danielhelfand: Changed to private and exposed via func for using in tests.

pkg/reftracker/ref_tracker.go
danielhelfand: Created a `RefKey` struct that is a combo of the resource kind, name, and namespace to address this.

pkg/reftracker/ref_tracker_test.go
danielhelfand: Changed to accessing through appRefTracker funcs.

cmd/controller/apps_reconciler.go
```diff
@@ -35,5 +39,41 @@ func (r *AppsReconciler) Reconcile(request reconcile.Request) (reconcile.Result,
...
+	// Remove any reference associations for App
+	// if the App no longer uses refs it once did.
+	r.appRefTracker.PruneAppFromRefs(refKeys, app.Name, app.Namespace)
+}
+
+func (r *AppsReconciler) AppRefTracker() *reftracker.AppRefTracker {
+	return r.appRefTracker
+}
+
+func (r *AppsReconciler) SetAppRefTracker(appRefTracker *reftracker.AppRefTracker) {
```
cppforlife: instead of adding SetAppRefTracker just for tests, let's add a constructor for AppsReconciler.

pkg/reftracker/ref_key.go
```diff
@@ -0,0 +1,24 @@
...
+
+type RefKey struct {
+	kind, refName, namespace string
+}
+
+func NewRefKey(kind, refName, namespace string) RefKey {
+	return RefKey{kind, refName, namespace}
+}
+
+func (r RefKey) Kind() string {
```
cppforlife, line 28: it seems that Kind() is used as part of user messaging. instead let's add Description() method that returns useful description including kind, ref name, ns.

pkg/reftracker/ref_tracker.go
```diff
@@ -0,0 +1,120 @@
...
+func (a AppRefTracker) AddAppForRef(refKey RefKey, appName string) {
+	a.lock.Lock()
+	defer a.lock.Unlock()
+
+	apps := a.refsToApps[refKey]
+	if apps == nil {
+		apps = map[string]struct{}{}
+	}
+
+	appKey := fmt.Sprintf(`%s:%s`, appName, refKey.Namespace())
```
cppforlife: can we use refkeys for this as well?

pkg/reftracker/ref_key.go
```diff
@@ -0,0 +1,24 @@
+// Copyright 2021 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package reftracker
+
+type RefKey struct {
+	kind, refName, namespace string
+}
+
+func NewRefKey(kind, refName, namespace string) RefKey {
```
cppforlife, line 12: might be convenient to add NewConfigMapRefKey and NewSecretRefKey constructors

pkg/reftracker/app_update_status.go
```diff
@@ -0,0 +1,46 @@
...
+type AppUpdateStatus struct {
+	lock               sync.Mutex
+	appsToUpdateStatus map[string]struct{}
+}
+
+func NewAppUpdateStatus() *AppUpdateStatus {
+	return &AppUpdateStatus{appsToUpdateStatus: map[string]struct{}{}}
+}
+
+func (a AppUpdateStatus) MarkNeedsUpdate(appName, namespace string) {
```
cppforlife: any reason not to use refkeys here as well?

pkg/app/crd_app.go
```diff
@@ -119,10 +120,38 @@ func (a *CRDApp) updateApp(updateFunc func(*kcv1alpha1.App)) error {
...
+func (a *CRDApp) Reconcile(force bool) (reconcile.Result, error) {
+	return a.app.Reconcile(force)
 }
 
 func (a *CRDApp) watchChanges(callback func(kcv1alpha1.App), cancelCh chan struct{}) error {
 	return NewCRDAppWatcher(*a.appModel, a.appClient).Watch(callback, cancelCh)
 }
+
+// Get all SecretRefs from App spec
+func (a *CRDApp) SecretRefs() map[reftracker.RefKey]struct{} {
```
cppforlife: do we need these two methods?

cmd/controller/apps_reconciler.go
```diff
@@ -35,5 +39,41 @@ func (r *AppsReconciler) Reconcile(request reconcile.Request) (reconcile.Result,
...
+
+	// Make sure refs for App are always up to date
+	// in appRefTracker.
+	for refKey := range refKeys {
+		r.appRefTracker.AddAppForRef(refKey, app.Name)
+	}
+
+	// Remove any reference associations for App
+	// if the App no longer uses refs it once did.
+	r.appRefTracker.PruneAppFromRefs(refKeys, app.Name, app.Namespace)
```
cppforlife: seems like it would be good to merge AddAppForRef and PruneAppFromRefs together since they are trying to "reconcile" refs (both are working on refKeys)

pkg/reftracker/ref_tracker_test.go
```diff
@@ -0,0 +1,49 @@
+// Copyright 2021 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package reftracker
```
cppforlife: should be in the reftracker_test package

cmd/controller/apps_reconciler.go
danielhelfand: Added

pkg/reftracker/ref_key.go
danielhelfand: Added

pkg/reftracker/ref_tracker.go
danielhelfand: Changed to RefKey in general, but left in place as func param since all that is needed here is AppName.

pkg/reftracker/app_update_status.go
danielhelfand: Changed to RefKey

cmd/controller/apps_reconciler.go
danielhelfand: Combined the two funcs as ReconcileRefs()

pkg/reftracker/ref_key.go
danielhelfand: Add for those and apps.

pkg/reftracker/ref_key.go
```diff
@@ -0,0 +1,42 @@
...
+func (r RefKey) RefName() string {
+	return r.refName
+}
+
+func (r RefKey) Namespace() string {
+	return r.namespace
+}
+
+func (r RefKey) Description() string {
+	return fmt.Sprintf(`%s:%s:%s`, r.kind, r.refName, r.namespace)
```
cppforlife: nit: double quotes instead of ticks for regular strings

test/e2e/app_secret_configmap_reconcile_test.go
```diff
@@ -0,0 +1,256 @@
...
+		var cr v1alpha1.App
+		err := yaml.Unmarshal([]byte(out), &cr)
+		if err != nil {
+			t.Fatalf("Failed to unmarshal: %s", err)
+		}
+
+		// check for successful deployment
+		if cr.Status.Deploy == nil || !cr.Status.Deploy.Finished || cr.Status.Deploy.ExitCode != 0 {
+			t.Fatalf("Expected %s deployment to succeed but got:\n%s", name, cr.Status.Deploy.Stdout)
+		}
```
cppforlife: kapp deploy technically waits for App CR to sucessfully reach ReconcileSucceeded state or otherwise fails if ReconcileFailed was reached.

pkg/reftracker/ref_tracker.go
```diff
@@ -0,0 +1,114 @@
...
+		a.refsToApps[refKey] = apps
+
+		refs[refKey] = struct{}{}
+		a.appsToRefs[appKey] = refs
+	}
+
+	// Compare current state against App's
+	// previous refs.
+	refsInState := a.appsToRefs[appKey]
+	var diff []RefKey
```
cppforlife: better name for diff is refsToRemove

pkg/reftracker/ref_tracker.go
```diff
@@ -0,0 +1,114 @@
...
+	defer a.lock.Unlock()
+
+	// Add all new refs to AppRefTracker
+	for refKey := range currentRefs {
+		apps := a.refsToApps[refKey]
+		if apps == nil {
+			apps = map[RefKey]struct{}{}
+		}
+
+		appKey := NewAppKey(appKey.RefName(), refKey.Namespace())
```
cppforlife: i dont think i quite understand why we need to "rebuild" app key. might have to zoom about it.

cmd/controller/apps_reconciler.go
```diff
@@ -5,16 +5,24 @@ package controller
...
-	log        logr.Logger
-	appFactory AppFactory
+	appClient       kcclient.Interface
+	log             logr.Logger
+	appFactory      AppFactory
+	appRefTracker   *reftracker.AppRefTracker
+	appUpdateStatus *reftracker.AppUpdateStatus
+}
+
+func NewAppsReconciler(appClient kcclient.Interface, log logr.Logger, appFactory AppFactory, appRefTracker *reftracker.AppRefTracker, appUpdateStatus *reftracker.AppUpdateStatus) *AppsReconciler {
```
cppforlife: nit: split over two lines

PR #136, opened 2021-03-12
Only enqueue installed packages eligible for upgrade on package updates

ewrenn8:


cmd/controller/handlers/installed_pkg_version_handler.go
```diff
@@ -31,48 +35,69 @@ func NewInstalledPkgVersionHandler(c kcclient.Interface, log logr.Logger) *Insta
...
+	err := ipvh.enqueueEligibleInstalledPackages(q, evt.Object)
 	if err != nil {
 		ipvh.log.Error(err, "enqueueing all installed pakcages")
 	}
 }
 
-func (ipvh *InstalledPkgVersionHandler) enqueueAllPackages(q workqueue.RateLimitingInterface) error {
+func (ipvh *InstalledPkgVersionHandler) enqueueEligibleInstalledPackages(q workqueue.RateLimitingInterface, obj runtime.Object) error {
+	pkg := obj.(*pkgv1alpha1.Package)
 	installedPkgList, err := ipvh.client.InstallV1alpha1().InstalledPackages("").List(metav1.ListOptions{})
```
danielhelfand, line 70: Could we use `meta.ListOptions{}` in any way here to only pull back installed package with a specific public name?

ewrenn8: Currently, `metadata.name` and `metadata.namespace` are the only supported field selectors, so we aren't able to filter on the `spec.packageRef.publicName` or `spec.packageRef.version`

danielhelfand: Gotcha, pulling everything back makes sense then.

danielhelfand (approved):
lgtm

PR #137, opened 2021-03-12
Adding Linters for Checking for Deadcode

danielhelfand:
This pull request adds linters to detect if any deadcode is in any changes being added to kapp-controller. 

Currently, there are only two instances of unused code. For now, I have marked them to not be linted since I can see the field and func being useful in the future, but can also remove them.

### DIFF
--- a/pkg/app/app_reconcile.go
+++ b/pkg/app/app_reconcile.go
@@ -267,12 +267,3 @@ func (a *App) setDeleteCompleted(result exec.CmdRunResult) {
 func (a *App) removeAllConditions() {
 	a.app.Status.Conditions = nil
 }
-
-func (a *App) removeCondition(appCondType v1alpha1.AppConditionType) {
-	for i, cond := range a.app.Status.Conditions {
-		if cond.Type == appCondType {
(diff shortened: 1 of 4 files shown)


pkg/app/app_reconcile.go
```diff
@@ -268,6 +268,7 @@ func (a *App) removeAllConditions() {
 	a.app.Status.Conditions = nil
 }
 
+// nolint: unused
 func (a *App) removeCondition(appCondType v1alpha1.AppConditionType) {
```
cppforlife: should we kill this function?

pkg/app/crd_app.go
```diff
@@ -26,6 +26,7 @@ func init() {
 type CRDApp struct {
 	app      *App
 	appModel *kcv1alpha1.App
+	// nolint: unused
 	nsName   types.NamespacedName // TODO fill in?
```
cppforlife: lets kill it for now?

pkg/app/app_reconcile.go
danielhelfand: Removed

pkg/app/crd_app.go
danielhelfand: Removed

PR #140, opened 2021-03-19
Add configuration to skip TLS verification for certain hostnames

### LINKED ISSUE #131: Support 'skip tls verify' when pulling images from a registry with self signed certificate

We already support specifying CA certs for kapp-controller in kapp-controller-config ConfigMap via https://github.com/vmware-tanzu/carvel-kapp-controller/commit/dec56a36d524368295c6211e8d64da9dc78e1cbc#diff-8a8c9da6bd19cf725d993a692648e3b2c41b311d938be83bf3590e8e64c1021e.
There is also a requirement for skipping the TLS verification. imgpkg already supports it via `--registry-verify-certs=false`. 

We'd like to add a new field `skipTLSVerify: false` in kapp-controller-config ConfigMap for this purpose.

ewrenn8:


ewrenn8:
Blocked on release of vendir that contains [d765cee](https://github.com/vmware-tanzu/carvel-vendir/commit/d765cee49b923b931cae95faec1a939cf5e1648b).

cmd/controllerinit/run.go
```diff
@@ -81,12 +81,12 @@ func runControllerCmd(cmdName string, args []string) error {
 }
 
 func configureSystem() error {
-	globalConfigurer, err := global.NewConfigurer()
+	globalConfig, err := global.GetConfig()
```
cppforlife: nice, do-er objects arent that great, and Config is the object we are dealing with.

pkg/app/app_fetch.go
```diff
@@ -21,9 +22,16 @@ func (a *App) fetch(dstPath string) (string, exec.CmdRunResult) {
 		return "", exec.NewCmdRunResultWithErr(fmt.Errorf("Expected at least one fetch option"))
 	}
 
-	vendir := a.fetchFactory.NewVendir(a.app.Namespace)
 	var result exec.CmdRunResult
 
+	globalConf, err := global.GetConfig()
+	if err != nil {
+		result.AttachErrorf("getting gloabl config: %v", err)
```
cppforlife: our errors are capitalized.

cppforlife: also `%s`

pkg/app/app_fetch.go
```diff
@@ -21,9 +22,16 @@ func (a *App) fetch(dstPath string) (string, exec.CmdRunResult) {
...
+		return "", result
+	}
+
+	vendir := a.fetchFactory.NewVendir(a.app.Namespace, globalConf)
```
cppforlife: one of primary purposes of an object (app) to take a factory (fetchFactory) is for that factory to hide dependencies that are necessary to create an object (vendir). in this case globalConfig is a dependency of vendir object. there is no need as far as i see for that dependency to be known by app itself -> hence we should move that into NewVendir.

as a side benefit it would be ideal if config would only be fetched once, and not everytime App is reconciled.

pkg/config/global/config.go
```diff
@@ -28,14 +28,15 @@ const (
 	httpsProxyEnvVar = "https_proxy"
 	noProxyKey       = "noProxy"
 	noProxyEnvVar    = "no_proxy"
+
+	skipTLSKey = "dangerousSkipTLSVerify"
```
cppforlife: skipTLSKey -> skipTLSVerifyKey

pkg/config/global/config.go
```diff
@@ -75,7 +75,27 @@ func (gc *Configurer) Configure() error {
 	return nil
 }
 
-func (gc *Configurer) addTrustedCerts(certChain string) (err error) {
+func (gc *Config) SkipTLS(candidateDomain string) bool {
```
cppforlife: SkipTLS -> ShouldSkipTLSVerificationForDomain

this makes it sound more like we are asking whether it should be asked vs telling config to skip tls for particular domain.

pkg/config/global/config.go
```diff
@@ -28,14 +28,15 @@ const (
...
 
-type Configurer struct {
-	namespace string
-	client    kubernetes.Interface
+type Config struct {
+	configMap *v1.ConfigMap
 }
 
-func NewConfigurer() (*Configurer, error) {
+func GetConfig() (*Config, error) {
```
cppforlife: it seems that this method is spinning it up its own kube client. we should be able to use it coreClient here: https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/pkg/fetch/factory.go#L12

pkg/fetch/vendir.go
```diff
@@ -7,26 +7,33 @@ import (
...
 	vendirconf "github.com/vmware-tanzu/carvel-vendir/pkg/vendir/config"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	kyaml "sigs.k8s.io/yaml"
 )
 
-const vendirEntireDirPath = "."
+const (
+	vendirEntireDirPath     = "."
+	registryDefaultHostname = "index.docker.io"
```
cppforlife: const can be used within function that uses registryDefaultHostname making it more contained.

pkg/fetch/vendir.go
```diff
@@ -327,3 +336,32 @@ func (v *Vendir) configMapBytes(configMapRef vendirconf.DirectoryContentsLocalRe
 
 	return kyaml.Marshal(configMap)
 }
+
+// TODO: Should this be more generic than just image refs? This would require
+// some parsing and standardization on our end
+func (v *Vendir) skipTLS(url string) bool {
```
cppforlife: link in comments to original location from ggcr.

pkg/fetch/vendir.go
```diff
@@ -327,3 +336,32 @@ func (v *Vendir) configMapBytes(configMapRef vendirconf.DirectoryContentsLocalRe
...
+		// iff it contains a '.' or ':' character, otherwise it is all repository
+		// and the domain defaults to Docker Hub.
+		hostname = strings.Split(hostname, ":")[0]
+	}
+
+	// handle defaulting of empty hostnames
+	if hostname == "" {
+		hostname = registryDefaultHostname
+	}
+	return v.kcConfig.SkipTLS(hostname)
```
cppforlife: i would separate this line from this function so that we have a 1-to-1 ggcr function copy.

pkg/fetch/vendir.go
```diff
@@ -7,26 +7,33 @@ import (
...
 
 type Vendir struct {
 	nsName     string
 	coreClient kubernetes.Interface
 	config     vendirconf.Config
+	kcConfig   *global.Config
 }
 
-func NewVendir(nsName string, coreClient kubernetes.Interface) *Vendir {
+func NewVendir(nsName string, coreClient kubernetes.Interface, globalConfig *global.Config) *Vendir {
```
cppforlife: seems a little weird for this package to depend on some global config. should this be an interface?

cmd/controller/run.go
```diff
@@ -70,9 +71,16 @@ func Run(opts Options, runLog logr.Logger) {
 		os.Exit(1)
 	}
 
+	kcConfig, err := kcconfig.GetConfig(coreClient)
```
cppforlife, line 74: nice, fetched one time!

pkg/fetch/vendir.go
```diff
@@ -327,3 +336,25 @@ func (v *Vendir) configMapBytes(configMapRef vendirconf.DirectoryContentsLocalRe
...
 }
+
+// This function only works on image refs. If in the future we decide to
+// expand this option to other fetch options, we will need to add hostname
+// extraction for those
+func (v *Vendir) shouldSkipTLSVerify(url string) bool {
+	hostname := v.extractImageRefHostname(url)
+	log.Printf("HOSTNAME:%s", hostname)
+	skip := v.skipTLSConfig.ShouldSkipTLSForDomain(hostname)
+	log.Printf("SKIP: %v", skip)
```
danielhelfand: WDYT of the following for here:

```suggestion
	if skip {
	  log.Printf("Skip TLS verify for hostname: %s", hostname)
	}
```

Or do we want to log this in general?

test/e2e/imgpkg_bundle_test.go
```diff
@@ -126,3 +126,75 @@ spec:
...
+	})
+
+	logger.Section("deploy", func() {
+		kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name},
+			RunOpts{AllowError: true, IntoNs: true, StdinReader: strings.NewReader(yaml1)})
+
+		out := kapp.Run([]string{"inspect", "-a", name, "--raw", "--tty=false", "--filter-kind=App"})
+
+		var cr v1alpha1.App
+
```
danielhelfand: Nit: Remove extra space

pkg/fetch/vendir.go
ewrenn8: Oh, this is my bad. I left that in when debugging the weirdness around the error imgpkg was printing. I can remove the log all together

ewrenn8: Updated. Thanks for catching this!

test/e2e/imgpkg_bundle_test.go
ewrenn8: Fixed

PR #146, opened 2021-03-25
Add NoopDelete Field to App Spec

### LINKED ISSUE #141: Expose configuration to specify app deletion should be a noop.
Sometimes it is desirable to have app deletion be a noop instead of running kapp delete.

For example, if kapp-controller is managing apps on a separate workload cluster and the cluster is being deleted, kapp-controller will attempt to delete all apps. Some of the apps that are managed may have installed resources that are critical to the functioning of the cluster (e.g. CNI). Since kapp-controller doesn't order the deletion of apps, and nor should it, it could potentially delete a required app before deleting the other apps on the system, which would cause all subsequent deletes to hang.

Expose some configuration on the appCR that can be used to mark an app as a noop delete. This will cause kapp-controller to skip issuing the kapp delete command.

This allows users in the case above to avoid managing the order of deletion, and instead specify critical apps as noop delete, so that kapp-controller will not actually delete the app, since the resources will be cleaned up once the cluster itself is deleted.

danielhelfand:
Closes #141 

Allows users to specify that only the App CR should be deleted when a delete request is issued for the App. All resources associated with the App will still exist after the deletion goes through.

Open to any suggestions on the field name. `NoopDelete` to me doesn't seem to be the most apt description, but it is how this feature request has been commonly referred to.

### DIFF
--- a/test/e2e/noopdelete_test.go
+++ b/test/e2e/noopdelete_test.go
@@ -1,0 +1,70 @@
+// Copyright 2021 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package e2e
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+)
+
+func Test_NoopDelete_DeletesAfterServiceAccountDeleted(t *testing.T) {
+	env := BuildEnv(t)
+	logger := Logger{}
+	kapp := Kapp{t, env.Namespace, logger}
+	kubectl := Kubectl{t, env.Namespace, logger}
+	sas := ServiceAccounts{env.Namespace}
+
+	appName := "noop-delete-app-with-no-sa"
+	cfgMapName := "noop-configmap-with-no-sa"
+	appYaml := fmt.Sprintf(`
+---
+apiVersion: kappctrl.k14s.io/v1alpha1
+kind: App
+metadata:
+  name: %s
+  annotations:
+    kapp.k14s.io/change-group: kappctrl-e2e.k14s.io/apps
+spec:
+  serviceAccountName: kappctrl-e2e-ns-sa
+  noopDelete: true
+  fetch:
+    - inline:
+        paths:
+          file.yml: |
+            apiVersion: v1
+            kind: ConfigMap
+            metadata:
+              name: %s
+            data:
+              key: value
+  template:
+    - ytt: {}
+  deploy:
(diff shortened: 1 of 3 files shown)


pkg/app/app_reconcile.go
```diff
@@ -60,7 +60,19 @@ func (a *App) reconcileDelete() error {
...
+		result = a.delete(a.updateLastDeployNoReturn)
+	} else {
+		// If NoopDelete, preserve App's associated resources
+		// but delete App CR itself by removing finalizer.
+		err = a.unblockDeletion()
+		if err != nil {
+			result = exec.CmdRunResult{
+				Error: err,
+			}
+		}
```
cppforlife: this code is also present inside delete(). it would be good if it only exists in one place.

test/e2e/noopdelete_test.go
```diff
@@ -0,0 +1,71 @@
...
+	})
+
+	logger.Section("delete", func() {
+		cleanUpApp()
+	})
+
+	logger.Section("check ConfigMap still exists after delete", func() {
+		kubectl.Run([]string{"get", "configmap", cfgMapName})
+	})
+}
```
cppforlife: newline?

pkg/app/app_reconcile.go
```diff
@@ -60,7 +60,19 @@ func (a *App) reconcileDelete() error {
 
 	a.resetLastDeployStartedAt()
 
-	result := a.delete(a.updateLastDeployNoReturn)
+	var result exec.CmdRunResult
+	if !a.app.Spec.NoopDelete {
```
cppforlife: typically if we are working with both if branches we should deal with positive case first to avoid unnecessary negation (extra code reading burden)

test/e2e/noopdelete_test.go
```diff
@@ -0,0 +1,71 @@
...
+            apiVersion: v1
+            kind: ConfigMap
+            metadata:
+              name: %s
+            data:
+              key: value
+  template:
+    - ytt:
+        paths:
+          - file.yml
```
cppforlife: this could just be `- ytt: {}`

pkg/app/app_reconcile.go
danielhelfand: Moved the logic into `delete()`

PR #158, opened 2021-04-08
feat(template.helm): Support for Helm Chart spec v2

### LINKED ISSUE #53: [template] support helm v3
consider supporting include CRDs flag. still want to support helm v2. should we make this consistent with how vendir selects helm version?

migmartri:
Hi,

UPDATE: Changed the approach to follow the "option b" described [here](https://github.com/vmware-tanzu/carvel-kapp-controller/issues/53#issuecomment-814368436). The code now chooses both the binary and the right arguments for the `helm template` command based on the apiVersion in the `Chart.yaml` file. If you are interested in seeing the more simplistic solution of maintaining only one binary (option a), visit back this previous commit 9e9bd0abe2e86e10abb1f6e9aa9ef4b83f901972.

> NOTE: The approach chosen here has not been agreed by the core maintainers ([waiting for feedback](https://github.com/vmware-tanzu/carvel-kapp-controller/issues/53#issuecomment-814394241)). I am creating this PR just to kick of the contribution tires :). That way I can get familiar with the source code, CI e2e testing, CLA and so on. Happy to change/scrap it if necessary.

~This PR adds support to templating Helm Chart v2 templates by upgrading the bundled Helm binary to V3. The approach followed is the "option a" described [here](https://github.com/vmware-tanzu/carvel-kapp-controller/issues/53#issuecomment-814368436), some additional comments about the change:~

* The e2e test now runs both an old chart (v1 spec) and a v2 one to showcase backward compatibility.
* ~The `template` command format has changed so it needed to be updated accordingly.~
* The template args and binary names are chosen programatically. I tried to encapsulate this logic so it can be stripped easily in the future.
* I took a look at `vendir` [fetch code](https://github.com/vmware-tanzu/carvel-vendir/blob/develop/pkg/vendir/fetch/helmchart/sync.go#L143) which seems to leverages the `fetch` command which to my knowledge has not changed from v2 to v3. I can see that in vendir there is code to explicitly load a Helm 3 binary `helmBinary = "helm3"` but I can't figure out its purpose since the code does not seem to leverage new Helm functionality such as OCI registry pulls. What am I missing? 
* I bumped the chart version in `examples/redis-helm.yml` to use a non deprecated Chart. I also tried to bump `examples/nginx-helm-git.yml` but for this one it can not be done since newer versions of the chart have subchart dependencies, and kapp-controller does not run `helm dep update` (this is not a problem in the redis case because we fetch from a release tarball not from git). We could choose a different app though, let me know your preference.
* ~I was thinking about adding unit tests to the `templateDir` method at `pkg/template/helm_template.go` but that (I think) would require some work to decouple the code from the `helm` binary, using interfaces for example. I decided not to do so because that change might be too intrusive taking into account that this is my first contribution and risky since I do not know the codebase well enough. if you find it necessary I am happy to revisit that.~
* Added some unit tests for the new branching logic
* I tried to be consistent with the used imported libraries for yaml unmarshalling, filesystem handling and so on. Let me know if I missed anything. 
 
Let me know your thoughts!

Cheers!
Miguel

Closes #53

### DIFF
--- a/pkg/template/helm_template.go
+++ b/pkg/template/helm_template.go
@@ -51,7 +51,7 @@ func (t *HelmTemplate) TemplateDir(dirPath string) (exec.CmdRunResult, bool) {
 		namespace = t.opts.Namespace
 	}
 
-	args := []string{"template", chartPath, "--name", name, "--namespace", namespace}
+	args := []string{"template", name, chartPath, "--namespace", namespace, "--include-crds"}
 
 	for _, source := range t.opts.ValuesFrom {
 		var paths []string
--- a/pkg/template/helm_template.go
+++ b/pkg/template/helm_template.go
@@ -179,7 +179,7 @@ func (t *HelmTemplate) writeFile(dstPath, subPath string, content []byte) (strin
 	return newPath, nil
 }
 
-type helmTemplateCMDArgs struct {
+type helmTemplateCmdArgs struct {
 	binaryName string
 	args       []string
 }
@@ -187,10 +187,10 @@ type helmTemplateCMDArgs struct {
 const helmBinaryName = "helm"
 
 // DEPRECATED
-const helm2ChartSpecVersion = "v1"
 const helm2BinaryName = "helmv2"
+const helm2ChartSpecVersion = "v1"
 
-// auxiliary struct used during Chart.yaml unmarshall
+// auxiliary struct used for Chart.yaml unmarshalling
 type chartSpec struct {
 	APIVersion string
 }
@@ -199,7 +199,8 @@ type chartSpec struct {
 // Returns the Helm Binary Name and the arguments required to be passed to the "helm template" subcommand
 // The returned values depend on the ApiVersion property inside the Chart.yaml file.
 // apiVersion==v1 will fallback to old Helm 2 binary and command format.
-func helmTemplateCmdLookup(releaseName, chartPath, namespace string) (*helmTemplateCMDArgs, error) {
+func helmTemplateCmdLookup(releaseName, chartPath, namespace string) (*helmTemplateCmdArgs, error) {
+	// Load [chartPath]/Chart.yaml and inspect apiVersion value.
 	bs, err := ioutil.ReadFile(path.Join(chartPath, "Chart.yaml"))
 	if err != nil {
 		return nil, fmt.Errorf("helmTemplateCmdLookup: %w", err)
@@ -210,8 +211,8 @@ func helmTemplateCmdLookup(releaseName, chartPath, namespace string) (*helmTempl
 		return nil, fmt.Errorf("helmTemplateCmdLookup: %w", err)
 	}
 
-	// By default use Helm 3+ format except for chart.apiSpec=v1
-	res := &helmTemplateCMDArgs{
+	// By default, use Helm 3+ format except for chart.apiSpec=v1
+	res := &helmTemplateCmdArgs{
 		binaryName: helmBinaryName,
 		args:       []string{"template", releaseName, chartPath, "--namespace", namespace, "--include-crds"},
 	}
--- a/pkg/template/helm_template.go
+++ b/pkg/template/helm_template.go
@@ -9,6 +9,9 @@ import (
 	"io"
 	"io/ioutil"
 	goexec "os/exec"
+	"path"
+
+	"github.com/ghodss/yaml"
 
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/exec"
@@ -51,7 +54,15 @@ func (t *HelmTemplate) TemplateDir(dirPath string) (exec.CmdRunResult, bool) {
 		namespace = t.opts.Namespace
 	}
 
-	args := []string{"template", name, chartPath, "--namespace", namespace, "--include-crds"}
+	// Return Helm binary name and arguments based on the version defined on the Chart.yaml file.
+	// NOTE: This will be removed once we remove retro-compatibility with Helm V2 binary
+	cmdCtx, err := helmTemplateCmdLookup(name, chartPath, namespace)
+	if err != nil {
+		return exec.NewCmdRunResultWithErr(err), true
+	}
+
+	// Actual helm template arguments
+	args := cmdCtx.args
 
 	for _, source := range t.opts.ValuesFrom {
 		var paths []string
@@ -97,11 +108,11 @@ func (t *HelmTemplate) TemplateDir(dirPath string) (exec.CmdRunResult, bool) {
 
 	var stdoutBs, stderrBs bytes.Buffer
 
-	cmd := goexec.Command("helm", args...)
+	cmd := goexec.Command(cmdCtx.binaryName, args...)
 	cmd.Stdout = &stdoutBs
 	cmd.Stderr = &stderrBs
 
-	err := cmd.Run()
+	err = cmd.Run()
 
 	result := exec.CmdRunResult{
 		Stdout: stdoutBs.String(),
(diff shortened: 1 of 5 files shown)


danielhelfand (commented):
Hey @migmartri,

Thanks for the pr. I think this looks great. 

While I think it makes sense to move with the helm community in terms of pushing for the use of helm3, maybe it does make sense to support both binaries for now. As you mention, the relationship isn't 1-1 for templating, so maybe we should at least have a period where we find a way to announce that it will eventually be removed entirely. 

Not sure how others feel on it (@ewrenn8 @cppforlife), but just my 2 cents on it.

migmartri:
hi @danielhelfand, thanks for your response!

I understand, it's a best practice to deprecate a functionality not to replace it. In this case though I have mixed feelings. It feels like a lot of throwaway code that would need to be added when backward compatibility seems pretty good based on my tests, we will be still bundling a non maintained binary and there has been 1.5 years since the release of Helm 3.

In any case, I agree with you that this can be done in two steps, and signal the removal of Helm 2 in a major release down the line. I'll change this accordingly.

Thanks!

migmartri:
Hi @danielhelfand, I've updated the patch based on your feedback. ptal! 

Thanks!!

pkg/template/helm_template.go
```diff
@@ -167,3 +178,49 @@ func (t *HelmTemplate) writeFile(dstPath, subPath string, content []byte) (strin
 
 	return newPath, nil
 }
+
+type helmTemplateCMDArgs struct {
```
migmartri: I will rename this from `helmTemplateCMDArgs` to `helmTemplateCmdArgs` to be consistent with camelcase naming.

pkg/template/helm_template.go
```diff
@@ -167,3 +178,50 @@ func (t *HelmTemplate) writeFile(dstPath, subPath string, content []byte) (strin
...
+	res := &helmTemplateCmdArgs{
+		binaryName: helmBinaryName,
+		args:       []string{"template", releaseName, chartPath, "--namespace", namespace, "--include-crds"},
+	}
+
+	// DEPRECATED: Helm V2 will be removed in a future release
+	if chartSpec.APIVersion == helm2ChartSpecVersion {
+		res.binaryName = helm2BinaryName
+		res.args = []string{"template", chartPath, "--name", releaseName, "--namespace", namespace}
+	}
```
danielhelfand: Probably can do the following here instead:

```suggestion
	
	res := &helmTemplateCmdArgs{}
	// DEPRECATED: Helm V2 will be removed in a future release
	if chartSpec.APIVersion == helm2ChartSpecVersion {
		res.binaryName = helm2BinaryName
		res.args = []string{"template", chartPath, "--name", releaseName, "--namespace", namespace}
	} else {
	      // By default, use Helm 3+ format except for chart.apiSpec=v1
	      res.binaryName = helmBinaryName,
	      res.args = []string{"template", releaseName, chartPath, "--namespace", namespace, "--include-crds"}
	}
```

pkg/template/helm_template.go
```diff
@@ -51,7 +54,15 @@ func (t *HelmTemplate) TemplateDir(dirPath string) (exec.CmdRunResult, bool) {
 		namespace = t.opts.Namespace
 	}
 
-	args := []string{"template", chartPath, "--name", name, "--namespace", namespace}
+	// Return Helm binary name and arguments based on the version defined on the Chart.yaml file.
+	// NOTE: This will be removed once we remove retro-compatibility with Helm V2 binary
+	cmdCtx, err := helmTemplateCmdLookup(name, chartPath, namespace)
```
danielhelfand: Nit: Something more explicit like `helmCmdCtx` I think would make more sense here.

pkg/template/helm_template.go
migmartri: Thanks @danielhelfand for the review.

I am happy to change that but I'd love to know more about the suggestion so I can tailor my code better in the future, I am also quite rusty so this is a good exercise to learn. Is it that what I did is not Golang idiomatic? Do you find it more difficult to understand? Is there a bug?

Some background context about the reason I did it like that is that I am expecting the Helm v2 to go away soon and in my code it would mean to just remove the override down below, leaving the struct initialization intact. I know this is quite nitpick and not a strong reason :P

I appreciate it! Thank you!

cppforlife: i would actually go with @migmartri's version as well due to the reason mentioned

> I am expecting the Helm v2 to go away soon and in my code it would mean to just remove the override down below, leaving the struct initialization intact

though if we are being nitpicky, i would have also reinitialized res to a new struct to avoid picking up any unintended fields. not a big deal here.

pkg/template/helm_template.go
```diff
@@ -9,6 +9,9 @@ import (
 	"io"
 	"io/ioutil"
 	goexec "os/exec"
+	"path"
+
+	"github.com/ghodss/yaml"
```
cppforlife: let's not bring in new dependency here. we already are using sigs.k8s.io/yaml (example: https://github.com/vmware-tanzu/carvel-kapp-controller/blob/1844e157b6de4048cec3ba0e53fc699d37e9c71e/pkg/fetch/vendir.go#L17)

pkg/template/helm_template_test.go
```diff
@@ -0,0 +1,90 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package template
```
cppforlife: should be in a template_test package so that we do not use template's privates.

pkg/template/helm_template_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+func TestHelmTemplateCmdLookup(t *testing.T) {
+	tests := []struct {
+		chartSpecVersion string                // i.e v1 or v2
+		binaryName       string                // returned helm binary name
+		args             func(string) []string // returned helm template arguments
+	}{
+		{
+			"v1",
+			helm2BinaryName,
```
cppforlife: ideally tests should not use same constants that they are testing. let's hard code these values.

migmartri: Makes sense, I had mixed feelings when i was writing it. Thanks for pointing it out! :)

pkg/template/helm_template.go
migmartri: I actually chose this one because it was imported in 

https://github.com/vmware-tanzu/carvel-kapp-controller/blob/1844e157b6de4048cec3ba0e53fc699d37e9c71e/pkg/app/app.go#L9

I'll change it to `"sigs.k8s.io/yaml"` as you suggested and maybe in another patch we can work on converging yaml deps?

cppforlife: hmm, yeah, we should fix that separately.

pkg/template/helm_template_test.go
migmartri: In that case I'll not be able to unit test the un-exported `helmTemplateCmdLookup` function. 

I can see three ways a) To export the function b) To export the function only in testing or c) Do not unit test the unexported method and instead do test `(t *HelmTemplate) **TemplateDir(...)`? 

What would you suggest? I think ideally we should have c) but that might require way more work in terms of decoupling the code from the binary and leveraging interfaces instead and I am not sure if it is in the scope of this specific PR.

cppforlife: i would recommend exporting this function and associated return struct. (also might be good to rename this function to NewHelmTemplateCmdArgs since it's really acting as a constructor for HelmTemplateCmdArgs)

migmartri: Thanks!

pkg/template/helm_template.go
danielhelfand: > though if we are being nitpicky, i would have also reinitialized res to a new struct to avoid picking up any unintended fields. not a big deal here.

Yup, that was my thought but understand the reasoning. Works for me.

migmartri:
@cppforlife, @danielhelfand I've applied most of your feedback, ptal

Thanks!

migmartri:
> @cppforlife, @danielhelfand I've applied most of your feedback, ptal
> 
> Thanks!

Hi! 

Any update on this? Anything you'd like me to apply to the PR?

Thanks!

cppforlife:
@migmartri sorry for not merging this immediately. we are trying to merge a large-ish branch right now into develop. we will merge this PR on top of that next week. thanks again for contributing it.

migmartri:
No rush, just checking in case there was anything I should do. Thanks!

migmartri:
Thanks for merging it @cppforlife! 

I have a question about releases, do we have any canary/develop release of kapp controller that could be used or do I need to build my own container image to use it? 

Is there any release cadence or timeframe I could expect for this change to be live?

Thanks!

PR #160, opened 2021-04-12
Add synchronize event to pull requests

danielhelfand:
This is needed for triggering CI from forks of kapp-controller. 

More details here: https://github.community/t/what-is-a-pull-request-synchronize-event/14784


migmartri:
It seems to have fixed my issue!

Thanks!

PR #162, opened 2021-04-12
Add aggregated api server

ewrenn8:
Adds an aggregated api server which is backed by an internal CRD. This will allow us to introducce custom behavior only available through aggregated api servers while delaying the complexity associated with other storage solutions.


ewrenn8:
I will open another PR as a follow up to this to rename our api groups

cmd/controller/run.go
```diff
@@ -79,12 +83,27 @@ func Run(opts Options, runLog logr.Logger) {
...
 	}
 
+	server, err := apiserver.NewAPIServer(restConfig)
+	if err != nil {
+		runLog.Error(err, "creating server")
+		os.Exit(1)
+	}
+	server.Run()
+
+	// TODO: we may need to sleep here to give the server time to start up
```
danielhelfand, line 105: How common of an issue is this? Is there any kind of wait we can put in place for this?

.golangci.yml
```diff
@@ -1,3 +1,5 @@
+run:
+  timeout: 5m
```
danielhelfand, line 2: Thanks for bumping this up. We can keep an eye on it to see what appropriate timeout should be with using it through GitHub Actions.

config/crd.yml
```diff
@@ -37,40 +37,6 @@ spec:
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
-metadata:
```
danielhelfand, line 40: Not following here on why Packages is removed. Mind explaining to me?

cmd/controller/run.go
ewrenn8: I have yet to hit it locally, so I didn't put any kind of waiting behavior in there. I think if we want to be safe we may be able to add something as simple as a couple second sleep.

config/crd.yml
ewrenn8: CRDs are managed by the kubernetes api, so if we left package in as a CRD, it would conflict with the group+version we are trying to register our api server with.

cmd/controller/run.go
```diff
@@ -79,12 +83,27 @@ func Run(opts Options, runLog logr.Logger) {
...
 		kcConfig:   kcConfig,
 		appClient:  kcClient,
 	}
 
+	server, err := apiserver.NewAPIServer(restConfig)
+	if err != nil {
+		runLog.Error(err, "creating server")
+		os.Exit(1)
+	}
+	server.Run()
```
cppforlife, line 103: does this return err?

cmd/controller/run.go
```diff
@@ -197,6 +217,7 @@ func Run(opts Options, runLog logr.Logger) {
 	}
 
 	runLog.Info("Exiting")
+	server.Stop()
```
cppforlife, line 220: do we need this? what kind of cleanup are we doing here?

config/crd.yml
```diff
@@ -134,3 +100,22 @@ spec:
     type: date
     description: Time since creation
     JSONPath: .metadata.creationTimestamp
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: internalpackages.install.package.carvel.dev
```
cppforlife, line 107: note for future: should this be in install or in data?

config/rbac.yml
```diff
@@ -24,17 +24,33 @@ rules:
...
+  verbs: ["update", "get"]
+- apiGroups: [""]
+  resources: ["namespaces"]
+  verbs: ["list", "watch"]
+- apiGroups: ["admissionregistration.k8s.io"]
+  resources: ["mutatingwebhookconfigurations"]
+  verbs: ["list", "watch"]
+- apiGroups: ["admissionregistration.k8s.io"]
+  resources: ["validatingwebhookconfigurations"]
+  verbs: ["list", "watch"]
```
cppforlife, line 53: do we need webhook perms?

pkg/apis/kappctrl/v1alpha1/status.go
```diff
@@ -25,6 +25,7 @@ const (
 )
 
 // TODO rename to Condition
+// +k8s:openapi-gen=true
```
cppforlife, line 28: why do we need openapi-gen here?

cmd/controller/run.go
ewrenn8: No not really. I just put in there for completeness, but because we are exiting, its not required.

config/crd.yml
ewrenn8: I was thinking these go in to their own group `internal.packaing.dev`, so that the group can just be removed entirely in the future

config/rbac.yml
ewrenn8: Yea, the webserver won't start if it can't watch these resources

pkg/apis/kappctrl/v1alpha1/status.go
ewrenn8: Packages use the same status as apps, so we need to have openapi schemas for the structs that make up a status in order for the apiserver to run.

PR #164, opened 2021-04-15
Elevate UsefulErrorMessage to InstalledPackage Status

### LINKED ISSUE #133: Add more debug info to failing InstalledPackages
When debugging an installed package that is failing reconciliation, the failure message just states "Reconcile failed: App failed reconciling", which doesn't give much direction to the user. This message also requires the user to move a level below the Packaging APIs and debug an app directly.

Surface more useful information when an InstalledPackage is failing to reconcile due to the underlying app. One idea is to copy any status fields that have a non zero exit code to the installed package to surface the debug info without having to look at the App CR

- [ ] Add debug info in package CR
- [ ] Add debug tips to docs

Is there more or other info we should surface on the installed package?

danielhelfand:
Closes #133 

As part of our effort to help users debug errors with the Packaging API much quicker, this pull request adds the `UsefulErrorMessage` of an App to the InstalledPackage status. Now, users will not need to know about the existence of Apps when working with InstalledPackages to troubleshoot errors.

```
$ k get ipkg -A

NAMESPACE   NAME                  PACKAGE NAME          PACKAGE VERSION   DESCRIPTION                                                            AGE
foo         instl-pkg-test-fail   pkg.fail.carvel.dev   1.0.0             Reconcile failed: Error (see .status.usefulErrorMessage for details)   12s
```

```
$ k describe ipkg -A

...
Status:
  Conditions:
    Message:             Error (see .status.usefulErrorMessage for details)
    Status:              True
    Type:                ReconcileFailed
  Friendly Description:  Reconcile failed: Error (see .status.usefulErrorMessage for details)
  Observed Generation:   1
  Useful Error Message:  kapp: Error: Checking existance of resource service/simple-app (v1) namespace: does-not-exist: services "simple-app" is forbidden:
  User "system:serviceaccount:foo:kappctrl-e2e-ns-sa" cannot get resource "services" in API group "" in the namespace "does-not-exist" (reason: Forbidden)
  Version:  1.0.0
Events:     <none>
```

```
$ k get ipkg -A -o yaml

...
status:
    conditions:
    - message: Error (see .status.usefulErrorMessage for details)
      status: "True"
      type: ReconcileFailed
    friendlyDescription: 'Reconcile failed: Error (see .status.usefulErrorMessage
      for details)'
    observedGeneration: 1
    usefulErrorMessage: |-
      kapp: Error: Checking existance of resource deployment/simple-app (apps/v1) namespace: does-not-exist: deployments.apps "simple-app" is forbidden:
        User "system:serviceaccount:foo:kappctrl-e2e-ns-sa" cannot get resource "deployments" in API group "apps" in the namespace "does-not-exist" (reason: Forbidden)
    version: 1.0.0
```


test/e2e/installed_package_test.go
```diff
@@ -143,3 +144,102 @@ stringData:
...
+	env := BuildEnv(t)
+	logger := Logger{}
+	kapp := Kapp{t, env.Namespace, logger}
+	kubectl := Kubectl{t, env.Namespace, logger}
+	sas := ServiceAccounts{env.Namespace}
+	name := "instl-pkg-test-fail"
+
+	// contents of this repo (k8slt/installedpackage-fail)
+	// under test/e2e/assets/installedpackage-fail-test
+	installPkgYaml := fmt.Sprintf(`---
```
ewrenn8, line 155: Can we just inline package yml and intalled pkg yml here? Why do we need to use a repo for this test?

test/e2e/installed_package_test.go
```diff
@@ -143,3 +144,102 @@ stringData:
...
+		return err
+	}
+	err := retry(30*time.Second, retryFunc)
+	if err != nil {
+		t.Fatalf("Expected error from InstalledPackage %s: %v", name, err)
+	}
+	// Remove error message since asserted on already
+	cr.Status.UsefulErrorMessage = ""
+
+	if !reflect.DeepEqual(expectedStatus, cr.Status) {
```
ewrenn8: can we consolidate this check with the one in the retry? Useful error message is a field on status, so can we just assert status is equal in there instead of specifically asserting on useful error message? Alternatively, if we want it separate for explicitness of test, maybe we just assert on UsefulErrorMessage and remove the deep equals check since the test scope seems to be focused on just UsefulErrorMessage

danielhelfand: I guess what we can do is also check that the `FriendlyDescription` updates in the retry. I think that is the other main property we care about.

danielhelfand: Updated

test/e2e/installed_package_test.go
danielhelfand: Inlined Package definition

ewrenn8:
lgtm

PR #170, opened 2021-04-19
bump vendir

ewrenn8:


cppforlife:
@ewrenn8 did you want to update Dockerfile here as well?

PR #171, opened 2021-04-20
Delete app if serviceaccount does not exist and no resources deployed

### LINKED ISSUE #114: App gets stuck deleting if service account is not created
**What steps did you take:**
Installed the "simple-app" from the instructions without first creating the service account for it. 
https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/docs/walkthrough.md

Tried to delete the app via `kapp -a simple-app delete` and see it hang/timeout.

**What happened:**
```
$ k get app
NAME         DESCRIPTION                                                                                         SINCE-DEPLOY   AGE
simple-app   Delete failed: Preparing kapp: Getting service account: serviceaccounts "default-ns-sa" not found   3s             56m
```

**What did you expect:**
Delete should succeed

**Environment:**

- kapp Controller version : v0.16.0
- Kubernetes version:
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:50:19Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.2", GitCommit:"faecb196815e248d3ecfb03c680a4507229c2a56", GitTreeState:"clean", BuildDate:"2021-01-21T01:11:42Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
```

danielhelfand:
Closes #114 

This pull request adds the ability to delete an App CR in the event that the user forgot to initially deploy appropriate RBAC for the App. It also preserves existing behavior (i.e. not proceeding with delete) for Apps that have been deployed successfully but where the serviceaccount has been deleted.

For Apps that have never deployed any resources successfully, this approach simply removes the finalizer for the App. It is known whether the App has deployed resources at some point via a new App status field called `HasDeployedResources`.

### DIFF
--- a/pkg/apis/kappctrl/v1alpha1/types.go
+++ b/pkg/apis/kappctrl/v1alpha1/types.go
@@ -77,8 +77,9 @@ type AppStatus struct {
 	ObservedGeneration int64          `json:"observedGeneration"`
 	Conditions         []AppCondition `json:"conditions"`
 
-	FriendlyDescription string `json:"friendlyDescription"`
-	UsefulErrorMessage  string `json:"usefulErrorMessage"`
+	FriendlyDescription  string `json:"friendlyDescription"`
+	UsefulErrorMessage   string `json:"usefulErrorMessage"`
+	HasDeployedResources bool
 }
 
 type AppStatusFetch struct {
--- a/pkg/app/app_deploy.go
+++ b/pkg/app/app_deploy.go
@@ -5,6 +5,7 @@ package app
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
 	ctldep "github.com/vmware-tanzu/carvel-kapp-controller/pkg/deploy"
@@ -63,6 +64,10 @@ func (a *App) delete(changedFunc func(exec.CmdRunResult)) exec.CmdRunResult {
 
 				kapp, err := a.newKapp(*dep.Kapp, cancelCh)
 				if err != nil {
+					if strings.Contains(err.Error(), "Getting service account") && !a.app.Status.HasDeployedResources {
+						err = a.unblockDeletion()
+					}
+
 					return exec.NewCmdRunResultWithErr(fmt.Errorf("Preparing kapp: %s", err))
 				}
 
--- a/pkg/app/app_deploy.go
+++ b/pkg/app/app_deploy.go
@@ -5,9 +5,9 @@ package app
 
 import (
 	"fmt"
-	"strings"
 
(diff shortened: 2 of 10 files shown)


pkg/app/app_deploy.go
```diff
@@ -63,6 +64,10 @@ func (a *App) delete(changedFunc func(exec.CmdRunResult)) exec.CmdRunResult {
 
 				kapp, err := a.newKapp(*dep.Kapp, cancelCh)
 				if err != nil {
+					if strings.Contains(err.Error(), "Getting service account") && !a.app.Status.HasDeployedResources {
```
ewrenn8: Can we use a typed error here? Not a huge fan of tying ourselves to error messages

pkg/apis/kappctrl/v1alpha1/types.go
```diff
@@ -77,8 +77,9 @@ type AppStatus struct {
 	ObservedGeneration int64          `json:"observedGeneration"`
 	Conditions         []AppCondition `json:"conditions"`
 
-	FriendlyDescription string `json:"friendlyDescription"`
-	UsefulErrorMessage  string `json:"usefulErrorMessage"`
+	FriendlyDescription  string `json:"friendlyDescription"`
+	UsefulErrorMessage   string `json:"usefulErrorMessage"`
+	HasDeployedResources bool
```
cppforlife, line 82: missing json name styling.

cppforlife: it seems that we are defaulting to having "no resources deployed". this could be problematic if status is cleared out and kapp-controller skips deletion even though resources were deployed. how would we fix this?

danielhelfand: Maybe it would it make sense to have something similar to `RefTracker` or `AppUpdateStatus` for Secret/ConfigMap updates? And then we could remove this from the status entirely.

cppforlife: > for Secret/ConfigMap updates

lost you there. what would this do?

danielhelfand: Less about Secret/ConfigMap updates and more about how their state is stored (i.e. in memory). We could have Apps register that they have deployed resources in a map that can be accessed to check whether to proceed with delete in this case. We could also then remove `HasDeployedResources` from the status.

cppforlife: this state though is not as transient as other state, since if you update kapp-controller you wouldnt know if something was previously deployed or not.

PR #173, opened 2021-04-20
Use deploy when calculating next reconcile

### LINKED ISSUE #168: should we calculate next desired reconcile time from the last deploy time

currently it's calculated from the last fetch time. in my case my fetch+template+deploy was taking ~30s, which means as soon as this deploy finished, next cycle was already ready to go.

should we switch to check against lastUpdatedAt for deploy? what would SINCE-DEPLOY field show in kubectl get?

ewrenn8:
fixes https://github.com/vmware-tanzu/carvel-kapp-controller/issues/168

### DIFF
--- a/pkg/app/reconcile_timer.go
+++ b/pkg/app/reconcile_timer.go
@@ -39,20 +39,29 @@ func (rt ReconcileTimer) IsReadyAt(timeAt time.Time) bool {
 		return false
 	}
 
-	// Did we deploy at least once?
+	var lastReconcileTime time.Time
+
+	// Use latest deploy time if available, otherwise fallback to fetch
+	// If no timestamp is available, enqueue immediately
+	lastDeploy := rt.app.Status.Deploy
 	lastFetch := rt.app.Status.Fetch
-	if lastFetch == nil {
-		return true
+	if lastDeploy != nil && !lastDeploy.UpdatedAt.Time.IsZero() {
+		lastReconcileTime = rt.app.Status.Deploy.UpdatedAt.Time
+	} else {
+		if lastFetch == nil {
+			return true
+		}
+		lastReconcileTime = lastFetch.UpdatedAt.Time
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/app/reconcile_timer.go
```diff
@@ -39,20 +39,27 @@ func (rt ReconcileTimer) IsReadyAt(timeAt time.Time) bool {
...
-	// Did we deploy at least once?
-	lastFetch := rt.app.Status.Fetch
-	if lastFetch == nil {
-		return true
+	var lastReconcileTime time.Time
+
+	// Use latest deploy time if available, otherwise fallback to fetch
+	// If no timestamp is available, enqueue immediately
+	if rt.app.Status.Deploy != nil {
+		lastReconcileTime = rt.app.Status.Deploy.UpdatedAt.Time
```
cppforlife, line 49: hmm, i wonder if we should be defensive here and check if UpdatedAt is zero time. wdyt?

ewrenn8: Is this even possible? Looks like any time we issue an update to the status we are also getting the time, so they should be either nil or have a time. I dont know that adding in the check would hurt anything so happy to do it.

cppforlife: i was just thinking it's a bit hard to keep this invariant true. should we be defensive here to avoid accidently hammering kapp-controller if we screw up somewhere?

ewrenn8: I'm not sure what the check would do here though. If we return true because it is 0, it is the same as returning true because currTime - 0 >= 30s, which will always be true. If we return false, we will never run a reconcile loop because the status will never be updated and the time will always be 0, so users would have to unstick the app by updating it somehow.

cppforlife: we would just not use deploy section, fallback to fetch.

PR #174, opened 2021-04-20
Add status to package repository

ewrenn8:


test/e2e/package_repo_test.go
```diff
@@ -5,15 +5,83 @@ package e2e
...
+	err := retry(30*time.Second, retryFunc)
+	if err != nil {
+		t.Fatalf(err.Error())
+	}
+
+}
+
+func Test_PackageRepoStatus_Success(t *testing.T) {
+
+}
```
danielhelfand: Are you going to implement this test? We could also assert on the Repo status in tests already created.

test/e2e/package_repo_test.go
```diff
@@ -5,15 +5,83 @@ package e2e
...
+	name := "repo"
+
+	repoYaml := `apiVersion: install.package.carvel.dev/v1alpha1
+kind: PackageRepository
+metadata:
+  name: test-repo
+spec:
+  fetch:
+    imgpkgBundle:
+      image: k8slt/i-dont-exist`
```
danielhelfand, line 36: Since this is a fetch failure test, should this actually be a non e2e test? Maybe something similar to app_reconcile tests: https://github.com/vmware-tanzu/carvel-kapp-controller/blob/11d7094fa5485ce75b448b06e0052ac9d4efe5c4/pkg/app/app_reconcile_test.go?

pkg/pkgrepository/pkgrepository.go
```diff
@@ -52,7 +54,36 @@ func (ip *PackageRepositoryCR) Reconcile() (reconcile.Result, error) {
...
+	ip.log.Info(fmt.Sprintf("Reconciling InstalledPackage '%s/%s'", ip.model.Namespace, ip.model.Name))
+	switch {
+	case appStatus.IsReconciling():
+		status.SetReconciling(ip.model.ObjectMeta)
+	case appStatus.IsReconcileSucceeded():
+		status.SetReconcileCompleted(nil)
+	case appStatus.IsReconcileFailed():
+		status.SetUsefulErrorMessage(existingApp.Status.UsefulErrorMessage)
+		status.SetReconcileCompleted(fmt.Errorf("Syncing packages (see .status.usefulErrorMessage for details)"))
+
```
danielhelfand: Nit: extra line

pkg/pkgrepository/pkgrepository.go
```diff
@@ -52,7 +54,36 @@ func (ip *PackageRepositoryCR) Reconcile() (reconcile.Result, error) {
...
+	appStatus := reconciler.Status{S: existingApp.Status.GenericStatus}
+	ip.log.Info(fmt.Sprintf("Reconciling InstalledPackage '%s/%s'", ip.model.Namespace, ip.model.Name))
+	switch {
+	case appStatus.IsReconciling():
+		status.SetReconciling(ip.model.ObjectMeta)
+	case appStatus.IsReconcileSucceeded():
+		status.SetReconcileCompleted(nil)
+	case appStatus.IsReconcileFailed():
+		status.SetUsefulErrorMessage(existingApp.Status.UsefulErrorMessage)
+		status.SetReconcileCompleted(fmt.Errorf("Syncing packages (see .status.usefulErrorMessage for details)"))
```
danielhelfand: Nit: 

```suggestion
		status.SetReconcileCompleted(fmt.Errorf("Syncing packages: (see .status.usefulErrorMessage for details)"))
```

test/e2e/package_repo_test.go
ewrenn8: Oh whoops. Yea, I will go back and either implement or add to another test

test/e2e/package_repo_test.go
ewrenn8: Hmm thats an interesting question. I am leaning towards leaving it as an e2e test because I would like to see the tool's stderr in the Useful error message. That said, I think maybe talking about what makes e2e vs unit tests could be helpful. I usually think of unit as not needing outside dependencies, which the second test in that file actually requires since it is shelling out to ytt.

cppforlife: failure scenarios imho are great tests to have at e2e level. it's as common code path as success case.

test/e2e/package_repo_test.go
```diff
@@ -5,15 +5,144 @@ package e2e
...
+		}
+
+		// assert on expectedStatus
+		if !reflect.DeepEqual(expectedStatus, cr.Status) {
+			return fmt.Errorf("\nstatus is not same:\nExpected:\n%#v\nGot:\n%#v", expectedStatus, cr.Status)
+		}
+		return nil
+	}
+
+	err := retry(30*time.Second, retryFunc)
```
cppforlife: instead of having retry return an error, i would pass down testing.T to it so that it can do t.Fatalf inside. once that's done it's also nicer to inline retryFunc closure.

test/e2e/package_repo_test.go
```diff
@@ -5,15 +5,144 @@ package e2e
...
+	cleanUp := func() {
+		kapp.Run([]string{"delete", "-a", name})
+	}
+
+	cleanUp()
+	defer cleanUp()
+
+	logger.Section("deploy", func() {
+		kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name},
+			RunOpts{IntoNs: true, StdinReader: strings.NewReader(repoYml)})
```
cppforlife: `IntoNs: true`?

test/e2e/package_repo_test.go
danielhelfand: FWIW, the tool's stderr would still be captured in the `usefulErrorMessage` in a non e2e test. Not saying e2e tests should not test errors. I am more saying that this can be tested without needing someone to set up what is needed for e2e tests (i.e. a Kubernetes cluster).

For me personally, I have been thinking e2e tests are needed when it would be valuable to actually have deployment take place as part of the test. Maybe something we can hash out a bit further as far as what appropriate testing levels are for kapp-controller.

I would say let's just revisit this test later on as I think it can be run quicker with less dependencies in a non e2e test.

cppforlife: the objective here isnt to specifically test that usefulErrorMessage is being set, but rather here is what the failing status looks like for the real failing package repo. keep in mind that we are trying to capture user facing behaviour of failing pkg repo -- if tomorrow we redo package repo entirely in a different way this test should not be affected at all. (this is different from a test like this: https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/pkg/app/app_reconcile_test.go where it's more coupled to implementation)

> without needing someone to set up what is needed for e2e tests (i.e. a Kubernetes cluster).

app_reconcile_test.go test has its own dependencies which could be considered more bothersome than k8s. + we need to have k8s anyway for other tests.

> I would say let's just revisit this test later on

sure, we can have longer discussion about this some time.

danielhelfand (commented):
lgtm

PR #178, opened 2021-04-26
Introduce informational fields to package CR

ewrenn8:


ewrenn8:
Associated docs PR: https://github.com/vmware-tanzu/carvel.dev/pull/97

pkg/apiserver/apis/packages/types.go
```diff
@@ -38,8 +38,17 @@ type PackageSpec struct {
...
+	DisplayName          string `json:"displayName,omitempty"`
+	LongDescription      string `json:"longDescription,omitempty"`
+	ShortDescription     string `json:"shortDescription,omitempty"`
+	Icon                 string `json:"icon,omitempty"`
+	ProviderName         string `json:"providerName,omitempty"`
+	Maintainers          string `json:"maintainers,omitempty"`
+	ReleaseNotes         string `json:"releaseNotes,omitempty"`
+	Category             string `json:"category,omitempty"`
+	SupportDescription   string `json:"supportDescription,omitempty"`
+	CapactiyRequirements string `json:"capacityRequirementsDescription,omitempty"`
```
cppforlife: CapactiyRequirements -> CapactiyRequirementsDescription

pkg/apiserver/apis/packages/types.go
```diff
@@ -38,8 +38,17 @@ type PackageSpec struct {
...
+	LicenseDescription   string `json:"licenseDescription,omitempty"`
```
cppforlife: should we turn this into `Licenses []string` (not description)?

pkg/apiserver/apis/packages/types.go
```diff
@@ -38,8 +38,17 @@ type PackageSpec struct {
...
 	Version    string `json:"version,omitempty"`
 
-	DisplayName string `json:"displayName,omitempty"`
-	Description string `json:"description,omitempty"`
+	DisplayName          string `json:"displayName,omitempty"`
+	LongDescription      string `json:"longDescription,omitempty"`
+	ShortDescription     string `json:"shortDescription,omitempty"`
+	Icon                 string `json:"icon,omitempty"`
+	ProviderName         string `json:"providerName,omitempty"`
+	Maintainers          string `json:"maintainers,omitempty"`
```
cppforlife: should we make this []Maintainer where Maintainer is struct{Name string}?

pkg/apiserver/apis/packages/types.go
```diff
@@ -38,8 +38,17 @@ type PackageSpec struct {
 	PublicName string `json:"publicName,omitempty"`
 	Version    string `json:"version,omitempty"`
 
-	DisplayName string `json:"displayName,omitempty"`
-	Description string `json:"description,omitempty"`
+	DisplayName          string `json:"displayName,omitempty"`
+	LongDescription      string `json:"longDescription,omitempty"`
+	ShortDescription     string `json:"shortDescription,omitempty"`
+	Icon                 string `json:"icon,omitempty"`
```
cppforlife: IconSVGBase64?

pkg/apiserver/apis/packages/types.go
```diff
@@ -38,8 +38,17 @@ type PackageSpec struct {
...
+	ProviderName         string `json:"providerName,omitempty"`
+	Maintainers          string `json:"maintainers,omitempty"`
+	ReleaseNotes         string `json:"releaseNotes,omitempty"`
+	Category             string `json:"category,omitempty"`
```
cppforlife: Category -> Categories []string?

pkg/apiserver/apis/packages/types.go
```diff
@@ -38,8 +38,17 @@ type PackageSpec struct {
 	PublicName string `json:"publicName,omitempty"`
 	Version    string `json:"version,omitempty"`
 
-	DisplayName string `json:"displayName,omitempty"`
-	Description string `json:"description,omitempty"`
```
cppforlife, line 42: should we also add a ReleasedAt?

pkg/apiserver/apis/packages/types.go
ewrenn8: Does this need to be an svg? Maybe just base64Icon?

pkg/apiserver/apis/packages/types.go
ewrenn8: Not sure we would want to inline the licenses, would we?

pkg/apiserver/apis/packages/types.go
ewrenn8: ReleaseDate?

pkg/apiserver/apis/packages/types.go
cppforlife: this would be somethig like this: []string{"MIT", "Apache 2.0"}

pkg/apiserver/apis/packages/types.go
cppforlife: if it's not svg it would be hard to incorporate it into various environments (light mode, dark mode, different sizes, etc.). separately i dont think it's good to use "base64" prefix instead of suffix since the main subject here is "icon".

pkg/apiserver/apis/packages/types.go
cppforlife: it would be a timestamp though. k8s does have a pattern for naming timestamps (creationTimestamp, deletionTimestamp, ...). alternatively "at" suffix is also commonly used for representing timestamps.

danielhelfand (commented):
lgtm

PR #181, opened 2021-04-28
Add schemas to crds

### LINKED ISSUE #117: Users should receive an error when Packaging CRDs do not follow the schema
Add openAPI schema to the packaging CRDs so users of the new apis receive an error when their packaging CRs are invalid

- [ ]  Package
- [ ]  PackageRepository
- [ ]  InstalledPackage

Some other things to consider in regards to Package CR
- [ ] Validations on Package metadata such as length, type, required (True or False) - refer to [issue 156](https://github.com/vmware-tanzu/carvel-kapp-controller/issues/156)
- [ ] Validations on package name to require 2 dots - refer to [issue 165](https://github.com/vmware-tanzu/carvel-kapp-controller/issues/165)

danielhelfand:
Closes #117 

This pull request generates kapp-controller's CRDs with schemas using [`controller-gen`](https://book.kubebuilder.io/reference/generating-crd.html). Using marker comments, the CRDs are generated from kapp-controller's types under `pkg/apis`.

In addition to generating the CRD yaml, `controller-gen` also generates the schema for the CRD. Using marker comments, it can be specified if fields are optional or required. These marker comments also allow for validations. Currently, the only validations I have added are for the App spec, which will require at least 1 item for `fetch`, `template`, and `deploy` steps: `// +kubebuilder:validation:MinItems=1`. Open to any thoughts on what should be optional versus required and additional validations.

To regenerate the CRDs, the `gen-crds.sh` script can be run to update the CRD yaml and the files will be place under the `config` folder.

Some to TODOs to consider for the future:
* We should provide descriptions for each property of the CRDs by providing a comment over each field
* Consider additional validation to be added to the schema


pkg/apis/kappctrl/v1alpha1/types.go
```diff
@@ -34,28 +41,46 @@ type AppList struct {
...
+	// +optional
+	Cluster *AppCluster `json:"cluster,omitempty"`
+
+	// +kubebuilder:validation:MinItems=1
+	Fetch []AppFetch `json:"fetch,omitempty"`
+
+	// +kubebuilder:validation:MinItems=1
 	Template []AppTemplate `json:"template,omitempty"`
-	Deploy   []AppDeploy   `json:"deploy,omitempty"`
+
```
ewrenn8: Curious what error message something like this gives? I know we have some specific error messages for the case of empty templates/deploys/fetches, so if those are better or more informative, may be we leave off the validations in the schema?

pkg/apis/kappctrl/v1alpha1/types.go
```diff
@@ -9,15 +9,22 @@ import (
...
 	metav1.TypeMeta `json:",inline"`
 
 	// Standard object metadata; More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata.
 	// +optional
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   AppSpec   `json:"spec"`
-	Status AppStatus `json:"status"`
+	// +optional
+	Spec AppSpec `json:"spec,omitempty"`
```
ewrenn8: Should AppSpec be required?

config/install.package.carvel.dev_installedpackages.yaml
```diff
@@ -0,0 +1,149 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.3.0
```
cppforlife: do we want these annotations here?

config/install.package.carvel.dev_installedpackages.yaml
```diff
@@ -0,0 +1,149 @@
...
+            version:
+              type: string
+          type: object
+      type: object
+  version: v1alpha1
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+status:
```
cppforlife, line 152: this is going to make kapp unhappy i believe. its stating that i want status to always be set to empty. (ideally we would also not have `metadata.creationTimestamp: null`)

config/install.package.carvel.dev_installedpackages.yaml
```diff
@@ -0,0 +1,149 @@
...
+    listKind: InstalledPackageList
+    plural: installedpackages
+    shortNames:
+    - ipkg
+    singular: installedpackage
+  scope: Namespaced
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
```
cppforlife, line 41: should we be setting spec.preserveUnknownFields=false somewhere here? (https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/)

hack/gen-crds.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# Using v0.3.0 of controller-gen
+# Install: go get -u sigs.k8s.io/controller-tools/cmd/controller-gen@v0.3.0
```
cppforlife: we got https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/hack/tools.go to record tools depenencies. check out other generation scripts for where binary is.

pkg/apis/kappctrl/v1alpha1/types.go
```diff
@@ -66,49 +91,100 @@ type AppClusterKubeconfigSecretRef struct {
...
+	// +optional
+	Stderr string `json:"stderr,omitempty"`
+
+	// +optional
+	ExitCode int `json:"exitCode"`
+
+	// +optional
+	Error string `json:"error,omitempty"`
+
+	// +optional
```
cppforlife, line 153: nit: do we want to have newline separation between fields (seems like kubernetes doesnt: https://github.com/kubernetes/api/blob/master/core/v1/types.go)?

pkg/apis/kappctrl/v1alpha1/types.go
danielhelfand: Yeah, it's a good call out. The error from this would be worse:

```
App.kappctrl.k14s.io "simple-app" is invalid: spec.deploy:
    Invalid value: "null": spec.deploy in body must be of type array: "null" (reason: Invalid)
```

versus

```
 Expected exactly one deploy option
```

An advantage would be that the App itself is not created if it is invalid, but maybe we should preserve the behavior for now and weigh whether it would be worthwhile to enforce this via schema.

pkg/apis/kappctrl/v1alpha1/types.go
danielhelfand: Not sure what best practice is here and was wondering myself. I think what you are suggesting makes sense, but I mainly did this because I was following what other projects had done (e.g. velero, knative).

config/install.package.carvel.dev_installedpackages.yaml
danielhelfand: They are autogenerated but can look into if they can be excluded/have use.

pkg/apis/kappctrl/v1alpha1/types.go
cppforlife: error above is for when you specify `deploy: null`. i think what we should be comparing this to is `deploy: []`

config/install.package.carvel.dev_installedpackages.yaml
danielhelfand: Added

pkg/apis/kappctrl/v1alpha1/types.go
danielhelfand: Removed

pkg/apis/kappctrl/v1alpha1/types.go
danielhelfand: Made spec required

config/install.package.carvel.dev_installedpackages.yaml
danielhelfand: There does not appear to be a way to generated with them. I have manually removed for now and maybe we can request this upstream to have an option not to include annotations.

config/install.package.carvel.dev_installedpackages.yaml
danielhelfand: Manually removed both and documented steps for after crd generation.

hack/gen-crds.sh
danielhelfand: They don't really offer a great way to temporarily install and run, so I vendored in and have a somewhat hacky workaround to not require user to have controller-gen installed.

pkg/apis/kappctrl/v1alpha1/types.go
danielhelfand: Yeah, good call. It actually is a bit nicer, so maybe it would be worthwhile. Personally prefer the error kapp-controller currently gives, but the message below is okay:

```
App.kappctrl.k14s.io "simple-app" is invalid: spec.deploy:
      Invalid value: 0: spec.deploy in body should have at least 1 items (reason: Invalid)
```

cppforlife:
im guessing we need to update this PR because https://github.com/vmware-tanzu/carvel-kapp-controller/pull/178 was merged.

cppforlife:
@danielhelfand do we need to update this PR with changes for new Package CR fields?

danielhelfand:
> @danielhelfand do we need to update this PR with changes for new Package CR fields?

It looks like `controller-gen` wasn't generating the InternalPackage CRD correctly cause of how the type was defined. It was as follows:

```go
type InternalPackage packages.Package
```

But changed to as follows to get it working:

```go
// +kubebuilder:subresource:status
// +kubebuilder:resource:scope=Cluster
type InternalPackage struct {
	metav1.TypeMeta `json:",inline"`
	// Standard object metadata; More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata.
	// +optional
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec packages.PackageSpec `json:"spec"`
	// +optional
	Status packages.PackageStatus `json:"status,omitempty"`
}
```

PR #185, opened 2021-04-30
Guard against events that have nil objects

ewrenn8:
* After some period of time, this watcher gets an event with a nil
  objet and causes a crash.


danielhelfand (commented):
lgtm

PR #190, opened 2021-05-06
Allow Users to Provide Schema for Package Values

### LINKED ISSUE #104: Package Authors should be able to provide openAPI schema for configurations in Package CR
As a Package Consumer, I need to know the details about the configuration that is accepted by a package and I should be able to have confidence that the values provided by me are in the correct format and valid.

This problem has been divided into three issues:
- Package consumers do not have an easy way to discover the list of configurations that are provided by a package (this issue).
- While installing a package, when Package consumers provide their own configuration values of a package, they should get actionable error messages when the provided configuration is not valid.
    - kapp-controller should do basic structure validation of values provided in InstalledPackage (Issue #148).
    - If Package authors use ytt schemas in their package, kapp-controller should pass on the configuration in the ytt templating step so that ytt can validate the provided configuration (Issue #149).

- The Package Author should be able to define the schema for their configuration in their package CR. 

- The Package Consumer should be able to view the schema for the configuration of a package so that they know what is the specific format for each configuration.

Short-Term:
- The configuration should be easily discoverable. Currently, Package Consumer will need to download the package bundle and check the data values YAML file to even know what configurations are exposed.
- Learn what type each configuration is
- Learn what is the purpose of each configuration is
- Learn what is the default of each configuration

Long-Term (Post MVP):
- Associated html type with each configuration. How should it be displayed on a UI form?

- [ ] This feature should come with documentation. Also note we should explicitly call out schemas should be included in the relevant templating steps and that kapp-controller, at least initially, won't do any validation using the schema included in the package definition; That schema is intended mainly for visibility in to what values can and should be provided.

danielhelfand:
Closes #104 

This pull request adds the property `valuesSchema` to Packages to allow users to provide an open API schema for data values that can be configured by users during installation. Since the implementation currently is represented as `map[string]string`, users can provide any key to access a schema like below:

```yaml
---
apiVersion: package.carvel.dev/v1alpha1
kind: Package
metadata:
  name: simple-app.corp.com.1.0.0
spec:
  publicName: simple-app.corp.com
  version: 1.0.0
  valuesSchema:
    openAPIV3JSON: |
      {...}
```

We can add further validation or change the `map[string]string` to a struct to restrict what keys are allowed if we would like and will supplement this feature with docs to recommend ways to specify key:value pairs for adding a schema. 

As mentioned in #104, validation of values with this schema is out of scope. Additionally, users will have to manually declare these value schemas until automation is available from tools like `ytt` to generate/add these schemas to Packages.

### DIFF
--- a/pkg/apiserver/apis/packages/types.go
+++ b/pkg/apiserver/apis/packages/types.go
@@ -6,6 +6,7 @@ package packages
 import (
 	kcv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
 // +genclient
@@ -57,7 +58,8 @@ type PackageSpec struct {
 	// can be configured by users when a Package is installed
 	// in an OpenAPI schema format.
 	// +optional
-	ValuesSchema map[string]string `json:"valuesSchema,omitempty"`
+	// +kubebuilder:pruning:PreserveUnknownFields
+	ValuesSchema runtime.RawExtension `json:"valuesSchema,omitempty"`
 }
 
 type Maintainer struct {
--- a/pkg/apiserver/apis/packages/types.go
+++ b/pkg/apiserver/apis/packages/types.go
@@ -58,8 +58,7 @@ type PackageSpec struct {
 	// can be configured by users when a Package is installed
 	// in an OpenAPI schema format.
 	// +optional
-	// +kubebuilder:pruning:PreserveUnknownFields
-	ValuesSchema runtime.RawExtension `json:"valuesSchema,omitempty"`
+	ValuesSchema ValuesSchema `json:"valuesSchema,omitempty"`
 }
 
 type Maintainer struct {
@@ -73,4 +72,10 @@ type AppTemplateSpec struct {
 type PackageStatus struct {
 	ObservedGeneration int64                     `json:"observedGeneration"`
 	Conditions         []kcv1alpha1.AppCondition `json:"conditions"`
+}
+
+type ValuesSchema struct {
+	// +optional
+	// +kubebuilder:pruning:PreserveUnknownFields
+	OpenAPISchemaV3 runtime.RawExtension `json:"openAPISchemaV3,omitempty"`
 }
--- a/pkg/apiserver/apis/packages/types.go
+++ b/pkg/apiserver/apis/packages/types.go
@@ -76,6 +76,7 @@ type PackageStatus struct {
 
 type ValuesSchema struct {
 	// +optional
+	// +nullable
 	// +kubebuilder:pruning:PreserveUnknownFields
-	OpenAPISchemaV3 runtime.RawExtension `json:"openAPISchemaV3,omitempty"`
+	OpenAPIv3 runtime.RawExtension `json:"openAPIv3,omitempty"`
 }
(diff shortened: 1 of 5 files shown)


cppforlife:
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/ uses actual structure to represent schema: 

```yaml
      schema:
        openAPIV3Schema:
          type: object
          properties:
```

any reason why we need to "stringify it"?

danielhelfand:
> https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/ uses actual structure to represent schema:
> 
> ```yaml
>       schema:
>         openAPIV3Schema:
>           type: object
>           properties:
> ```
> 
> any reason why we need to "stringify it"?

We could, but we would need to think about how we generate our CRDs with `controller-gen`. If we were to add more structure like something like below it will break how the CRD schemas are generated due to our use of `apiVersion: apiextensions.k8s.io/v1beta1`:

```go
type ValuesSchema struct {
	openAPIV3Schema v1beta1.JSONSchemaProps `json:"openAPIV3Schema"`
}
```

We would need to consider upgrading many dependencies, including most of our main Kubernetes deps.

ewrenn8:
I think this also depends on what form ytt is planning to output the schema. As an author, I wouldnt want to have to do ytt schema ==> json/openapi schema JSON ==> json/openapi schema YAML. That extra step to convert openapi stuff would be pretty annoying.

Also, is the schema kubernetes code uses fully compatible with all the potential outputs from ytt? I thought their schema implementation has some explicit rules that aren't in openapi. We may be able to use some other impl of openapi in this case, though.

danielhelfand:
> I think this also depends on what form ytt is planning to output the schema. As an author, I wouldnt want to have to do ytt schema ==> json/openapi schema JSON ==> json/openapi schema YAML. That extra step to convert openapi stuff would be pretty annoying.

The plan is to support openAPI/JSONSchema formats from `ytt`: https://github.com/vmware-tanzu/carvel-ytt/issues/357

> Also, is the schema kubernetes code uses fully compatible with all the potential outputs from ytt? I thought their schema implementation has some explicit rules that aren't in openapi. We may be able to use some other impl of openapi in this case, though.

The `JSONSchemaProps` is what is used to generate the schema for the CRDs by `controller-gen`, but am not sure if it is compatible.

cppforlife:
> That extra step to convert openapi stuff would be pretty annoying.

that could be a flag on ytt. (arguably may be we should output as yaml by default?)

> Also, is the schema kubernetes code uses fully compatible with all the potential outputs from ytt?

all json is valid yaml so structure wise speaking it should be fine. i didnt mean that we actually needed to use kubernetes provided openapi libraries, what i meant is that it shouldnt be a string value but rather a structure. @danielhelfand isnt there a think like RawExtensions or something like that effectively allows to put anything in the value?

danielhelfand:
> @danielhelfand isnt there a think like RawExtensions or something like that effectively allows to put anything in the value?

I have changed the ValuesSchema property to be `runtime.RawExtension` and having the InternalPackage schema allow unknown fields for the property. This should allow for unstructured data to be passed.

cppforlife:
we still probably have a subkey `openAPIV3Schema` under `valuesSchema` so that if we can support in future other types of schema (if we need to).

pkg/apiserver/apis/packages/types.go
```diff
@@ -68,3 +73,9 @@ type PackageStatus struct {
 	ObservedGeneration int64                     `json:"observedGeneration"`
 	Conditions         []kcv1alpha1.AppCondition `json:"conditions"`
 }
+
+type ValuesSchema struct {
+	// +optional
+	// +kubebuilder:pruning:PreserveUnknownFields
+	OpenAPISchemaV3 runtime.RawExtension `json:"openAPISchemaV3,omitempty"`
```
cppforlife: should we change it to OpenAPIv3 (openAPIv3)? schema seems a bit redundant.

danielhelfand: Changed to `openAPIv3`

PR #191, opened 2021-05-06
fix typo

2muchgit:
Just a typo :D

### DIFF
--- a/cmd/controller/handlers/configmap_handler.go
+++ b/cmd/controller/handlers/configmap_handler.go
@@ -15,7 +15,7 @@ import (
 
 type ConfigMapHandler struct {
 	log             logr.Logger
-	appRefTacker    *reftracker.AppRefTracker
+	appRefTracker    *reftracker.AppRefTracker
 	appUpdateStatus *reftracker.AppUpdateStatus
 }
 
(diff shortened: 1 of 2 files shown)


danielhelfand (commented):
Thanks for finding this, @2muchgit. We would also need to update the following lines in configmap_handler.go:
* https://github.com/vmware-tanzu/carvel-kapp-controller/blob/46df5dae94c7e2c89f2eb7a32223450b0a16c2e2/cmd/controller/handlers/configmap_handler.go#L38
* https://github.com/vmware-tanzu/carvel-kapp-controller/blob/46df5dae94c7e2c89f2eb7a32223450b0a16c2e2/cmd/controller/handlers/configmap_handler.go#L44

We also have this same typo in secret_handler.go if you are interested in fixing that:
* https://github.com/vmware-tanzu/carvel-kapp-controller/blob/46df5dae94c7e2c89f2eb7a32223450b0a16c2e2/cmd/controller/handlers/secret_handler.go#L18

2muchgit:
Sure thing!

2muchgit:
I think that should be it. I grepped further through the repository for any other instances of the typo "appRefTacker" and it seems to be corrected now. Thanks for the quick review and I am glad I could be of some help.

PR #206, opened 2021-05-19
Keep valid crd definitions in repo

ewrenn8:


hack/gen-crds.sh
```diff
@@ -4,7 +4,13 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+rm -rf crds-tmp
```
cppforlife: should this be tmp/crds?

PR #208, opened 2021-05-25
Regroup packaging CRs

ewrenn8:


ewrenn8:
Making draft until #207 is merged. Part of that PR moves the pause test to testing against an app, so I skipped updating that test in this pr and instead just did a `t.Skip()`. Once validations PR is merged, I will rebase and unskip.

PR #210, opened 2021-05-27
Prevent double fire of github action on PRs

joaopapereira:


.github/workflows/golangci-lint.yml
```diff
@@ -2,6 +2,8 @@ name: golangci-lint
 
 on:
   push:
+    branches:
```
danielhelfand, line 5: Would this prevent CI from running for other pushes to non-develop branches?

joaopapereira: Yes, are you actively using other branches that need tests to run on per push?

danielhelfand: No, just more thinking if we would want to stop running CI on pushes to branches without open pull requests to develop. 

But that's not an uncommon practice and our testing/linting isn't too difficult to run locally. So I would be all right with this.

joaopapereira: I think that if this is a long-running branch that we want to keep tests we can add that branch to the `branches` field in these 2 files for it to run tests and linting.

PR #230, opened 2021-06-14
Supplement hack/gen-crds.sh with overlay for internalpackagemetadata

danielhelfand:
Since the autogenerated clients use [internalpackagemetadatas](https://github.com/vmware-tanzu/carvel-kapp-controller/blob/c7eb79b7d62aad63339928d561081898c4918352/pkg/client/clientset/versioned/typed/internalpackaging/v1alpha1/internalpackagemetadata.go#L55) as plural, crds need s added to be consistent.

By default, `controller-gen` has `internalpackagemetadata` as the plural name of the CRD so this preserves the automation of the script.

### DIFF
--- a/.github/workflows/test-gh.yml
+++ b/.github/workflows/test-gh.yml
@@ -43,6 +43,8 @@ jobs:
         mkdir /tmp/bin
         export PATH=/tmp/bin:$PATH
 
+        ./hack/verify-no-dirty-files.sh
+
         wget -O- https://github.com/kubernetes/minikube/releases/download/v1.10.0/minikube-linux-amd64 > /tmp/bin/minikube
         echo "9d34cb50bc39f80d39f92d1fb7cb23a271504b519f5e805574894d395ce3e7b3  /tmp/bin/minikube" | sha256sum -c -
         chmod +x /tmp/bin/minikube
--- a/config/crd-overlay.yml
+++ b/config/crd-overlay.yml
@@ -18,3 +18,11 @@ spec:
 #@overlay/remove
 #@overlay/match missing_ok=True
 status:
+
+#@overlay/match by=overlay.subset({"metadata":{"name":"internalpackagemetadata.internal.packaging.carvel.dev"}}), expects="0+"
+---
+metadata:
+  name: internalpackagemetadatas.internal.packaging.carvel.dev
+spec:
+  names:
+    plural: internalpackagemetadatas
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -13,6 +13,7 @@ go mod tidy
 go build $repro_flags -mod=vendor -o controller ./cmd/main.go
 ls -la ./controller
 
+./hack/gen-crds.sh
 ytt -f config/ >/dev/null
 
 echo SUCCESS
--- a/hack/crd-overlay.yml
+++ b/hack/crd-overlay.yml
@@ -19,7 +19,7 @@ spec:
 #@overlay/match missing_ok=True
 status:
 
-#@overlay/match by=overlay.subset({"metadata":{"name":"internalpackagemetadata.internal.packaging.carvel.dev"}}), expects="0+"
+#@overlay/match by=overlay.subset({"metadata":{"name":"internalpackagemetadata.internal.packaging.carvel.dev"}})
 ---
 metadata:
   name: internalpackagemetadatas.internal.packaging.carvel.dev
--- a/hack/deploy-test.sh
+++ b/hack/deploy-test.sh
@@ -3,4 +3,3 @@
 set -e
 
 ./hack/build.sh && ytt -f config/ -f config-test/ | kbld -f- | kapp deploy -a kc -f- -c -y
-
--- a/hack/gen-crds.sh
+++ b/hack/gen-crds.sh
@@ -11,6 +11,6 @@ go run ./vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go \
   output:dir=./tmp/crds \
   paths=./pkg/apis/...
 
-ytt -f tmp/crds -f config/crd-overlay.yml > config/crds.yml
+ytt -f tmp/crds -f ./hack/crd-overlay.yml > config/crds.yml
 
 rm -rf tmp/crds
--- a/hack/verify-no-dirty-files.sh
+++ b/hack/verify-no-dirty-files.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+./hack/build.sh
+
+if ! git diff --exit-code >/dev/null; then
+  echo "Error: Running ./hack/build.sh resulted in non zero exit code from git diff. Please run './hack/build.sh' and 'git add' the generated file(s)."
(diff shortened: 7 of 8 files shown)


config/crd-overlay.yml
```diff
@@ -18,3 +18,13 @@ spec:
...
 status:
+
+#@overlay/match by=overlay.subset({"metadata":{"name":"internalpackagemetadata.internal.packaging.carvel.dev"}}), expects="0+"
+---
+metadata:
+  #@overlay/replace
+  name: internalpackagemetadatas.internal.packaging.carvel.dev
+spec:
+  names:
+    #@overlay/replace
```
cppforlife: no need for replace since it's a scalar.

config/crd-overlay.yml
```diff
@@ -18,3 +18,13 @@ spec:
 #@overlay/remove
 #@overlay/match missing_ok=True
 status:
+
+#@overlay/match by=overlay.subset({"metadata":{"name":"internalpackagemetadata.internal.packaging.carvel.dev"}}), expects="0+"
+---
+metadata:
+  #@overlay/replace
```
cppforlife: no need for replace since it's a scalar.

config/crd-overlay.yml
danielhelfand: Thanks, addressed

config/crd-overlay.yml
danielhelfand: Thanks, addressed

config/crd-overlay.yml
```diff
@@ -18,3 +18,11 @@ spec:
 #@overlay/remove
 #@overlay/match missing_ok=True
 status:
+
+#@overlay/match by=overlay.subset({"metadata":{"name":"internalpackagemetadata.internal.packaging.carvel.dev"}}), expects="0+"
```
ewrenn8: Do we wanna do expects=1 here so that we blow up if we ever match multiple CRDs?

danielhelfand: We don't generate the crds as part of ./hack/deploy-test.sh, so it would fail since the generated crds are already produced.

ewrenn8: Thoughts on generating them as part of deploy-test? I go back and forth here since it would be a good way to ensure crds are regenerated when changes are made and to ensure you're testing with the same deployment that would be included in a release, but I can also see arguments against generating them there.

danielhelfand: I went ahead and added a step to check if generation has happened. This means that the overlay will be used strictly for purposes of generating the crds and not as part of the actual deploy process. The pro for this would be having more control with expects=1 for the internalpackagemetadata specific overlay. The con is knowing we need to exclude it for all the places ytt uses it. Alternatively, we could add the overlay to a separate directory to keep it separate from config folder.

hack/verify-gen-crds.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+./hack/gen-crds.sh
+
+if ! git diff --exit-code config/crds.yml >/dev/null; then
+  echo "CRD verification - failed! Generated CRDs are out of date. Please run './hack/gen-crds.sh' and 'git add' the generated file(s)."
+  exit 1
+fi
```
cppforlife: i think this should happen only in ci with a failure. it's a little weird for the script to gen things, then fail, to expect you to git add, then you have to rerun it.

danielhelfand: Thanks, moved to test-gh workflow

hack/build.sh
```diff
@@ -13,6 +13,6 @@ go mod tidy
 go build $repro_flags -mod=vendor -o controller ./cmd/main.go
 ls -la ./controller
 
-ytt -f config/ >/dev/null
+ytt -f config/ --file-mark 'crd-overlay.yml:exclude=true' >/dev/null
```
cppforlife: why do we need to exclude this file? is it because it's used only during generation of crds.yml? if so, i think we should move it to a different location (like hack directory next to the script that it uses it?)

danielhelfand: Yes, mentioned in comment above. As of now, overlay is only used for supplementing the generation of crds. I have moved to hack folder for now and removed file-marks. Let me know if there is a more logical place for it though.

hack/verify-gen-crds.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+./hack/gen-crds.sh
```
cppforlife: should we still do this line in build?

hack/verify-gen-crds.sh
```diff
@@ -0,0 +1,10 @@
...
+
+if ! git diff --exit-code config/crds.yml >/dev/null; then
```
cppforlife: should we expand scope of the check? e.g. why are any of the files dirty in git at the end of the build process for example.

.github/workflows/test-gh.yml
```diff
@@ -43,6 +43,8 @@ jobs:
         mkdir /tmp/bin
         export PATH=/tmp/bin:$PATH
 
+        ./hack/verify-no-dirty-files.sh
```
cppforlife, line 46: this probably need to be at the end, since nothing has happened that would have resulted in dirty git.

danielhelfand: `verify-no-dirty-files` is running `./hack/build.sh` so we should see diff. FWIW, it's failing now in CI cause there appears to be a vendor diff: https://github.com/vmware-tanzu/carvel-kapp-controller/runs/2842176794?check_suite_focus=true

cppforlife: ah i totally missed it. didnt expect that verify-* script was going to do a build based on the name, but fair enough.

hack/verify-gen-crds.sh
danielhelfand: There appears to be an issue when running `git diff -exit-code` in CI  on the whole project. It says `gnostic/OpenAPIv2` is being deleted. It may be related to our use of `replace github.com/googleapis/gnostic => github.com/googleapis/gnostic v0.4.1` in go.mod. We point gnostic 0.4.1 when [OpenAPIv2 was renamed to openapiv2](https://github.com/google/gnostic/pull/155). What is vendored in on develop is [OpenAPIv2](https://github.com/vmware-tanzu/carvel-kapp-controller/tree/7896992649b3247d284cca126cc488479dc120d5/vendor/github.com/googleapis/gnostic/OpenAPIv2). 

Not quite sure how to resolve the issue. For now, I have gone back to just checking that crds.yml is generated. Unless there is something simple I am missing, I would prefer to move forward with just checking crds.yml as this pr is mainly focused on fixing crd generation. I can open a separate issue for resolving the issue in CI mentioned above.

danielhelfand: Looks like this was a [vendor issue](https://github.com/vmware-tanzu/carvel-kapp-controller/pull/240). Hopefully this pr can catch these kinds of issues moving forward. I have gone ahead and added back the more comprehensive check.

PR #234, opened 2021-06-15
Validate Package version is valid semver

ewrenn8:


pkg/apiserver/apis/datapackaging/validation/validations.go
```diff
@@ -67,6 +69,13 @@ func ValidatePackageSpecVersion(version string, fldPath *field.Path) field.Error
 			field.Invalid(fldPath, version, "cannot be empty"))
 	}
 
+	// taken from https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
+	const semverRegex = `^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`
```
cppforlife: we probably should use vendir's https://github.com/vmware-tanzu/carvel-vendir/blob/develop/pkg/vendir/versions/v1alpha1/semvers.go#L23 (might have to expose a new public method to parse single version without a v stripping). this way our validation will be inline with down the line parsing of semvers when we are sorting.

PR #235, opened 2021-06-16
Backfill test for PackageInstall using already existing App with same name

### LINKED ISSUE #184: InstalledPackage should take over the existing App CR

Currently, we use App CR to deploy core addons like CNI and CPI. We want to migrate these core addons to packaging API. 

During the cluster upgrade, core addons like CNI and CPI should continue to work without interrupt. That is when we apply the InstalledPackge in the upgraded cluster, the old App CR should be seamlessly taken over by the newly applied InstalledPackage.

InstalledPackage should take over the existing App CR if they have the same namespaced name.

---
Vote on this request

danielhelfand:
Backfilling a test for the behavior of PackageInstalls using an App that already exists in PackageInstall namespace that has same name.

### DIFF
--- a/test/e2e/packageinstall_test.go
+++ b/test/e2e/packageinstall_test.go
@@ -338,3 +338,134 @@ spec:
 		kubectl.Run([]string{"delete", "pkgi", name, "--wait=true"})
 	})
 }
+
+func Test_PackageInstall_UsesExistingAppWithSameName(t *testing.T) {
+	env := BuildEnv(t)
+	logger := Logger{}
+	kapp := Kapp{t, env.Namespace, logger}
+	kubectl := Kubectl{t, env.Namespace, logger}
+	sas := ServiceAccounts{env.Namespace}
+	name := "pkg-instl-uses-app"
+
+	appYaml := fmt.Sprintf(`
+---
+apiVersion: kappctrl.k14s.io/v1alpha1
+kind: App
+metadata:
+  name: %s
+  annotations:
+    kapp.k14s.io/change-group: kappctrl-e2e.k14s.io/apps
+spec:
+  serviceAccountName: kappctrl-e2e-ns-sa
+  fetch:
+    - inline:
+        paths:
+          file.yml: |
+            apiVersion: v1
+            kind: ConfigMap
+            metadata:
+              name: configmap
+  template:
+  - ytt: {}
+  deploy:
+    - kapp: {}
+`, name) + sas.ForNamespaceYAML()
+
+	pkginstallYaml := fmt.Sprintf(`---
+apiVersion: data.packaging.carvel.dev/v1alpha1
+kind: PackageMetadata
+metadata:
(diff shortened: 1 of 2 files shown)


test/e2e/packageinstall_test.go
```diff
@@ -338,3 +338,108 @@ spec:
...
+	cleanUp()
+	defer cleanUp()
+
+	// Create App
+	kubectl.RunWithOpts([]string{"apply", "-f", "-"}, RunOpts{StdinReader: strings.NewReader(appYaml)})
+
+	// Create PackageInstall
+	kapp.RunWithOpts([]string{"deploy", "-a", name, "-f", "-"}, RunOpts{StdinReader: strings.NewReader(pkginstallYaml)})
+	kubectl.Run([]string{"wait", "--for=condition=ReconcileSucceeded", "pkgi/" + name, "--timeout", "1m"})
+
```
cppforlife, line 438: i think we need stronger assertion that App CR is "taken over". how about making Package CR contain some different in spec.template and checking that App CR has that change after PackageInstall is reconciled?

test/e2e/packageinstall_test.go
```diff
@@ -338,3 +338,108 @@ spec:
...
+	cleanUp()
+	// Assert that App CR is deleted as a result of PackageInstall
+	// with same name as App being deleted. Since the App was created
+	// first without the PackageInstall, the result of the PackageInstall
+	// using the existing App should be that it will be deleted when PackageInstall
+	// gets deleted.
+	out, err := kubectl.RunWithOpts([]string{"get", fmt.Sprintf("apps/%s", name)}, RunOpts{AllowError: true})
+	if err == nil {
+		t.Fatalf("Expected no App to be found after PackageInstall is deleted\nGot: %s", out)
+	}
```
cppforlife: probably a good idea to assert on error kubectl returned (some not found bit).

test/e2e/packageinstall_test.go
```diff
@@ -338,3 +338,108 @@ spec:
...
+	cleanUp := func() {
+		kapp.Run([]string{"delete", "-a", name})
+	}
+	cleanUp()
+	defer cleanUp()
+
+	// Create App
+	kubectl.RunWithOpts([]string{"apply", "-f", "-"}, RunOpts{StdinReader: strings.NewReader(appYaml)})
+
+	// Create PackageInstall
```
cppforlife: instead of comments like this we can use logger.Section(... (check out examples of it in other tests)

test/e2e/packageinstall_test.go
danielhelfand: Sure, I am going to recommend keeping #184 open after feedback is implemented though as I still think this behavior of PackageInstalls taking over Apps is not intuitive as far as UX is concerned. But I agree that the assertion can be made stronger as far as preserving existing behavior.

Issue #237, opened 2021-06-16
Minor changes for Packaging APIs MVP


vibhas:
**Describe the problem/challenge you have**
We identified several small changes that need to be done for the packaging APIs MVP.

- [x]  Bump imgpkg to the version with performance enhancements
- [x]  Update examples with correct packaging concepts
- [x]  Run kbld on package repository bundle used in e2e tests (currently the imageslock is empty)
- [x]  Refine delete finalizer names for pkgrepo and pkginstall
- [x]  PackageRepository CRD should have annotation with the namespace for global packages
- [x]  Update PackageInstall's packageVersionRef => packageRef
- [x]  Change Package's spec.releasedAt to metav1.Time

Some things that still need more consideration:
- [x]  Should we add --wait-timeout=30s for pkg repo deploys? (seems like we don't ever wait it to get stuck for 15min) + --kube-api-qps/--kube-api-burst
- [ ]  Should we add waiting behavior to kapp for package repos/installs?

ewrenn8:
waiting on new imgpkg release

aaronshurley:
> Should we add waiting behavior to kapp for package repos/installs?
Extracting this to its own issue in kapp: https://github.com/vmware-tanzu/carvel-kapp/issues/237

ewrenn8:
Included in `v0.20.0`

PR #247, opened 2021-06-22
Change releasedAt from string to metav1.Time

danielhelfand:
Part of #237

Change Package's spec.releasedAt to metav1.Time


danielhelfand:
Test failures seem to be related to needing to make `releasedAt` nullable in schema.

PR #249, opened 2021-06-22
Revert k8s deps

danielhelfand:
In #241 k8s deps were updated as a result of updating controller-tools, but we only really need to update the controller-tools dep. This pull request uses replace directives to revert back to previous versions of k8s deps to avoid cluttering logs kapp-controller logs with output like below:

```
E0622 17:40:19.916955      14 reflector.go:138] k8s.io/client-go@v0.20.2/tools/cache/reflector.go:167: Failed to watch *v1beta1.FlowSchema: failed to list *v1beta1.FlowSchema: the server could not find the requested resource (get flowschemas.flowcontrol.apiserver.k8s.io)
E0622 17:40:47.629070      14 reflector.go:138] k8s.io/client-go@v0.20.2/tools/cache/reflector.go:167: Failed to watch *v1beta1.PriorityLevelConfiguration: failed to list *v1beta1.PriorityLevelConfiguration: the server could not find the requested resource (get prioritylevelconfigurations.flowcontrol.apiserver.k8s.io)
```

These logs appear for Kubernetes clusters below v1.19 since the resources above are not considered beta until 1.20. Thus client-go does not run into these issues in v0.19.0. 

We can also remove RBAC for `flowcontrol.apiserver.k8s.io` with this change.


go.mod
```diff
@@ -23,4 +23,13 @@ require (
...
+replace (
+	github.com/googleapis/gnostic => github.com/googleapis/gnostic v0.4.1
+	golang.org/x/crypto => golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
+	k8s.io/api => k8s.io/api v0.19.2
+	k8s.io/apimachinery => k8s.io/apimachinery v0.19.2
+	k8s.io/apiserver => k8s.io/apiserver v0.19.2
+	k8s.io/client-go => k8s.io/client-go v0.19.2
+	k8s.io/code-generator => k8s.io/code-generator v0.19.2
+	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200805222855-6aeccd4b50c6
+)
```
cppforlife: hmm, is there no way to do this without replace? replace afaik is only meant to be used for development.

ewrenn8: Its a pattern ive seen in k8s libraries as well https://github.com/kubernetes/apiserver/blob/3492629d2b9bafd6ac5e6cdf7521a5c9d2e84227/go.mod#L53-L58. Open to other ways, though.

danielhelfand: We could also use `sigs.k8s.io/controller-tools v0.4.1` instead of `v0.5.0`. That's where it seems we run into this unintended dep bumping issue. Would provide same exact features of generating v1 crd. Can try it out locally.

hack/crd-overlay.yml
```diff
@@ -8,6 +8,16 @@ metadata:
   #@overlay/remove
   #@overlay/match missing_ok=True
   annotations:
+spec:
+  versions:
+  #@overlay/match by=overlay.index(0)
```
danielhelfand: Open to thoughts here on if there is a better way to match. 

We need the overlay since it seems `description` for metadata causes issues: https://github.com/kubernetes-sigs/controller-tools/pull/511.

cppforlife: should we do `overlay.all,expects="0+"` since this would be true for all versions.

PR #254, opened 2021-06-29
speed up Package installation in PackageRepository CR

cppforlife:


pkg/pkgrepository/package_repo_app.go
```diff
@@ -52,6 +52,14 @@ func NewPackageRepoApp(pkgRepository *pkgingv1alpha1.PackageRepository) (*kcv1al
 #@overlay/remove
 ---
 
+#@overlay/match by=overlay.or_op(pkg, pkgm, pkgv),expects="0+"
```
danielhelfand, line 55: Any reason to not remove `pkgv` entirely for GA?

PR #258, opened 2021-06-30
Backfill PackageInstall Upgrade Test and Fix Test_PackageRepoDelete

danielhelfand:
This pull request backfills a test for PackageInstalls upgrading from one Package version to another. 

It also makes changes to `Test_PackageRepoDelete` as documented below:
* PackageRepo creation may have slightly slower performance with addition of b380458 so bumping timeout for test
* In event checking if Packages exist fails, kapp delete may not be run for test. Added to cleanUp() func.

This pull request changes minikube version to 1.21 to address earlier issues with deploying kapp-controller in CI: https://github.com/vmware-tanzu/carvel-kapp-controller/runs/2956127108?check_suite_focus=true

### DIFF
--- a/.github/workflows/test-gh.yml
+++ b/.github/workflows/test-gh.yml
@@ -45,8 +45,8 @@ jobs:
 
         ./hack/verify-no-dirty-files.sh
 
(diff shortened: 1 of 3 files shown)


ewrenn8:
This lgtm

PR #285, opened 2021-07-08
Jk linters for godoc

joe-kimmel-vmw:
This is a draft / POC for enforcing godoc comments on Exports going forward.  It doesn't complain about existing code, but it should complain about new code,  so it'll force us to slowly ratchet up our commenting game.   

At the moment I'm not 100% sure how it handles changes; I changed the name of a receiver on an exported but uncommented function, and the linter did not flag that,  so unclear how 'big' the change has to be to trigger the linter.

Of slight note is that it's complaining about some existing decoration/metadata patterns that start with `// +` e.g. 
`// +groupName=packaging.carvel.dev` but i'm sure there's a way to preserve those while adding Godocs.

Sample output of how it complains:
```
pkg/apis/packaging/v1alpha1/register.go:30:1: exported: exported function Resource should have comment or be unexported (revive)
func Resource(resource string) schema.GroupResource {
^
cmd/controllerinit/run.go:22:2: exported: exported const InternalControllerFlag should have comment (or a comment on this block) or be unexported (revive)
	InternalControllerFlag = "internal-controller"
	^
[...]
pkg/apis/kappctrl/v1alpha1/types_fetch.go:34:1: exported: comment on exported type AppFetchImage should be of the form "AppFetchImage ..." (with optional leading article) (revive)
// +k8s:openapi-gen=true
[...]
pkg/apis/kappctrl/v1alpha1/types_template.go:51:1: exported: comment on exported type AppTemplateKustomize should be of the form "AppTemplateKustomize ..." (with optional leading article) (revive)
// TODO implement kustomize
```


joe-kimmel-vmw:
interestingly the version of golangci-lint being run doesn't seem to recognize the revive linter,  whereas the newer version i installed indicates that golint is deprecated and we should replace it with revive.

ewrenn8:
I think we are a few versions out of date on the linter. I dont think there is any reason we are tied to the version its currently using, so feel free to bump that if it is needed.

joe-kimmel-vmw:
this is hitting some snags with the golangci-lint version and revive, but that's really orthogonal to the actual go of enforcing godoc via linting,  so i'm going to cherry-pick the relevant bit into a new branch and abandon this one.

PR #286, opened 2021-07-08
add mac-install-deps.sh for automating the fetch/instal of carvel too…

joe-kimmel-vmw:
…ls required by kapp-controller for a mac dev env


ewrenn8:
Is there a way we could have this always dynamically pull the versions to install from the Dockerfile? I could definitely see this falling out of date as we bump the tools in the Dockerfile and forget to regenerate this script. Alternatively, we could script the bumping of tools versions and have that script update this one, but that seems more difficult.

joe-kimmel-vmw:
> Is there a way we could have this always dynamically pull the versions to install from the Dockerfile? 

likely!!   Daniel and I tried wrapping it in an `eval $()` but it couldn't find chmod?  which was weird.  I can keep playing with it;  agree that "one more thing to maintain"  is a cost.

hack/mac-install-deps.sh
```diff
@@ -0,0 +1,12 @@
+#! /bin/bash
+# scrapes Dockerfile for wget commands for fetching current versions of tanzu carvel tools,
+#  replaces "linux" with "darwin" so we pull the osx images, cleans up lingering Docker and multiline-grep formatting,
```
cppforlife, line 3: how is this meant to deal with checksums?

danielhelfand: I guess the script could have the checksums hardcoded and set as vars that could be `sed` into the script in place of where the linux checksums are.

joe-kimmel-vmw: Optimistically?  :)  right now it removes them in one of the `grep -v` calls,  since the Dockerfile checksums were for the linux build. 

I felt ok being optimistic about checksums for a development or semi-manual environment setup, and we do still run the `<tool> version` command as a sanity check so there would be prompt feedback if the installed binary were inoperable. 

 Is there a different location from which we could try to scrape the correct checksums (and would that add enough complication to this little script that it became harder to use/read/maintain)?

joe-kimmel-vmw:
will close in favor of https://github.com/vmware-tanzu/carvel-kapp-controller/pull/296

PR #291, opened 2021-07-12
bump ytt and imgpkg

ewrenn8:


Dockerfile
```diff
@@ -22,8 +22,8 @@ RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" && chmod -R 777 "$GOPATH"
 WORKDIR /go/src/github.com/vmware-tanzu/carvel-kapp-controller/
 
 # carvel
-RUN wget -O- https://github.com/vmware-tanzu/carvel-ytt/releases/download/v0.34.0/ytt-linux-amd64 > /usr/local/bin/ytt && \
-  echo "49741ac5540fc64da8566f3d1c9538f4f0fec22c62b8ba83e5e3d8efb91ee170  /usr/local/bin/ytt" | sha256sum -c - && \
+RUN wget -O- https://github.com/vmware-tanzu/carvel-ytt/releases/download/v0.35.0/ytt-linux-amd64 > /usr/local/bin/ytt && \
```
danielhelfand: We should potentially put a hold on this update as there may have been a minor issue with the release process for ytt 0.35.0. A follow up release may be put out that we should look to target.

ewrenn8: I will convert to draft until thats figured out

danielhelfand: 0.35.1 is out now. Think we should be good to go.

danielhelfand (commented):
lgtm

PR #292, opened 2021-07-13
api_port is required; new default api port

joe-kimmel-vmw:
with @ewrenn8 we did verify manually that this panics if the value is removed from deployment.yml.

### DIFF
--- a/config/values.yml
+++ b/config/values.yml
@@ -11,4 +11,4 @@ push_images: false
 image_cache: true
 image_repo: docker.io/k14stest/kapp-controller-test
 
-api_port: 10349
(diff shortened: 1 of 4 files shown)


config/values.yml
```diff
@@ -11,4 +11,4 @@ push_images: false
 image_cache: true
 image_repo: docker.io/k14stest/kapp-controller-test
 
-api_port: 10349
+api_port: 10350
```
danielhelfand, line 14: Due to how this is required by the start up code itself, should we add anything with ytt to make sure this field is specified? Maybe even think about using a simple schema here?

PR #295, opened 2021-07-15
test-e2e.sh: idempotently and safely create test ns if not exists

joe-kimmel-vmw:


hack/test-e2e.sh
```diff
@@ -4,6 +4,8 @@ set -e -x -u
 
 go clean -testcache
 
+# create ns if not exists because the `apply -f -` won't complain if fed an empty yaml to apply due to the ns already existing.
```
cppforlife: i dont think `kubectl create` will ever return empty yaml

joe-kimmel-vmw: that's fair , probably more of a "no-op" yaml. i'll adjust the comment.

PR #296, opened 2021-07-15
One installer script usable for both Docker and MacOS

joe-kimmel-vmw:
updated to resolve merge conflicts from https://github.com/vmware-tanzu/carvel-kapp-controller/pull/291   -- versions were already correct. 

manually verified both docker image and running on my mac seem to result in the right versions of things.

### DIFF
--- a/hack/install-deps.sh
+++ b/hack/install-deps.sh
@@ -18,7 +18,12 @@ install() {
     dl_bin="curl -s -L"
   fi
 
-  shasum -v 1>/dev/null 2>&1 || (echo "Missing shasum binary" && exit 1)
+  if which sha256sum; then
+	echo "found sha256sum"
+  else
+	echo "Missing sha256sum binary"
+	exit 1
+  fi
 
   ytt_version=v0.35.1
   kbld_version=v0.30.0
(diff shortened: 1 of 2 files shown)


hack/install-deps.sh
```diff
@@ -0,0 +1,90 @@
...
+
+if test -z "$BASH_VERSION"; then
+  echo "Please run this script using bash, not sh or any other shell." >&2
+  exit 1
+fi
+
+install() {
+  set -euo pipefail
+
+  dst_dir="${K14SIO_INSTALL_BIN_DIR:-/usr/local/bin}"
```
danielhelfand: Since this script is new, it might be nice to rebrand with something Carvel as opposed to k14s.

joe-kimmel-vmw: I like it, but my inclination is to maintain "backwards" compatibility since it's a 'new' script ripped off of an old script.  bash seems to have no issues nesting the little `:-` default-if-null operator.  e.g.  (in my environment the K14 env var isn't set,  but the EDITOR is) : 

```
➜  carvel-kapp-controller git:(develop) echo "${K14SIO_INSTALL_BIN_DIR:-${EDITOR:-/usr/local/bin}}"                                  
nvim
➜  carvel-kapp-controller git:(develop) echo "${K14SIO_INSTALL_BIN_DIR:-${CARVEL_ALSO_NOT_SET:-/usr/local/bin}}"                     
/usr/local/bin
```

PR #299, opened 2021-07-19
Updates to PackageMetadata to display more relevant information and f…

mlunadia:
…ields

Some fields have been updated with new information, other fields have been added eg. icon


cppforlife:
included suggested changes in https://github.com/vmware-tanzu/carvel-kapp-controller/pull/298

PR #302, opened 2021-07-20
Added metadata to Package CRs for Carvel test package

mlunadia:
Adding metadata changes to Package CRs (v1.0,2.0 &3.0)

### DIFF
--- a/test/e2e/assets/kc-e2e-test-repo/packages/pkg.test.carvel.dev/pkg.test.carvel.dev.2.0.0.yml
+++ b/test/e2e/assets/kc-e2e-test-repo/packages/pkg.test.carvel.dev/pkg.test.carvel.dev.2.0.0.yml
@@ -8,8 +8,9 @@ spec:
   # In this new version of the package, we are including some overlays
   # which will increase the pod count for the simple app.
   version: 2.0.0
+  releasedAt: 2021-05-05T18:57:06Z
   licenses:
-  - Apache 2.0
+  - VMware's End User License Agreement [hyperlink: https://www.vmware.com/content/dam/digitalmarketing/vmware/en/pdf/downloads/eula/universal_eula.pdf] (underlying OSS license: Apache License 2.0 [hyperlink: https://www.apache.org/licenses/LICENSE-2.0.txt])
   capacityRequirementsDescription: "cpu: 1,RAM: 2, Disk: 3"
   releaseNotes: |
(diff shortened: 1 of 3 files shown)


danielhelfand:
Thanks for the update @mlunadia. Think these look great.

Are there any additional properties to add moving forward that we should take some time to think about now? The reason I ask is we try to keep this example pushed and up to date and use in tests. So every time an update is made, we usually need to push the example and also update in tests. 

Just want to make sure that we have collected all needed information for the time being so we can not update it so frequently. It might also be worth it for us to look into automating the pushing of the example to make it so example updates aren't so challenging moving forward.

mlunadia:
Hey @danielhelfand We don't have plans on doing any changes on the key-value pairs for the metadata in the short term. These PR was about adding fields that were missing from the original example and all have now been completed. Thanks!

test/e2e/assets/kc-e2e-test-repo/packages/pkg.test.carvel.dev/pkg.test.carvel.dev.2.0.0.yml
```diff
@@ -8,8 +8,10 @@ spec:
   # In this new version of the package, we are including some overlays
   # which will increase the pod count for the simple app.
   version: 2.0.0
+  releasedAt: 2021-05-05T18:57:06Z
```
danielhelfand, line 11: Since this is for illustrative purposes, would it actually make sense to show this as released later than version 1.0.0?

test/e2e/assets/kc-e2e-test-repo/packages/pkg.test.carvel.dev/pkg.test.carvel.dev.3.0.0-rc.1.yml
```diff
@@ -9,8 +9,10 @@ spec:
   # PackageInstalls to unless it explicitly allowed within the PackageInstall
   # version selection.
   version: 3.0.0-rc.1
+  releasedAt: 2021-05-05T18:57:06Z
```
danielhelfand, line 12: Since this is for illustrative purposes, would it actually make sense to show this as released later than version 2.0.0?

cppforlife:
merged a different PRs for for spec.releasedAt. dont think changing license fields for this pkg repo makes sense.

PR #303, opened 2021-07-20
sops: improve error handling

zmb3:
- include sops stderr in the error when execution fails
- update sops to 3.7.1 which itself includes better error reporting

### DIFF
--- a/pkg/template/sops.go
+++ b/pkg/template/sops.go
@@ -143,7 +143,7 @@ func (t *Sops) decryptSopsFile(path, newPath string, args, env []string) error {
 
 	err := cmd.Run()
 	if err != nil {
-		return fmt.Errorf("Running sops: %s", err)
+		return fmt.Errorf("Running sops: %s, %v", stderrBs.String(), err)
 	}
 
 	err = os.Remove(path)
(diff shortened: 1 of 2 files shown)


pkg/template/sops.go
```diff
@@ -143,7 +143,7 @@ func (t *Sops) decryptSopsFile(path, newPath string, args, env []string) error {
 
 	err := cmd.Run()
 	if err != nil {
-		return fmt.Errorf("Running sops: %s", err)
+		return fmt.Errorf("Running sops: %s, %v", stderrBs.String(), err)
```
danielhelfand, line 146: Would we still want to include `err` here? 

Do you have an example of what this looks like when an error occurs?

danielhelfand (commented):
Thanks for opening this @zmb3. Quick comment below.

pkg/template/sops.go
zmb3: Yes. Before this change, the error looks like:

```
Running sops: exit code 1
```

After the change, it looks something like:

```
Running sops: Failed to get the data key required to decrypt the SOPS file.
Group 0: FAILED
  0979BA19CE2BAEF3: FAILED
    - | could not decrypt data key with PGP key:
      | golang.org/x/crypto/openpgp error: Reading PGP message
      | failed: EOF; GPG binary error: exec: "gpg": executable file
      | not found in $PATH
Recovery failed because no master key was able to decrypt the file. In
order for SOPS to recover the file, at least one key has to be successful,
but none were. exit status 128
```

danielhelfand: Ah my mistake. I didn't see the conversion to `%v`. Thanks and lgtm.

PR #305, opened 2021-07-20
WIP: Updates to kapp-controller tests for kapp waiting behavior for PackageRepos/PackageInstalls

danielhelfand:
**DO NOT MERGE**

This pull request is dependent on a new release of kapp that will include https://github.com/vmware-tanzu/carvel-kapp/pull/245. Tests will not pass until the new version is installed in CI via the test-gh workflow.

This pull request mainly serves to show what changes would occur after this new version of kapp is released. Of note:
* PackageRepository tests will no longer need to use the `retry()` func for checking if Packages are successfully available/deleted from a cluster
* PackageInstalls will no longer need to use `kubectl wait` to verify the PackageInstall was successful or failed

Behavior changes noticed:
* kapp seems to usually create PackageInstalls before Packages/PackageMetadata are created. This was always the behavior but now kapp fails more immediately since it is waiting on the success of the resource. A workaround for now is to include a kapp change rule to make the PackageInstall be created after the PackageRepository/Packages are created. This pattern of creating the PackageInstall with the Package/PackageRepository is probably something more common for testing than a typical user experience, but it may be something to work on supporting in kapp.

### DIFF
--- a/test/e2e/app_secret_configmap_reconcile_test.go
+++ b/test/e2e/app_secret_configmap_reconcile_test.go
@@ -208,3 +208,18 @@ data:
 		})
 	})
 }
+
+func retry(t *testing.T, timeout time.Duration, f func() error) {
+	var err error
+	stopTime := time.Now().Add(timeout)
+	for {
+		err = f()
+		if err == nil {
+			return
+		}
+		if time.Now().After(stopTime) {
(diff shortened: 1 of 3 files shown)


danielhelfand:
Closing in favor of #348

Issue #139, opened 2021-03-18
Determine What Version of kapp-controller is Available on a Cluster


danielhelfand:
**Describe the problem/challenge you have**

Unless I look through logs, there is no simple way to determine what version of kapp-controller is installed on a Kubernetes cluster. 

**Describe the solution you'd like**

In general, this can be further abstracted through client tools, but it might be easy enough for now to add a label to the kapp-controller deployment that has the version of kapp-controller. This would allow client tools to get the value of the label and show what the version is.

An example with `kubectl` would be:

`kubectl get deployment/kapp-controller  -o jsonpath={.metadata.labels['version']} -n kapp-controller`

cppforlife:
currently you can also identify git sha by looking at kbld's images annotation on kapp-controller Deployment and look in the git repo what that sha correlates to.

aaronshurley:
From our discussion...

There are multiple approaches here:
- build time (such as adding an annotation)
- run time (such as kapp-controller creating a ConfigMap in a particular namespace)

joe-kimmel-vmw:
there is now a version annotation if you describe the deployment.   I think this went out with 0.24 or 0.25. 
detailed instructions are here:
https://github.com/vmware-tanzu/carvel/commit/78fb8439e86098f1f1a2bd495c464c299140e6a4
(`kubectl describe deployment kapp-controller -n kapp-controller | grep version`)

PR #313, opened 2021-07-24
kapp controller version (#139)

joe-kimmel-vmw:
This branch presents five possible solutions to  #139 ,  with my preferred sol'n as the present diff. 

advantages of doing it as shown in this last commit:
- we are forwards-compatible with introducing goreleaser, the release tool we're starting to use elsewhere in Carvel suite
- the scripts under `hack` can each be run individually or chained together as one would expect
- version information is stored in exactly one canonical place
- no env var pollution, minimal state dependencies
- no change needed in the way we make tags for our releases.

examples / demo:
just running ytt render works:
```
./hack/ytt-me.sh | grep kapp_controller_vers                                                                                      
    kapp_controller_version: v0.21.0
```
querying kubectl for the deployed version works:
```
 kubectl describe deployment kapp-controller -n kapp-controller | grep kapp_controller_version                                       
                          {"apiVersion":"apps/v1","kind":"Deployment","metadata":{"annotations":{"kapp_controller_version":"v0.21.0","kbld.k14s.io/images":"- Metas:...
                        kapp_controller_version: v0.21.0
```
version info is still in the logs:
```
 ./controller                                                                                                                        
{"level":"info","ts":1627086436.27919,"logger":"kc.main","msg":"kapp-controller","version":"v0.21.0"}
{"level":"info","ts":1627086436.2792408,"logger":"kc.init","msg":"start init"}
{"level":"info","ts":1627086436.279288,"logger":"kc.init","msg":"starting zombie reaper"}
{"level":"info","ts":1627086436.343254,"logger":"kc.main","msg":"kapp-controller","version":"v0.21.0"}
{"level":"info","ts":1627086436.343317,"logger":"kc.controller","msg":"start controller"}
```

briefly, options considered include:
- [have two copies of the version](https://github.com/vmware-tanzu/carvel-kapp-controller/commit/81c9ed6bdbe384203f750be8352b8bbdfe3775a0) 
- scrape the version from our go code 
   - [with ytt](https://github.com/vmware-tanzu/carvel-kapp-controller/commit/2011446a0b85b255116ab80204abb8ee918ee277)
   - [with sed](https://github.com/vmware-tanzu/carvel-kapp-controller/commit/25d2de3cdaf9b721170f9b090c11ce253ad1ff3d)
- pull the version from git tags, like goreleaser does
  - [with a global/env var](https://github.com/vmware-tanzu/carvel-kapp-controller/commit/1b29700475c13e6f707266cecff08519bbde79a7)
   - [functional style](https://github.com/vmware-tanzu/carvel-kapp-controller/commit/03777092ce5fb8e850b71f308c43a1eb95bc3b8c)
   

extra thanks to @DennisDenuto for explaining how imgpkg is using goreleaser!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 FROM photon:3.0
 
 ARG KCTRL_VER=development
-LABEL kapp_controller_version=$KCTRL_VER
 
 RUN tdnf install -y tar wget gzip
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,8 @@
 FROM photon:3.0
 
+ARG KCTRL_VER=development
+LABEL kapp_controller_version=$KCTRL_VER
+
 RUN tdnf install -y tar wget gzip
 
 # adapted from golang docker image
@@ -42,7 +45,8 @@ RUN wget -O- https://github.com/mozilla/sops/releases/download/v3.7.1/sops-v3.7.
 
 # kapp-controller
 COPY . .
-RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags=-buildid= -trimpath -o controller ./cmd/main.go
+# helpful ldflags reference: https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications
+RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags="-X 'main.Version=$KCTRL_VER' -buildid=" -trimpath -o controller ./cmd/main.go
 
 # --- run image ---
 FROM photon:3.0
--- a/config/build.yml
+++ b/config/build.yml
@@ -5,7 +5,6 @@ kind: Sources
 sources:
 - image: kapp-controller
   path: .
-  #@ def docker_rawOpts(): return ["--build-arg", "KCTRL_VER="+data.values.kapp_controller_version]
   #@ if/end not data.values.image_cache:
   docker:
     build:
@@ -13,7 +12,7 @@ sources:
       pull: true
       noCache: true
       #! pass kapp_controller_version into Dockerfile.
-      rawOptions: #@ docker_rawOpts()
+      rawOptions: #@ ["--build-arg", "KCTRL_VER="+data.values.kapp_controller_version]
 
 #@ if/end data.values.push_images:
 ---
--- a/config/build.yml
+++ b/config/build.yml
@@ -5,12 +5,15 @@ kind: Sources
 sources:
 - image: kapp-controller
   path: .
+  #@ def docker_rawOpts(): return ["--build-arg", "KCTRL_VER="+data.values.kapp_controller_version]
   #@ if/end not data.values.image_cache:
   docker:
     build:
       #! Always rebuild image
       pull: true
       noCache: true
+      #! pass kapp_controller_version into Dockerfile.
+      rawOptions: #@ docker_rawOpts()
 
 #@ if/end data.values.push_images:
 ---
--- a/config/deployment.yml
+++ b/config/deployment.yml
@@ -6,7 +6,7 @@ metadata:
   name: kapp-controller
   namespace: #@ data.values.namespace
   annotations:
-    kapp_controller_version: #@ data.values.kapp_controller_version
+    kapp_controller.carvel.dev/version: #@ data.values.kapp_controller_version
 spec:
   selector:
     matchLabels:
--- a/config/deployment.yml
+++ b/config/deployment.yml
@@ -6,7 +6,7 @@ metadata:
   name: kapp-controller
   namespace: #@ data.values.namespace
   annotations:
-    kapp_controller.carvel.dev/version: #@ data.values.kapp_controller_version
+    kapp-controller.carvel.dev/version: #@ data.values.kapp_controller_version
 spec:
   selector:
     matchLabels:
--- a/config/deployment.yml
+++ b/config/deployment.yml
@@ -5,6 +5,8 @@ kind: Deployment
 metadata:
   name: kapp-controller
   namespace: #@ data.values.namespace
+  annotations:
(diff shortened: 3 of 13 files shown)


config/deployment.yml
```diff
@@ -5,6 +5,8 @@ kind: Deployment
 metadata:
   name: kapp-controller
   namespace: #@ data.values.namespace
+  annotations:
+    kapp_controller_version: #@ data.values.kapp_controller_version
```
cppforlife: i think we should namespace this annotation e.g. `kapp-controller.carvel.dev/version`

config/values.yml
```diff
@@ -1,4 +1,6 @@
 #@data/values
+
+#@ load("@ytt:data", "data")
```
cppforlife: no need for load data here.

hack/build.sh
```diff
@@ -2,18 +2,24 @@
 
 set -e -x -u
 
-# makes builds reproducible
```
cppforlife, line 5: i dont think we need to change ./hack/build.sh as its used for development, and development version should probably stay as "develop" (similar to https://github.com/vmware-tanzu/carvel-imgpkg/blob/develop/hack/build.sh)

hack/deploy.sh
```diff
@@ -2,4 +2,4 @@
 
 set -e
 
-./hack/build.sh && ytt -f config/ | kbld -f- | kapp deploy -a kc -f- -c -y
+./hack/build.sh && ./hack/ytt-me.sh | kbld -f- | kapp deploy -a kc -f- -c -y
```
cppforlife: same goes for deploy.sh as build.sh. we should probably change ./hack/build-release.sh to include a version. @danielhelfand build-release.sh is what gets used for building final artifacts (oss and internal), right?

hack/ytt-me.sh
```diff
@@ -0,0 +1,5 @@
+# ytt-me is intended to be replace ytt -f args... in a chain of piped commands and so does not provide its own shebang
+
+source $(dirname "$0")/version-me.sh
+
+ytt -f config/ -v kapp_controller_version="$(get_kappctrl_ver)"
```
cppforlife: i woudl recommend just adding -v ... to build-release.sh directly instead of adding indirection to this file.

config/build.yml
```diff
@@ -5,12 +5,15 @@ kind: Sources
 sources:
 - image: kapp-controller
   path: .
+  #@ def docker_rawOpts(): return ["--build-arg", "KCTRL_VER="+data.values.kapp_controller_version]
```
cppforlife: any reason to not inline this?

config/values.yml
```diff
@@ -1,4 +1,5 @@
 #@data/values
+
```
cppforlife: nit: remove newline

Dockerfile
```diff
@@ -1,5 +1,8 @@
 FROM photon:3.0
 
+ARG KCTRL_VER=development
+LABEL kapp_controller_version=$KCTRL_VER
```
cppforlife: any reason to label this image? (if we really want to, then lets namespace label `kapp-controller.carvel.dev/version=...`)

hack/deploy.sh
danielhelfand: `./hack/build-release.sh` is used for oss but not internal. The internal build is just using a go build, so the ldflag would need to be added along with the version information.

config/build.yml
joe-kimmel-vmw: nah

Dockerfile
joe-kimmel-vmw: thought it would be thorough somehow, but say no more! it's gone

hack/deploy.sh
joe-kimmel-vmw: huh. -  thanks.  i think that needs a little more thought then.  I appreciated the idea of having the default for dev builds actually keep the develop flag but it sounds like we need to either slightly adjust our internal build process or accept that everything will get versioned.

danielhelfand: We could adjust the internal build to use the hack script. We basically adopted our internal process from another team and I am not sure why some choices were made. I do appreciate the idea of providing the version build at run time when releasing. It makes it very clear when developing or if someone is building from source that an official release is not being used.

hack/build-release.sh
```diff
@@ -4,7 +4,10 @@ set -e -x -u
 
 mkdir -p tmp/
 
-ytt -f config/ -f config-release | kbld -f- > ./tmp/release.yml
+# makes the get_kappctrl_ver function available (scrapes version from git tag)
+source $(dirname "$0")/version-me.sh
```
cppforlife: nit: version-me.sh -> version-utils.sh?

hack/version-me.sh
```diff
@@ -0,0 +1,16 @@
+# source this file:
+# use this via `source ./hack/version-me.sh` because `./hack/version-me.sh` won't make the functions available to other scripts
+
+# We extract version information from git tags
+# the implicit contract is that our git tags will be in ~semver (three-part) format and prefaced with the letter 'v'.
+# this contract is required by the goreleaser tool and used throughout Carvel suite.
+
+# git tag version extraction graciously provided by https://github.com/vmware-tanzu/carvel-imgpkg/blob/develop/hack/build-binaries.sh
+function get_latest_git_tag {
```
cppforlife: i was thinking right before merging, what does it mean to get a latest tag? should we be getting a tag on the current commit?

joe-kimmel-vmw: I agree that "latest tag" is fragile,  but the goal was to keep a consistent interface with GoReleaser:

> GoReleaser will use the latest Git tag of your repository.

https://goreleaser.com/quick-start/

hack/build-release.sh
joe-kimmel-vmw: :)  imperative "version me" vs. declarative "these are the version utils" ,  eh?   np.

Issue #310, opened 2021-07-22
Support kapp-controller Proxy Configuration via a Secret


soumyapanigrahi-code:
**Describe the problem/challenge you have**
Kapp controller exposes the ability to configure the controller via a config map which kapp controller will look for and apply as part of the startup processes. However some of the proxy information including passwords are sensitive in nature.

**Describe the solution you'd like**
Kapp controller exposes the ability to configure the controller via a secret.

**Anything else you would like to add:**
Given that kapp controller is gaining popularity and expected to be widely used in air-gapped environments as well, it would be great if this feature request is prioritised.

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

aaronshurley:
@soumyapanigrahi-code Thank you for surfacing this. We understand the use case and are discussing prioritization right now. We'll update this issue soon once we have a better idea of when we'd be able to start working on this.

danielhelfand:
Following up here with some initial design thoughts on how to approach this:
* We will continue to support ConfigMap configuration, but we will recommend the use of Secrets moving forward. We will also reinforce this idea by having kapp-controller prefer using the Secret in a scenario where both the kapp-controller-config ConfigMap/Secret exist. Additionally, [documentation](https://carvel.dev/kapp-controller/docs/latest/controller-config/) should be changed to push the use of Secrets.
* As a means of testing this feature, one thing we should consider is converting our [kapp-controller-config ConfigMap](https://github.com/vmware-tanzu/carvel-kapp-controller/blob/8873d1b0e91aac089536f6050a5835d4b301734b/config-test/config-map.yml) to a Secret.

aaronshurley:
Due to the proposed security improvement of this request, we have moved this to the top of the backlog and will begin working on this today.

danielhelfand:
This feature is available in release v0.22.0: https://github.com/vmware-tanzu/carvel-kapp-controller/releases/tag/v0.22.0

I will keep this open as a reminder to add documentation changes and close once those updates are made.

soumyapanigrahi-code:
Thanks a lot everyone. The quick turn-around time is much appreciated.

PR #318, opened 2021-07-28
Change Preferred Controller Config to Use Secret Instead of ConfigMap

danielhelfand:
Related issue is #310 

This pull request adds the ability for users to configure controller runtime values as a Secret instead of a ConfigMap. The current approach to support backwards compatibility will be to look for a secret named kapp-controller-config first and if not found will search for a ConfigMap with the same name.

This pull request also converts kapp-controller's ConfigMap used in testing with `./hack/deploy-test.sh` to a Secret to encourage the practice. 

A follow up to this will be documentation to promote the use of secrets as opposed to ConfigMaps for controller configuration.


pkg/config/config.go
```diff
@@ -129,3 +165,38 @@ func (gc *Config) configureProxies(httpProxy, httpsProxy, noProxy string) {
...
+
+	addDataToConfig(config, extractedValues)
+}
+
+func addConfigMapDataToConfig(config *Config, configMap *v1.ConfigMap) {
+	addDataToConfig(config, configMap.Data)
+}
+
+func addDataToConfig(config *Config, data map[string]string) {
+	if val, exists := data[caCertsKey]; exists {
```
cppforlife: no need to check for existence. its all empty string by default so can just assign

pkg/config/config.go
```diff
@@ -129,3 +165,38 @@ func (gc *Config) configureProxies(httpProxy, httpsProxy, noProxy string) {
...
+	}
+
+	addDataToConfig(config, extractedValues)
+}
+
+func addConfigMapDataToConfig(config *Config, configMap *v1.ConfigMap) {
+	addDataToConfig(config, configMap.Data)
+}
+
+func addDataToConfig(config *Config, data map[string]string) {
```
cppforlife: this could be a private method on the Config struct. then no need to pass config as first arg.

same for related methods.

pkg/config/config.go
```diff
@@ -129,3 +165,38 @@ func (gc *Config) configureProxies(httpProxy, httpsProxy, noProxy string) {
 		os.Setenv(strings.ToUpper(noProxyEnvVar), noProxy)
 	}
 }
+
+func addSecretDataToConfig(config *Config, secret *v1.Secret) {
+	extractedValues := map[string]string{}
+	keys := []string{caCertsKey, httpProxyKey, httpsProxyKey, noProxyKey, skipTLSVerifyKey}
```
cppforlife: might as well convert all keys. no need to filter out by specific key.

pkg/config/config.go
danielhelfand: Added

pkg/config/config.go
danielhelfand: Added

pkg/config/config.go
danielhelfand: @joe-kimmel-vmw Any opinions on this? I can see pros and cons for either approach.

Mainly, my thought is that we should not have it process unknown fields at all.

However, I see the benefit in maintaining this long term as far as not needing to maintain this list of keys in this func. In `addDataToConfig`, we are already ignoring unknown fields.

danielhelfand: After discussing further, we will go with approach of just iterating over all key-value pairs. 

This has been addressed in 68f0cf1.

Issue #283, opened 2021-07-07
Review for Helm Chart


linkdd:
# Describe the problem/challenge you have

At our company, [Link Society](https://github.com/link-society), we are currently working on a supervision engine on Kubernetes: [Kubirds](https://kubirds.com).

Its dependencies provide no Helm charts, therefore upgrading and uninstalling becomes an issue.

Fortunately `kapp` and `kapp-controller` resolves this versioning problem.

Since we're looking at integrating our product on the [DigitalOcean Marketplace](https://github.com/digitalocean/marketplace-kubernetes/), we were looking to integrate our dependencies as well.

This means also integrating **kapp-controller** in the marketplace (Work In Progress on our side).

But only `helm` and `kubectl` are officially supported in their workflow (no `kapp`)

# Describe the solution you'd like

To integrate **kapp-controller** to their marketplace, we started writing and maintaining an Helm Chart based on the latest release (`0.20.0`), the PR to our own Helm repository can be found here: [link-society/charts#29](https://github.com/link-society/charts/pull/29).

We would appreciate if people could review this chart and maybe give us insight on how it could be integrated upstream.

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

danielhelfand:
Hey @linkdd! Thank you for opening this.

Think it would be great for some of us in the Carvel community to take a look at Link Society's use of kapp-controller. We can follow up in the pr you have linked.

I had some follow up questions as well based on some of the information provided. Looking forward to your responses.

> Its dependencies provide no Helm charts, therefore upgrading and uninstalling becomes an issue.

> Fortunately kapp and kapp-controller resolves this versioning problem.

To understand a little more about your use case, are you saying Kubirds has dependencies and you are using kapp-controller's [package management API concepts](https://carvel.dev/kapp-controller/docs/latest/packaging/) to address this? Something else? Just trying to gain more understanding of how you are all using kapp-controller.

> Since we're looking at integrating our product on the DigitalOcean Marketplace, we were looking to integrate our dependencies as well.

> This means also integrating kapp-controller in the marketplace (Work In Progress on our side).

> But only helm and kubectl are officially supported in their workflow (no kapp)

[...]

PR #321, opened 2021-07-31
Add environment variable to parametrize service name

linkdd:
The `Service` resource name was statically defined as `packaging-api`.

This change introduces the environment variable `KAPPCTRL_SYSTEM_SERVICE` allowing the user to override those values.

If the environment variable isn't defined, they will default to the previously statically defined value.

See [this comment](https://github.com/vmware-tanzu/carvel-kapp-controller/issues/283#issuecomment-889989228) in the issue #283 for more informations.

### DIFF
--- a/pkg/apiserver/apiserver.go
+++ b/pkg/apiserver/apiserver.go
@@ -46,7 +46,7 @@ const (
 	kappctrlSVCEnvKey = "KAPPCTRL_SYSTEM_SERVICE"
 	kappctrlAPIEnvKey = "KAPPCTRL_SYSTEM_API"
 
-	defaultApiServiceName = "v1alpha1.data.packaging.carvel.dev"
+	apiServiceName = "v1alpha1.data.packaging.carvel.dev"
 )
 
 var (
@@ -134,7 +134,6 @@ func (as *APIServer) Stop() {
 }
 
 func (as *APIServer) isReady() (bool, error) {
-	apiServiceName := getEnvVal(kappctrlAPIEnvKey, defaultApiServiceName)
 	apiService, err := as.aggClient.ApiregistrationV1().APIServices().Get(context.TODO(), apiServiceName, metav1.GetOptions{})
 	if err != nil {
 		return false, fmt.Errorf("error getting APIService %s: %v", apiServiceName, err)
@@ -190,7 +189,6 @@ func newServerConfig(aggClient aggregatorclient.Interface, bindPort int) (*gener
 
 func updateAPIService(client aggregatorclient.Interface, caProvider dynamiccertificates.CAContentProvider) error {
 	klog.Info("Syncing CA certificate with APIServices")
-	apiServiceName := getEnvVal(kappctrlAPIEnvKey, defaultApiServiceName)
 	apiService, err := client.ApiregistrationV1().APIServices().Get(context.TODO(), apiServiceName, metav1.GetOptions{})
 	if err != nil {
 		return fmt.Errorf("error getting APIService %s: %v", apiServiceName, err)
--- a/pkg/apiserver/apiserver.go
+++ b/pkg/apiserver/apiserver.go
@@ -42,8 +42,11 @@ const (
 
 	TokenPath = "/token-dir"
 
-	kappctrlNSEnvKey = "KAPPCTRL_SYSTEM_NAMESPACE"
-	apiServiceName   = "v1alpha1.data.packaging.carvel.dev"
+	kappctrlNSEnvKey  = "KAPPCTRL_SYSTEM_NAMESPACE"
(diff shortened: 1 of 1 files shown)


pkg/apiserver/apiserver.go
```diff
@@ -131,6 +134,7 @@ func (as *APIServer) Stop() {
 }
 
 func (as *APIServer) isReady() (bool, error) {
+	apiServiceName := getEnvVal(kappctrlAPIEnvKey, defaultApiServiceName)
```
cppforlife: @linkdd whast the thinking behind `v1alpha1.data.packaging.carvel.dev` override? this is a name that is used by packages.

linkdd: I'm currently working on an Helm Chart (see #283). This would allow me to have multiple release of the same chart in the same cluster (not that's something you would want, but who knows with multi-tenancy?).

By overriding this value, one could provide a `v1alpha1` and a `v1alpha2` without conflicts in the chart. Useful for A/B testing?

The `APIService` is required by the operator to work. So if only one chart in the cluster can deploy it (because of the static name), this introduces a tight coupling between the operator and how it is deployed.

pkg/apiserver/apiserver.go
```diff
@@ -198,11 +203,19 @@ func updateAPIService(client aggregatorclient.Interface, caProvider dynamiccerti
 }
 
 func apiServiceEndoint() string {
-	const apiServiceName = "packaging-api"
+	const apiServiceName = getEnvVal(kappctrlSVCEnvKey, "packaging-api")
```
b3j0f:
```suggestion
	var apiServiceName = getEnvVal(kappctrlSVCEnvKey, "packaging-api")
```

b3j0f (changes requested):

pkg/apiserver/apiserver.go
cppforlife: > By overriding this value, one could provide a v1alpha1 and a v1alpha2 without conflicts in the chart. Useful for A/B testing?

in future versions of kapp-controller i would expect kc to support multiple versions at the same time for backwards compatible upgrades (standard behaviour for k8s controllers; resources within the cluster and in different packages would be using different versions in parallel). so i dont think this really works by making this configurable.

linkdd: Ok, I removed it and will add a condition in the Helm Chart to skip the creation of the APIService.

This means that if there are multiple release of the same chart in the cluster, the `Package` resources will be served by only one of them (is this an issue?)

cppforlife: > This means that if there are multiple release of the same chart in the cluster, the Package resources will be served by only one of them (is this an issue?)

i dont think we really can support multiple kapp-controllers on one cluster without dynamic renaming of CRDs and APIServices (and that's very uncommon in k8s ecosystem). renaming leads to a ton of complexity/confusion since rename has to propagate not thru all the code but also into consumer configuration (e.g. wouldnt be able to use regular App CR api group, or package repositories wouldnt be able to use packaging apiGroups). if you want to reliably do blue-green testing for kapp-controller, you ll have to do it against multiple clusters (eg staging vs prod).

linkdd: Fair enough, I'll just have to document my chart and make it clear that it cannot be deployed multiple times.

So now, this PR only introduce the dynamic service name. Hope it can be merged 🙂

cppforlife: @linkdd to push further on this... sorry to be bombarding you with questions... how are you planning to take advantage of changeable service name? any specific use case for allowing it to be renamed?

linkdd: @cppforlife At the moment, in my [Helm Chart](https://github.com/link-society/charts/blob/incubator/kapp-controller/incubator/kapp-controller/templates/service.yml), the `Service` is named `{{ include "kapp-controller.fullname" . }}-packaging-api`.
 
 Since the controller can be deployed in any namespace this avoid name clashes (what if I already have a service named `packaging-api` from a completely different project?).

The ytt templates used to generate the release manifest defines the namespace as `kapp-controller` by default, but it can be overridden. For Helm, the default namespace is the one specified in your `.kube/config`, so one should be careful about that 🙂

linkdd:
```
./hack/test-e2e.sh: line 11: KAPPCTRL_E2E_SECRETGEN_CONTROLLER: unbound variable
```

I don't really get this error, it doesn't seem it's due to my changes, is it?

cppforlife:
hmm, may be try rebasing?

linkdd:
Hi @cppforlife 
> hmm, may be try rebasing?

Sorry for the late reply, I've just pushed the rebased branch 😄

PR #327, opened 2021-08-10
Support Testing Integration with secretgen-controller

### LINKED ISSUE #282: Tests for Integration with secretgen-controller
As kapp-controller implements the proposal for authentication to private registries, there will need to be a way to write tests for how kapp-controller works with [secretgen-controller](https://github.com/vmware-tanzu/carvel-secretgen-controller). 

secretgen-controller will be the primary way of supporting authentication to private registries via kapp-controller so it will be important to have a way to support adding secretgen-controller to a cluster to run tests around private registry authentication.

Related issues that this feature will support: 
* #242
* #243

The desired behavior is for adding a testing solution around secretgen-controller is that by default kapp-controller does not run tests where secretgen-controller is assumed to be installed. This is mainly because kapp-controller should still be able to run even without secretgen-controller installed. However, in CI, these new tests should always run.

Some considerations for this issue are as follows:
* Should we create a dedicated Package CR for secretgen-controller so that it can easily be installed for testing?
* How should tests for kapp-controller with secretgen-controller be enabled (e.g. an environment variable?)
* What changes should happen in CI (i.e. [GitHub action](https://github.com/vmware-tanzu/carvel-kapp-controller/blob/7c1bfe00fd9aa229c0356d2ac030709813e759df/.github/workflows/test-gh.yml)/hack scripts) to support this?
* Provide documentation on how to enable the running of these tests

danielhelfand:
Closes #282 

This pull request adds the ability to deploy secretgen-controller and run associated tests for kapp-controller's integration with secretgen-controller based on an environment variable being set named `KAPPCTRL_E2E_SECRETGEN-CONTROLLER`. 

If `KAPPCTRL_E2E_SECRETGEN-CONTROLLER=true`, both deploy.sh and deploy-test.sh will deploy a specific version of secretgen-controller. Additionally, if `KAPPCTRL_E2E_SECRETGEN-CONTROLLER=true`, test-all.sh will run tests under the secretgencontroller folder.

### DIFF
--- a/.github/workflows/test-gh.yml
+++ b/.github/workflows/test-gh.yml
@@ -51,6 +51,7 @@ jobs:
         minikube start --driver=docker
         eval $(minikube docker-env --shell=bash)
 
+        export KAPPCTRL_E2E_SECRETGEN_CONTROLLER=true
         ./hack/deploy-test.sh
 
         export KAPPCTRL_E2E_NAMESPACE=kappctrl-test
--- a/docs/dev.md
+++ b/docs/dev.md
@@ -5,8 +5,16 @@ Install ytt, kbld, kapp beforehand (https://k14s.io).
 ```
 ./hack/build.sh # to build locally
 
+# deploys secretgen-controller with kapp-controller
+# and also runs tests where kapp-controller integrates 
+# with secretgen-controller
+export KAPPCTRL_E2E_SECRETGEN_CONTROLLER=true
+
 # add `-v image_repo=docker.io/username/kapp-controller` with your registry to ytt invocation inside
 ./hack/deploy.sh # to deploy
+
+# deploys test assets in addition to kapp-controller for e2e tests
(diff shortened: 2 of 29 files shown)


danielhelfand:
Just a note to anyone reviewing this that GitHub actions are experiencing issues at the moment: https://www.githubstatus.com/. We should let CI run before introducing this if reviews look good.

.github/workflows/test-gh.yml
```diff
@@ -51,6 +51,7 @@ jobs:
         minikube start --driver=docker
         eval $(minikube docker-env --shell=bash)
 
+        export SECRETGEN-CONTROLLER=true
```
cppforlife: i think we should prefix this env var with KAPPCTRL_E2E_. lets also use _ instead of -

danielhelfand: Thanks, added

PR #331, opened 2021-08-11
Private Registry Auth for PackageInstalls

### LINKED ISSUE #243: Feature:  PackageInstall's App can automatically auth to private registries when secretgen-controller is in the cluster
### Description of Problem
There is currently no way to provide private registry auth details during the installation of a package. This means any packages with fetch directives which fetch from a private registry will fail.

### Description of Solution
Kapp-controller is able to automatically handle the authorization setup for PackageInstalls.

#### 🟢 Public Registry

**Given** a package with a fetch stage that pulls from a public registry
**When** a user creates a PackageInstall to install that package
**And** the package will eventually be installed successfully

**Note:** Because kapp-controller can't determine if a registry url is public or private, it may need to create the placeholder secret anyway

---
#### 🟢 Simple Auth Scenario (1 registry, 1 fetch, 1 user)

**Given** a package with a fetch stage that pulls from a private registry
**When** a user creates a PackageInstall to install that package
**Then** kapp-controller will create a placeholder secret for the fetch stage
**And** the package will eventually be installed successfully

---
#### 🟢 Multi registry (2+ registries, 2+ fetch stages, 1 user)

**Given** a package with multiple fetch stages that pull from multiple private registries
**When** a user creates a PackageInstall to install that package
**Then** kapp-controller will create a **single** placeholder secret (gets added to auth details of each fetch stage, though)
**And** the package will eventually be installed successfully

**Note:** Placeholder secrets will get all exported dockerconfigs, so fetch stages can reuse. Can we reuse across PackageInstalls as well?

---
#### 🟢 Private Repo with explicit auth

**Given** A package with fetch stages
**And** The fetch stages have explicit auth details
**When** A user adds a PackageInstall CR to install the package
**Then** kapp-controller **does not** create a placeholder secret for the fetch stages
**And** Eventually kapp-controller successfully fetches the repository bundle

---
#### 🔴 Credentials are not exported

**Given** a package with a fetch stage pulling from a private registry
**And** No secrets are exported (secretgen-controller behavior)
**When** a user creates a PackageInstall to install that package
**Then** kapp-controller will create a placeholder secret
**And** Will add an error message detailing the auth failure from the registry when fetching

---
#### 🔴 Credentials are exported but not valid

**Given** a package with a fetch stage pulling from a private registry
**And** The exported credentials don't work (secretgen-controller behavior)
**When** a user creates a PackageInstall to install that package
**Then** kapp-controller will create a placeholder secret
**And** Will add an error message detailing the auth failure from the registry when fetching

---
#### 🔴  Multiple users for same registry (Error)

**Given** a package with two or more fetch stages pulling from the same private registry but different repos with different credentials
**And** Both credentials are exported for the same registry url
**When** a user creates a PackageInstall to install that package
**Then** kapp-controller will fail to fetch the bundle, since only one credential can be used for the url

### More Info

Similar to #242 , we will leverage secretgen-controller (SGC) to automate this process. The kapp-controller will be responsible for creating placeholder secrets for the fetch stages of a package, which will get populated by SGC if there is an exported secret matching the criteria. There are a few considerations here:

1. We should consider the race condition between secretgen-controller populating the place holder secret and kapp-controller reconciling. Given valid credentials, kapp-controller may reconcile before they're available causing failures that will eventually succeed. How does this affect the UX of users using kapp to deploy PackageInstalls and what can we do to improve that?

---
Vote on this request

danielhelfand:
Closes #243


cmd/controller/packageinstall_reconciler.go
```diff
@@ -12,13 +12,20 @@ import (
...
-	pkgClient pkgclient.Interface
-	log       logr.Logger
+	kcClient   kcclient.Interface
+	pkgClient  pkgclient.Interface
+	coreClient kubernetes.Interface
+	log        logr.Logger
+}
+
+// NewPackageInstallReconciler is the constructor for the PackageInstallReconciler struct
+func NewPackageInstallReconciler(kcClient kcclient.Interface, pkgClient pkgclient.Interface, coreClient kubernetes.Interface, log logr.Logger) *PackageInstallReconciler {
```
cppforlife: nit: wrap line

pkg/app/app_fetch.go
```diff
@@ -43,6 +44,31 @@ func (a *App) fetch(dstPath string) (string, exec.CmdRunResult) {
 	}
 
 	result = a.runVendir(confReader, dstPath)
+	// retry if error occurs before reporting failure.
+	// This is mainly done to support private registry
+	// authentication for images/bundles since placeholder
+	// secrets may not be populated in time.
+	if result.Error != nil && a.HasImageOrImgpkgBundle() {
```
cppforlife, line 51: idea: what if we record whether we created any placeholder secrets earlier and only do retry if we did.

pkg/app/app_fetch.go
```diff
@@ -43,6 +44,31 @@ func (a *App) fetch(dstPath string) (string, exec.CmdRunResult) {
...
+		// Only retrying once resulted in flaky behavior
+		// for private auth so use 5 iterations.
+		for i := 0; i < 5; i++ {
```
cppforlife: TODO: figure out why sg populating is that slow that we need 5s

pkg/packageinstall/packageinstall.go
```diff
@@ -27,16 +30,16 @@ type PackageInstallCR struct {
...
+	coreClient kubernetes.Interface
 }
 
 func NewPackageInstallCR(model *pkgingv1alpha1.PackageInstall, log logr.Logger,
-	kcclient kcclient.Interface, pkgclient pkgclient.Interface) *PackageInstallCR {
+	kcclient kcclient.Interface, pkgclient pkgclient.Interface, coreClient kubernetes.Interface) *PackageInstallCR {
 
-	return &PackageInstallCR{model: model, unmodifiedModel: model.DeepCopy(),
-		log: log, kcclient: kcclient, pkgclient: pkgclient}
+	return &PackageInstallCR{model: model, unmodifiedModel: model.DeepCopy(), log: log, kcclient: kcclient, pkgclient: pkgclient, coreClient: coreClient}
```
cppforlife: nit: wrap line

pkg/packageinstall/packageinstall.go
```diff
@@ -89,6 +92,10 @@ func (pi *PackageInstallCR) reconcile(modelStatus *reconciler.Status) (reconcile
 	existingApp, err := pi.kcclient.KappctrlV1alpha1().Apps(pi.model.Namespace).Get(context.Background(), pi.model.Name, metav1.GetOptions{})
 	if err != nil {
 		if errors.IsNotFound(err) {
+			err := pi.createPlaceHolderSecrets(&pv)
```
cppforlife: createPlaceHolderSecrets -> reconcileFetchPlaceholderSecrets?

pkg/packageinstall/packageinstall.go
```diff
@@ -266,3 +278,62 @@ func (pi *PackageInstallCR) update(updateFunc func(*pkgingv1alpha1.PackageInstal
 
 	return fmt.Errorf("Updating package install: %s", lastErr)
 }
+
+func (pi *PackageInstallCR) createPlaceHolderSecrets(pv *datapkgingv1alpha1.Package) error {
+	for i, fetch := range pv.Spec.Template.Spec.Fetch {
```
cppforlife: instead of modifying Package, lets make a DeepCopy() and return a modified copy.

pkg/packageinstall/packageinstall.go
```diff
@@ -266,3 +278,62 @@ func (pi *PackageInstallCR) update(updateFunc func(*pkgingv1alpha1.PackageInstal
...
+				return err
+			}
+			pv.Spec.Template.Spec.Fetch[i].Image.SecretRef = &kcv1alpha1.AppFetchLocalRef{secretName}
+		}
+	}
+	return nil
+}
+
+func (pi PackageInstallCR) createSecretForSecretgenController(iteration int) (string, error) {
+	secretName := fmt.Sprintf("%s-%s", pi.model.Name, "fetch"+strconv.Itoa(iteration))
```
cppforlife:
```
secretName := fmt.Sprintf("%s-%s", pi.model.Name, "fetch"+strconv.Itoa(iteration))
secretName := fmt.Sprintf("%s-fetch-%d", pi.model.Name, iteration)
```

pkg/packageinstall/packageinstall.go
```diff
@@ -266,3 +278,62 @@ func (pi *PackageInstallCR) update(updateFunc func(*pkgingv1alpha1.PackageInstal
...
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretName,
+			Namespace: pi.model.Namespace,
+			Annotations: map[string]string{
+				"secretgen.carvel.dev/image-pull-secret": "",
+			},
+			OwnerReferences: []metav1.OwnerReference{
```
cppforlife: controllerutil has SetOwnerReference i believe. can we use that here?

pkg/packageinstall/packageinstall.go
```diff
@@ -266,3 +278,62 @@ func (pi *PackageInstallCR) update(updateFunc func(*pkgingv1alpha1.PackageInstal
...
+		Data: map[string][]byte{
+			corev1.DockerConfigJsonKey: []byte(`{"auths":{}}`),
+		},
+		Type: corev1.SecretTypeDockerConfigJson,
+	}
+
+	_, err := pi.coreClient.CoreV1().Secrets(pi.model.Namespace).Create(context.Background(), secret, metav1.CreateOptions{})
+	if err != nil {
+		if errors.IsAlreadyExists(err) {
+			_, err = pi.coreClient.CoreV1().Secrets(pi.model.Namespace).Update(context.Background(), secret, metav1.UpdateOptions{})
```
cppforlife: should we bother with updates right now? one challenge is that secret updates are sensitive to type changes -- not allowed, so it would have to be delete/create cycle.

test/e2e/assets/registry/htpasswd-auth
```diff
@@ -0,0 +1,11 @@
+---
```
cppforlife, line 1: nit: throw in .yml extension

test/e2e/secretgencontroller/private_registry_auth_test.go
```diff
@@ -0,0 +1,260 @@
...
+func Test_PlaceholderSecrets_DeletedWhenPackageInstallDeleted(t *testing.T) {
+	env := e2e.BuildEnv(t)
+	logger := e2e.Logger{}
+	kapp := e2e.Kapp{t, env.Namespace, logger}
+	kubectl := e2e.Kubectl{t, env.Namespace, logger}
+	name := "placeholder-garbage-collection"
+	sas := e2e.ServiceAccounts{env.Namespace}
+
+	pkgiYaml := fmt.Sprintf(`---
+apiVersion: data.packaging.carvel.dev/v1alpha1
```
cppforlife, line 26: no need for PackageMetadata

test/e2e/secretgencontroller/private_registry_auth_test.go
```diff
@@ -0,0 +1,260 @@
...
+  namespace: %[1]s
+spec:
+  refName: pkg.test.carvel.dev
+  version: 1.0.0
+  licenses:
+  - Apache 2.0
+  capactiyRequirementsDescription: "cpu: 1,RAM: 2, Disk: 3"
+  releaseNotes: |
+    - Introduce simple-app package
+  releasedAt: 2021-05-05T18:57:06Z
```
cppforlife: no need for these properties

test/e2e/secretgencontroller/private_registry_auth_test.go
```diff
@@ -0,0 +1,260 @@
...
+  - secretRef:
+      name: pkg-demo-values
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: pkg-demo-values
+stringData:
+  values.yml: |
+    hello_msg: "hi"
```
cppforlife: no need for data values

test/e2e/secretgencontroller/private_registry_auth_test.go
```diff
@@ -0,0 +1,260 @@
...
+		kubectl.Run([]string{"wait", "--for=condition=ReconcileSucceeded", "pkgi/" + name, "--timeout", "1m"})
+	})
+
+	logger.Section("Check placeholder secret created", func() {
+		kubectl.Run([]string{"get", "secret", name + "-fetch0"})
+	})
+
+	logger.Section("Check placeholder secret deleted after PackageInstall deleted", func() {
+		cleanUp()
+		out, err := kubectl.RunWithOpts([]string{"get", "secret", name + "fetch0"}, e2e.RunOpts{AllowError: true})
```
cppforlife: this could be all kinds of errors. want to check on NotFound error.

test/e2e/secretgencontroller/private_registry_auth_test.go
```diff
@@ -0,0 +1,260 @@
...
+		kubectl.RunWithOpts([]string{"delete", "apps/" + name}, e2e.RunOpts{AllowError: true})
+		kapp.Run([]string{"delete", "-a", registryName, "-n", registryNamespace})
+		kapp.Run([]string{"delete", "-a", name})
+		kapp.Run([]string{"delete", "-a", "secret-export"})
+	}
+	cleanUp()
+	defer cleanUp()
+
+	logger.Section("deploy registry with self signed certs", func() {
+		kapp.Run([]string{"deploy", "-f", "../assets/registry/registry2.yml", "-f", "../assets/registry/certs-for-skip-tls.yml", "-f", "../assets/registry/htpasswd-auth", "-f", "../assets/registry/registry-contents.yml", "-a", registryName})
```
cppforlife, line 186: any reason why we just dont use entire directory for -f, or combine all registry config into a single file?

test/e2e/secretgencontroller/private_registry_auth_test.go
```diff
@@ -0,0 +1,260 @@
...
+			e2e.RunOpts{StdinReader: strings.NewReader(pkgiYaml)})
+		out := kapp.Run([]string{"inspect", "-a", name, "--raw", "--tty=false", "--filter-kind=PackageInstall"})
+
+		var cr v1alpha12.PackageInstall
+		err := yaml.Unmarshal([]byte(out), &cr)
+		if err != nil {
+			t.Fatalf("Failed to unmarshal: %s", err)
+		}
+
+		require.Equal(t, "", cr.Status.UsefulErrorMessage)
```
cppforlife: seems unnecessary given below waits for success and check on configmap.

test/e2e/secretgencontroller/private_registry_auth_test.go
```diff
@@ -0,0 +1,260 @@
...
+	})
+
+	logger.Section("Check PackageInstall/App succeed", func() {
```
cppforlife: i typically recommend to have logger section cover action+check. so lets combine into one logger section.

test/e2e/secretgencontroller/private_registry_auth_test.go
danielhelfand: It's mainly to check that the PackageInstall never starts in a failing state. If we only rely on the kubectl wait, then it's possible it will start failing but eventually recover after the placeholder secret is populated.

pkg/app/app_fetch.go
danielhelfand: Would it be enough to have the App check that it has secrets in the particular format of the placeholder secret (i.e. `%s-fetch-%d`)? Otherwise, would this information be on the App status or somewhere else?

test/e2e/secretgencontroller/private_registry_auth_test.go
danielhelfand: I think there may be additional test assets under that folder which are shared with other tests.

pkg/packageinstall/packageinstall.go
danielhelfand: The issue I forsee with not including is how we deal with updates to secrets after the PackageInstall is initially created? What is our recommended flow for having the PackageInstall's placeholder secrets include secret updates?

pkg/app/app_fetch.go
danielhelfand: This probably does not actually take 5s. Five iterations was chosen arbitrarily.

pkg/app/app_fetch.go
danielhelfand: Added to cleanup list in #336

pkg/app/app_fetch.go
danielhelfand: Added to #336. In a simple case, we observed there can be a 3s difference between PackageInstall being created and placeholder secret being populated.

cmd/controller/run.go
```diff
@@ -163,11 +163,7 @@ func Run(opts Options, runLog logr.Logger) {
 
 	{ // add controller for PackageInstall
 		pkgInstallCtrlOpts := controller.Options{
-			Reconciler: &PackageInstallReconciler{
-				kcClient:  kcClient,
-				pkgClient: pkgClient,
-				log:       runLog.WithName("ipr"),
-			},
+			Reconciler:              NewPackageInstallReconciler(kcClient, pkgClient, coreClient, runLog.WithName("ipr")),
```
joe-kimmel-vmw, line 166: do we want to update the arg here to `WithName("PackageInstallReconciler")` ?   or `pir` ?

danielhelfand: Yes, but maybe outside this pr preferably.

joe-kimmel-vmw: ok

cppforlife: dont want to have noisy logs with PackageInstallReconciler on every single line, so "pir" is fine by me.

PR #332, opened 2021-08-12
Fix rapid reconciliation

ewrenn8:
Case exists where IsReadyAt will use a stale deploy time,
causing it to always return true and reconcile every two seconds.
Check if deploy status is stale before using the time from it

Authored-by: Eli Wrenn <ewrenn@pivotal.io>

### DIFF
--- a/pkg/app/reconcile_timer.go
+++ b/pkg/app/reconcile_timer.go
@@ -4,10 +4,12 @@
 package app
 
 import (
+	"fmt"
 	"math"
 	"time"
 
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
@@ -39,19 +41,9 @@ func (rt ReconcileTimer) IsReadyAt(timeAt time.Time) bool {
 		return false
 	}
 
-	var lastReconcileTime time.Time
-
-	// Use latest deploy time if available, otherwise fallback to fetch
-	// If no timestamp is available, enqueue immediately
-	lastDeploy := rt.app.Status.Deploy
-	lastFetch := rt.app.Status.Fetch
-	if lastDeploy != nil && !lastDeploy.UpdatedAt.Time.IsZero() {
-		lastReconcileTime = rt.app.Status.Deploy.UpdatedAt.Time
-	} else {
-		if lastFetch == nil {
-			return true
-		}
-		lastReconcileTime = lastFetch.UpdatedAt.Time
+	lastReconcileTime, err := rt.lastReconcileTime()
+	if err != nil {
+		return true
 	}
 
 	if rt.hasReconcileStatus(v1alpha1.ReconcileFailed) {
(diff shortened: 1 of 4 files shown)


pkg/app/reconcile_timer.go
```diff
@@ -40,12 +40,23 @@ func (rt ReconcileTimer) IsReadyAt(timeAt time.Time) bool {
...
 	lastFetch := rt.app.Status.Fetch
-	if lastDeploy != nil && !lastDeploy.UpdatedAt.Time.IsZero() {
+	lastTemplate := rt.app.Status.Template
+
+	// check if we have deployed at least once in the past
+	haveDeployed := lastDeploy != nil && !lastDeploy.UpdatedAt.Time.IsZero()
+
+	// check to see if we are failing in either fetch or template. If we are, the time
+	// on the deploy struct might be stale, since it was updated the last time we
+	// __successful__ completed both fetch and template
```
joe-kimmel-vmw: nit :  truncated spelling of "successfully"

pkg/app/reconcile_timer_test.go
```diff
@@ -193,3 +193,33 @@ func TestFailedIsReadyAt(t *testing.T) {
...
+			},
+			ConsecutiveReconcileFailures: 1,
+			GenericStatus: v1alpha1.GenericStatus{
+				Conditions: []v1alpha1.AppCondition{{Type: v1alpha1.ReconcileFailed}},
+			},
+		},
+	}
+
+	isReady := NewReconcileTimer(app).IsReadyAt(timeOfReady.Add(1 * time.Second))
+	if isReady {
```
joe-kimmel-vmw: nit: let's start using assert/require libs, e.g. 
`assert.False(t, isReady, "Expected app not to be ready because deploy time is stale")`

pkg/pkgrepository/reconcile_timer.go
```diff
@@ -41,11 +41,23 @@ func (rt ReconcileTimer) IsReadyAt(timeAt time.Time) bool {
...
 	lastFetch := rt.app.Status.Fetch
-	if lastDeploy != nil && !lastDeploy.UpdatedAt.Time.IsZero() {
+	lastTemplate := rt.app.Status.Template
+
+	// check if we have deployed at least once in the past
+	haveDeployed := lastDeploy != nil && !lastDeploy.UpdatedAt.Time.IsZero()
+
+	// check to see if we are failing in either fetch or template. If we are, the time
+	// on the deploy struct might be stale, since it was updated the last time we
+	// __successful__ completed both fetch and template
```
joe-kimmel-vmw: (same)

pkg/pkgrepository/reconcile_timer_test.go
```diff
@@ -193,3 +193,33 @@ func TestFailedIsReadyAt(t *testing.T) {
...
+			ConsecutiveReconcileFailures: 1,
+			GenericStatus: v1alpha1.GenericStatus{
+				Conditions: []v1alpha1.AppCondition{{Type: v1alpha1.ReconcileFailed}},
+			},
+		},
+	}
+
+	isReady := NewReconcileTimer(app).IsReadyAt(timeOfReady.Add(1 * time.Second))
+	if isReady {
+		t.Fatalf("Expected app not to be ready, because deploy time is stale")
```
joe-kimmel-vmw: (same),   and maybe an obligatory "have we already weighed the costs/benefits of DRY-ing out the reconcile timers?"

ewrenn8: Agreed, I think DRYing these up would be good, but it should probably be a piece of a lager effort to DRY up the entire package repository impl (its currently a copy and paste of the app package).

cppforlife: intent to keep them separate (we copy-pasta them intentionally) because we should be moving towards package repository be decoupled from CRs-as-storage so current implementation is temporary.

pkg/app/reconcile_timer.go
```diff
@@ -104,3 +96,37 @@ func (rt ReconcileTimer) applyJitter(t time.Duration) time.Duration {
 	const appJitter time.Duration = 5 * time.Second
 	return t - appJitter + wait.Jitter(appJitter, 1.0)
 }
+
+func (rt ReconcileTimer) lastReconcileTime() (time.Time, error) {
+	lastDeploy := rt.app.Status.Deploy
+	lastFetch := rt.app.Status.Fetch
+	lastTemplate := rt.app.Status.Template
```
cppforlife: nit: can we inline these? would remove extra indirection - extra mind jump.

PR #333, opened 2021-08-12
support "ext.packaging.carvel.dev/fetch-%d-secret-name" on PackageInstall to set secret on underlying App CR

cppforlife:


pkg/packageinstall/app_test.go
```diff
@@ -364,3 +364,88 @@ func TestAppManuallyControlled(t *testing.T) {
...
+		},
+	}
+
+	// Not interesting in metadata in this test
+	app.ObjectMeta = metav1.ObjectMeta{}
+
+	if !reflect.DeepEqual(expectedApp, app) {
+		bs, _ := yaml.Marshal(app)
+		t.Fatalf("App does not match expected app: (actual)\n%s", bs)
+	}
```
danielhelfand: Let's try and use the new testify library here to provide a better diff.

pkg/packageinstall/app_test.go
```diff
@@ -364,3 +364,88 @@ func TestAppManuallyControlled(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	}
+
+	app, err := packageinstall.NewApp(&kcv1alpha1.App{}, ipkg, pkgVersion)
+	if err != nil {
+		t.Fatalf("Expected no err, but was: %s", err)
+	}
```
danielhelfand: Let's try and use the new testify library here: `require.NotNil(t, err)`

joe-kimmel-vmw:
I have some newbie questions about larger context:   is there a "user story" for "I need to associate secrets to specific fetch stages" that we currently aren't serving at all?   or in other words is there already a partial solution for a more common version of this end-user desire,  or is this going to be our answer for all such cases?

cppforlife:
@joe-kimmel-vmw this is a quick&dirty way to expose users of PackageInstall to be able to set creds when installing a package. there is no way to do that today from the package consumer side (only from package author side since they can hard code some secret ref name). we may introduce first class support for specifying fetch secrets on PackageInstall in future versions.

joe-kimmel-vmw:
> [...] from the package consumer side (only from package author side since [...]

cool,  thanks - i think ⬆️  is what i was missing: possibilities as a consumer vs. as an author.   

anyways agree it's a little on the hacky side but i can't think of a better proposal atm.   within that (and I do agree with @danielhelfand  that the testify libraries give better error messages),   LGTM

Issue #337, opened 2021-08-18
Expose reconcile status as metrics to allow monitoring 


mhoshi-vm:
**Describe the problem/challenge you have**
One thing we need is at least be able to expose the reconcile status to an external monitoring sysytem. We have seen a cluster that was running fine for several days until we realized that it was actually failing to reconcile due to someone mistakenly changed the service account role.

This cannot be identified until somebody executes `kubectl get apps`

**Describe the solution you'd like**
Something like `kubernetes_apps_reconcile_status` metrics being exposed where 0 = success , 1 = fail , 2 = unknown then i think it's easy to monitor

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

mhoshi-vm:
Hi, 
I am currently trying to develop a pull request. Will try to submit in few days for review.

aaronshurley:
Thanks for creating this issue and submitting a PR!

cppforlife:
PR was merged and shipped some time ago. thank you again @mhoshi-vm

PR #346, opened 2021-08-23
Add metrics to monitor reconcile status

mhoshi-vm:
Feature Request for #337.
Add additional counter metrics to count the number reconciles with failures also.

- Metrics will get initialised to all 0 during reconcile
- The counter increments when status changes to "Reconciling" > kapp_reconcile_attempt_total
- The counter increments when status changes to "Reconcile Succeed" > kapp_reconcile_success_total
- The counter increments when status changes to "Reconcile Failed" > kapp_reconcile_failure_total
- The counter increments when status changes to "Deleting" > kapp_reconcile_delete_attempt_total
- The counter increments when status changes to "Delete Failed" > kapp_reconcile_delete_failed_total
- The metrics get deleted when the delete completes. This is to avoid metrics existing forever while the real app has been deleted.


cmd/controller/apps_reconciler.go
```diff
@@ -47,6 +48,7 @@ func (r *AppsReconciler) Reconcile(ctx context.Context, request reconcile.Reques
 	}
 
 	crdApp := r.appFactory.NewCRDApp(existingApp, log)
+	metrics.InitMetrics(existingApp.Name, existingApp.Namespace)
```
cppforlife: i was looking at bit more around what code is required to run this metrics service. i do think we should inject an object that holds appropriate fields. seems like having it be a singleton is not a great thing.

cppforlife: would you be interested in tweaking it to make metrics a separate objects, or we could potentially change that ourselves?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,135 @@
...
+// RegisterReconcileSuccess increments reconcileSuccessTotal
+func RegisterReconcileSuccess(appName string, namespace string) {
+	if c, ok := sm.metrics[reconcileSuccessTotal].(*prometheus.CounterVec); ok {
+		c.WithLabelValues(appName, namespace).Inc()
+	}
+}
+
+// RegisterReconcileFailure increments reconcileFailureTotal
+func RegisterReconcileFailure(appName string, namespace string) {
+	if c, ok := sm.metrics[reconcileFailureTotal].(*prometheus.CounterVec); ok {
```
cppforlife: i was looking at antrea metrics [0], and it seems that instead of casting it's better to have that variable be typed.

[0] https://github.com/antrea-io/antrea/blob/main/pkg/agent/proxy/metrics/metrics.go#L29

cmd/controller/run.go
```diff
@@ -92,6 +93,10 @@ func Run(opts Options, runLog logr.Logger) {
 		os.Exit(1)
 	}
 
+	runLog.Info("setting up metrics")
+	servMetrics := metrics.NewServerMetrics()
+	servMetrics.RegisterAllMetrics()
+
 	// assign bindPort to env var KAPPCTRL_API_PORT if available
```
mhoshi-vm, line 89: I'd like to help as much as possible,  as this helps me understand kapp-controller a lot better :) but please let me know if you believe you can solve this much faster.

> i do think we should inject an object that holds appropriate fields. seems like having it be a singleton is not a great thing.

Hope this is what you mean

pkg/metrics/metrics.go
mhoshi-vm: I may not understand this, can you elaborate a bit more ?
BTW, the current structure is inspired a lot from velero metrics.
https://github.com/vmware-tanzu/velero/blob/main/pkg/metrics/metrics.go

cmd/controller/app_factory.go
```diff
@@ -23,11 +24,12 @@ type AppFactory struct {
 	kcConfig   *config.Config
 }
 
-func (f *AppFactory) NewCRDApp(app *kcv1alpha1.App, log logr.Logger) *ctlapp.CRDApp {
+func (f *AppFactory) NewCRDApp(app *kcv1alpha1.App, log logr.Logger, servMetrics *metrics.ServerMetrics) *ctlapp.CRDApp {
 	fetchFactory := fetch.NewFactory(f.coreClient, f.kcConfig)
 	templateFactory := template.NewFactory(f.coreClient, fetchFactory)
 	deployFactory := deploy.NewFactory(f.coreClient)
-	return ctlapp.NewCRDApp(app, log, f.appClient, fetchFactory, templateFactory, deployFactory)
+	servMetrics.InitMetrics(app.Name, app.Namespace)
```
mhoshi-vm: Still not sure where is the most appropriate place to init the metrics...

pkg/metrics/metrics.go
cppforlife: > .(*prometheus.CounterVec)

casting is typically avoided where unnecessary. the only reason why it's necessary here is because you added all these metrics into an object that "hides" their real type (ie uses generic type). in antrea example individual metrics maintain their concrete type [0] hence do not require casting (which is good).

[0] https://github.com/antrea-io/antrea/blob/main/pkg/agent/proxy/metrics/metrics.go#L29

mhoshi-vm: Please review my latest commit. Hope I understood it correctly...

cmd/controller/apps_reconciler.go
mhoshi-vm: Please review my latest commit. Hope I understood it correctly...

cmd/controller/run.go
cppforlife: yup, looks like you got it!

cmd/controller/app_factory.go
```diff
@@ -23,11 +24,13 @@ type AppFactory struct {
 	kcConfig   *config.Config
 }
 
-func (f *AppFactory) NewCRDApp(app *kcv1alpha1.App, log logr.Logger) *ctlapp.CRDApp {
+// NewCRDApp creates new CRD app
+func (f *AppFactory) NewCRDApp(app *kcv1alpha1.App, log logr.Logger, servMetrics *metrics.ServerMetrics) *ctlapp.CRDApp {
```
cppforlife: instead of passing down `servMetrics *metrics.ServerMetrics` to this function, lets give metrics into constructor. this is why factories are useful -- to shield callers from knowing dependencies (metrics) of objects being built (app).

cmd/controller/app_factory.go
cppforlife: probably somewhere in the reconcile method of the CRDApp

cmd/controller/apps_reconciler.go
```diff
@@ -5,10 +5,10 @@ package controller
 
 import (
 	"context"
-
```
cppforlife: nit: keep newline (we separate std lib imports from everything else)

cmd/controller/apps_reconciler.go
```diff
@@ -18,24 +18,28 @@ import (
...
 	log             logr.Logger
+	servMetrics     *metrics.ServerMetrics
 	appFactory      AppFactory
 	appRefTracker   *reftracker.AppRefTracker
 	appUpdateStatus *reftracker.AppUpdateStatus
 }
 
-func NewAppsReconciler(appClient kcclient.Interface, log logr.Logger, appFactory AppFactory,
+// NewAppsReconciler reconciles new app
+func NewAppsReconciler(appClient kcclient.Interface, log logr.Logger, servMetrics *metrics.ServerMetrics, appFactory AppFactory,
```
cppforlife: assuming you pass metrics into AppFactory changes to this file are not necessary.

cmd/controller/run.go
```diff
@@ -5,6 +5,7 @@ package controller
 
 import (
 	"fmt"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/metrics"
```
cppforlife: nit: please add to below section (not std lib import)

pkg/app/app.go
```diff
@@ -4,6 +4,7 @@
 package app
 
 import (
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/metrics"
```
cppforlife: nit: please add to below section (not std lib import)

pkg/app/app_reconcile_test.go
```diff
@@ -23,11 +24,15 @@ import (
 func Test_NoInspectReconcile_IfNoDeployAttempted(t *testing.T) {
 	log := logf.Log.WithName("kc")
 
+	servMetrics := metrics.NewServerMetrics()
```
cppforlife: for tests lets create NoopMetrics that would implement new Metrics interface we define (that would also be satisfied by ServerMetrics).

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,142 @@
...
+
+const (
+	metricNamespace = "kapp"
+
+	// kapp metrics
+	reconcileAttemptTotal       = "reconcile_attempt_total"
+	reconcileSuccessTotal       = "reconcile_success_total"
+	reconcileFailureTotal       = "reconcile_failure_total"
+	reconcileDeleteAttemptTotal = "reconcile_delete_attempt_total"
+	reconcileDeleteFailedTotal  = "reconcile_delete_failed_total"
```
cppforlife: these could be inlined since they are only used once.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,142 @@
...
+	reconcileDeleteAttemptTotal *prometheus.CounterVec
+	reconcileDeleteFailedTotal  *prometheus.CounterVec
+}
+
+var (
+	once sync.Once
+)
+
+const (
+	metricNamespace = "kapp"
```
cppforlife: it's not really kapp, but kapp-controller. could do `kappctrl`

cppforlife: i would also add that as a const into ServerMetrics constructor.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// kapp metrics
+	reconcileAttemptTotal       = "reconcile_attempt_total"
+	reconcileSuccessTotal       = "reconcile_success_total"
+	reconcileFailureTotal       = "reconcile_failure_total"
+	reconcileDeleteAttemptTotal = "reconcile_delete_attempt_total"
+	reconcileDeleteFailedTotal  = "reconcile_delete_failed_total"
+
+	// Labels
+	kappNameLabel      = "appName"
```
cppforlife: should this be `app_name`?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,142 @@
...
+	kappNamespaceLabel = "namespace"
```
cppforlife: lets also move these into NewServerMetrics as consts. nobody else should need them.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+// NewServerMetrics creates ServerMetrics object
+func NewServerMetrics() *ServerMetrics {
```
cppforlife: since these metrics are really relevant to App CRs, should this be called ServerMetrics  -> AppMetrics?

also im not too familiar with prometheus conventions: how should we indicate that these metrics are app related? should that be part of namespace: kappctrl-apps? or reconcile_attempt_total -> app_reconcile_attempt_total?

cppforlife (commented):
starting to look really good!

pkg/metrics/metrics.go
mhoshi-vm: Reading this as namespace is a single word, i guess "kappctrl" should be namespace.
https://prometheus.io/docs/practices/naming/

cmd/controller/app_factory.go
mhoshi-vm: Moved, to reconcile in CRDApps

cmd/controller/app_factory.go
mhoshi-vm: Updated code, hope I got it right

cmd/controller/apps_reconciler.go
mhoshi-vm: Fixed

cmd/controller/apps_reconciler.go
mhoshi-vm: Hope i got this right also by fixing appfactory

cmd/controller/run.go
mhoshi-vm: fixed

pkg/metrics/metrics.go
mhoshi-vm: Updated

pkg/app/app.go
mhoshi-vm: fixed

pkg/app/app_reconcile_test.go
mhoshi-vm: I hope i got this right....

pkg/metrics/metrics.go
mhoshi-vm: Fixed

pkg/metrics/metrics.go
mhoshi-vm: Fixed

pkg/metrics/metrics.go
mhoshi-vm: Fixed

pkg/metrics/metrics.go
mhoshi-vm: Also renamed ServerMetrics > AppMetrics

pkg/app/crd_app.go
```diff
@@ -6,6 +6,7 @@ package app
 import (
 	"context"
 	"fmt"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/metrics"
```
cppforlife: nit: move into next import subsection

pkg/metrics/app_metrics.go
```diff
@@ -0,0 +1,133 @@
...
+	reconcileDeleteAttemptTotal *prometheus.CounterVec
+	reconcileDeleteFailedTotal  *prometheus.CounterVec
+}
+
+var (
+	once sync.Once
+)
+
+// NewServerMetrics creates AppMetrics object
+func NewServerMetrics() *AppMetrics {
```
cppforlife: nit: NewServerMetrics -> NewAppMetrics

pkg/metrics/app_metrics.go
```diff
@@ -0,0 +1,133 @@
...
+	metricNamespace := "kappctrl"
```
cppforlife: nit: you can actually use const within functions as well e.g. 

```go
const (
	metricNamespace = "kappctrl"
	kappNameLabel = "app_name"
	kappNamespaceLabel = "namespace"
)
```

cppforlife:
might have to rebase...

pkg/app/crd_app.go
```diff
@@ -114,6 +117,7 @@ func (a *CRDApp) updateApp(updateFunc func(*kcv1alpha1.App)) error {
 }
 
 func (a *CRDApp) Reconcile(force bool) (reconcile.Result, error) {
+	a.appMetrics.InitMetrics(a.app.Name(), a.app.Namespace())
```
cppforlife: i think it might be better to move this call into app_reconcile.go. it'll be closer to other calls to appMetrics.

pkg/app/crd_app.go
mhoshi-vm: updated

pkg/metrics/app_metrics.go
mhoshi-vm: updated

pkg/metrics/app_metrics.go
mhoshi-vm: updated

pkg/app/crd_app.go
mhoshi-vm: I put it in to app_reconcile.go, however wont this be called every time a reconcile operation is performed ?
Simple test shows fine but I think that is because it adds 0 for counters.

pkg/app/app_reconcile_test.go
```diff
@@ -5,6 +5,8 @@ package app
 
 import (
 	"fmt"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/metrics"
```
cppforlife: nit: next import section

pkg/app/app_reconcile_test.go
```diff
@@ -20,14 +22,54 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+var noopMetrics = &metrics.AppMetrics{
```
cppforlife: i dont think you have to reconstruct AppMetrics here. can just use NewAppMetrics() -- just dont register them.

pkg/metrics/app_metrics.go
```diff
@@ -0,0 +1,133 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+// AppMetrics holds server metrics
+type AppMetrics struct {
+	ReconcileAttemptTotal       *prometheus.CounterVec
+	ReconcileSuccessTotal       *prometheus.CounterVec
+	ReconcileFailureTotal       *prometheus.CounterVec
+	ReconcileDeleteAttemptTotal *prometheus.CounterVec
+	ReconcileDeleteFailedTotal  *prometheus.CounterVec
```
cppforlife: all these fields should be private.

pkg/app/crd_app.go
cppforlife: > Simple test shows fine but I think that is because it adds 0 for counters.

yeah it would. and adding 0 i guess is fine. factory's New method is also called every time, so same behaviour

cppforlife:
super close @mhoshi-vm!

pkg/metrics/app_metrics.go
mhoshi-vm: fixed

pkg/app/app_reconcile_test.go
mhoshi-vm: ficed

pkg/app/app_reconcile_test.go
mhoshi-vm: fixed

mhoshi-vm:
I should have asked at the beginning but do I need to implement some test code ?

mhoshi-vm:
> I should have asked at the beginning but do I need to implement some test code ?

Guessing I need to build something like the following.
Would very appreciate if I can get some guide (where I should place the test code) in the current kapp controller structure
https://github.com/antrea-io/antrea/blob/4aaa90edf1f2d8b70282f55511b860765c342801/test/e2e/prometheus_test.go

cppforlife:
@mhoshi-vm thanks for sticking with it! 🔥🔥🔥

PR #353, opened 2021-08-28
various enhancements

cppforlife:
review via commits view

### DIFF
--- a/cmd/controller/run.go
+++ b/cmd/controller/run.go
@@ -164,7 +164,7 @@ func Run(opts Options, runLog logr.Logger) {
 	{ // add controller for PackageInstall
 		pkgInstallCtrlOpts := controller.Options{
 			Reconciler:              NewPackageInstallReconciler(kcClient, pkgClient, coreClient, runLog.WithName("ipr")),
-			MaxConcurrentReconciles: opts.Concurrency,
+			MaxConcurrentReconciles: 1,
 		}
 
 		pkgInstallCtrl, err := controller.New("kapp-controller-packageinstall", mgr, pkgInstallCtrlOpts)
--- a/pkg/apis/packaging/v1alpha1/package_install.go
+++ b/pkg/apis/packaging/v1alpha1/package_install.go
@@ -93,4 +93,8 @@ type PackageInstallStatus struct {
 	// TODO this is desired resolved version (not actually deployed)
 	// +optional
 	Version string `json:"version,omitempty"`
+	// LastAttemptedVersion specifies what version was last attempted to be installed.
+	// It does _not_ indicate it was successfully installed.
+	// +optional
+	LastAttemptedVersion string `json:"lastAttemptedVersion,omitempty"`
 }
--- a/pkg/packageinstall/packageinstall.go
+++ b/pkg/packageinstall/packageinstall.go
@@ -124,7 +124,8 @@ func (pi *PackageInstallCR) reconcile(modelStatus *reconciler.Status) (reconcile
 
 	pi.model.Status.LastAttemptedVersion = pkg.Spec.Version
 
-	existingApp, err := pi.kcclient.KappctrlV1alpha1().Apps(pi.model.Namespace).Get(context.Background(), pi.model.Name, metav1.GetOptions{})
+	existingApp, err := pi.kcclient.KappctrlV1alpha1().Apps(pi.model.Namespace).Get(
+		context.Background(), pi.model.Name, metav1.GetOptions{})
 	if err != nil {
 		if errors.IsNotFound(err) {
 			pkgWithPlaceholderSecrets, err := pi.reconcileFetchPlaceholderSecrets(pkg)
@@ -176,7 +177,8 @@ func (pi *PackageInstallCR) reconcileAppWithPackage(existingApp *kcv1alpha1.App,
 	}
 
 	if !equality.Semantic.DeepEqual(desiredApp, existingApp) {
-		_, err = pi.kcclient.KappctrlV1alpha1().Apps(desiredApp.Namespace).Update(context.Background(), desiredApp, metav1.UpdateOptions{})
+		_, err = pi.kcclient.KappctrlV1alpha1().Apps(desiredApp.Namespace).Update(
+			context.Background(), desiredApp, metav1.UpdateOptions{})
 		if err != nil {
 			return reconcile.Result{Requeue: true}, err
 		}
@@ -192,7 +194,8 @@ func (pi *PackageInstallCR) referencedPkgVersion() (datapkgingv1alpha1.Package, 
 
 	semverConfig := pi.model.Spec.PackageRef.VersionSelection
 
-	pkgList, err := pi.pkgclient.DataV1alpha1().Packages(pi.model.Namespace).List(context.Background(), metav1.ListOptions{})
+	pkgList, err := pi.pkgclient.DataV1alpha1().Packages(pi.model.Namespace).List(
+		context.Background(), metav1.ListOptions{})
 	if err != nil {
 		return datapkgingv1alpha1.Package{}, err
 	}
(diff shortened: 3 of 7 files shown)


cmd/controller/run.go
```diff
@@ -164,7 +164,7 @@ func Run(opts Options, runLog logr.Logger) {
 	{ // add controller for PackageInstall
 		pkgInstallCtrlOpts := controller.Options{
 			Reconciler:              NewPackageInstallReconciler(kcClient, pkgClient, coreClient, runLog.WithName("ipr")),
-			MaxConcurrentReconciles: opts.Concurrency,
+			MaxConcurrentReconciles: 1,
```
danielhelfand, line 167: What is the motivation behind this change?

Wouldn't this be a breaking change if we remove this? Should we maybe consider a formal deprecation process?

pkg/packageinstall/packageinstall.go
```diff
@@ -28,6 +28,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	// DowngradableAnnKey specifies annotation that user can place on
```
joe-kimmel-vmw, line 32: in this diff it's easy to see how this gets used as `canDowngrade` so it's just a boolean flag.  Would we ever allow someone to specify a version floor,  e.g. `"canDowngradeUntil": "1.0.1"`. indicating that they won't accept a version lower than the value?

pkg/apis/packaging/v1alpha1/package_install.go
```diff
@@ -93,4 +93,8 @@ type PackageInstallStatus struct {
 	// TODO this is desired resolved version (not actually deployed)
 	// +optional
 	Version string `json:"version,omitempty"`
+	// LastAttemptedVersion specifies what version was last attempted to be installed.
+	// It does _not_ indicate it was successfully installed.
+	// +optional
+	LastAttemptedVersion string `json:"lastAttemptedVersion,omitempty"`
```
danielhelfand, line 99: Would we want to eventually expand this and make it a struct that would also include if it was successfully installed? Should there be a follow up issue here?

pkg/packageinstall/packageinstall.go
```diff
@@ -284,25 +355,25 @@ func (pi *PackageInstallCR) update(updateFunc func(*pkgingv1alpha1.PackageInstal
 }
 
 func (pi *PackageInstallCR) reconcileFetchPlaceholderSecrets(pkg datapkgingv1alpha1.Package) (datapkgingv1alpha1.Package, error) {
-	pkgCopy := pkg.DeepCopy()
-	for i, fetch := range pkgCopy.Spec.Template.Spec.Fetch {
+	pkg = *pkg.DeepCopy()
```
joe-kimmel-vmw, line 358: imo it's more clear to name it `pkgCopy`  as it eliminates all possible confusions

pkg/packageinstall/packageinstall.go
```diff
@@ -91,9 +98,34 @@ func (pi *PackageInstallCR) reconcile(modelStatus *reconciler.Status) (reconcile
 		return reconcile.Result{Requeue: true}, err
 	}
 
+	// Set new desired version before checking if it's not applicable
 	pi.model.Status.Version = pkg.Spec.Version
 
-	existingApp, err := pi.kcclient.KappctrlV1alpha1().Apps(pi.model.Namespace).Get(context.Background(), pi.model.Name, metav1.GetOptions{})
+	_, canDowngrade := pi.model.Annotations[DowngradableAnnKey]
+	if !canDowngrade && len(pi.model.Status.LastAttemptedVersion) > 0 {
```
danielhelfand: Would `pi.model.Status.LastAttemptedVersion != ""` read a bit clearer here?

pkg/packageinstall/packageinstall_downgrade_test.go
```diff
@@ -0,0 +1,257 @@
...
+
+		assert.Equal(t, "ver-1.0.0", getApp(t, appClient, "instl-pkg").Spec.Fetch[0].ImgpkgBundle.Image)
+	})
+}
+
+func getPackageInstall(t *testing.T, clientset *fakekappctrl.Clientset, name string) *pkgingv1alpha1.PackageInstall {
+	gvr := schema.GroupVersionResource{"packaging.carvel.dev", "v1alpha1", "packageinstalls"}
+	obj, err := clientset.Tracker().Get(gvr, "", name)
+	assert.Nil(t, err)
+	assert.NotNil(t, obj)
```
danielhelfand: I think `require.NotNil` would be helpful here to avoid nil being returned

pkg/packageinstall/packageinstall_downgrade_test.go
```diff
@@ -0,0 +1,257 @@
...
+	return obj.(*pkgingv1alpha1.PackageInstall)
+}
+
+func getApp(t *testing.T, clientset *fakekappctrl.Clientset, name string) *v1alpha1.App {
+	gvr := schema.GroupVersionResource{"kappctrl.k14s.io", "v1alpha1", "apps"}
+	obj, err := clientset.Tracker().Get(gvr, "", name)
+	assert.Nil(t, err)
+	assert.NotNil(t, obj)
```
danielhelfand: I think `require.NotNil` would be helpful here to avoid nil being returned

joe-kimmel-vmw (commented):
just one general question and a nit re. naming things

cmd/controller/run.go
cppforlife: opts.Concurrency is used for App CRs. probably was copy pasted for PackageInstalls.

> Wouldn't this be a breaking change if we remove this?

im not sure i see what it breaks.

pkg/packageinstall/packageinstall.go
cppforlife: it leads to potential of modifying and/or returning the wrong thing.

pkg/apis/packaging/v1alpha1/package_install.go
cppforlife: i would say not... at least at this point. if we would want to track last successful version, then we would probably introduce status.LastSuccessfullyReconciledVersion.

pkg/packageinstall/packageinstall.go
cppforlife: yeah i was thinking to support that, but i couldnt find concrete use cases. the main goal is to allow spec.versionSelection to be able to select version lower than what was previously attempted. ultimately spec.versionSelection is the thing that specifies version floor of whats allowed.

pkg/packageinstall/packageinstall_downgrade_test.go
cppforlife: fixed.

pkg/packageinstall/packageinstall_downgrade_test.go
cppforlife: fixed.

pkg/packageinstall/packageinstall.go
cppforlife: updated.

joe-kimmel-vmw (approved):

cmd/controller/run.go
danielhelfand: Are we treating this as a bug then? My thought is that this is a configurable property that is now being made unconfigurable and that is what is considered breaking.

danielhelfand: > i should have added to the comment: currently it seems that multiple items will get reconciled in parallel. that i think also includes same items which would in theory introduce conflicts in updates. i just set it to 1 explicitly (that's actually controller runtime default afaik) because unlike App reconciler, reconciling PackageInstall should take millis/seconds.

This is considered not breaking in follow up convo. Closing this out.

danielhelfand (approved):
lgtm

PR #356, opened 2021-08-31
Reconciler split

cppforlife:
- moved reconcilers to their related packages

(easier to review by commits.)


pkg/app/apps_reconciler_test.go
```diff
@@ -1,14 +1,14 @@
...
-package controller_test
+package app_test
 
 import (
 	"reflect"
 	"testing"
 
-	"github.com/vmware-tanzu/carvel-kapp-controller/cmd/controller"
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
+	app_ "github.com/vmware-tanzu/carvel-kapp-controller/pkg/app"
```
danielhelfand: Can we use something besides `_` here? Maybe even just call appDep or appPkg?

cmd/controller/run.go
```diff
@@ -70,157 +63,109 @@ func Run(opts Options, runLog logr.Logger) {
...
+		appFactory := app.AppFactory{coreClient, kcClient, kcConfig}
+		reconciler := app.NewAppsReconciler(kcClient, runLog.WithName("app"),
+			appFactory, refTracker, updateStatusTracker)
+
+		ctrl, err := controller.New("app", mgr, controller.Options{
 			Reconciler: NewUniqueReconciler(&ErrReconciler{
-				delegate: NewAppsReconciler(kcClient, runLog.WithName("ar"), appFactory, refTracker, updateStatusTracker),
-				log:      runLog.WithName("pr"),
+				delegate: reconciler,
+				log:      runLog.WithName("er"),
```
danielhelfand, line 116: ar?

pkg/app/app_factory.go
```diff
@@ -0,0 +1,28 @@
+// Copyright 2020 VMware, Inc.
```
danielhelfand, line 1: 2021?

joe-kimmel-vmw:
those godocs linters got you good,  but the changes LGTM.     I do agree with Daniel's request for a more descriptive name than "app_" but i have no suggestions,  maybe because i'm missing what else called "app" we might collide with in that file.

cmd/controller/run.go
cppforlife: err reconciler

pkg/app/app_factory.go
cppforlife: i was under impression we keep existing copyright notices when files get moved around.

cmd/controller/run.go
danielhelfand: I see. The diff was throwing me off.

pkg/app/app_factory.go
danielhelfand: Yeah, not sure. Fine to keep unless we know otherwise.

pkg/app/apps_reconciler_test.go
cppforlife: i think typically camelcase is not used in aliases. apppkg?

cppforlife: renamed to apppkg

PR #390, opened 2021-10-06
disable finding existing non-labeled resources for pkgr reconciliation

cppforlife:


pkg/pkgrepository/package_repo_app.go
```diff
@@ -4,13 +4,18 @@
...
 	"time"
 
 	kcv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
 	pkgingv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+var (
+	kappDebug = os.Getenv("KAPPCTRL_PKGR_KAPP_DEBUG") == "true"
+)
```
danielhelfand, line 17: Should this be carried out in `NewPackageRepoApp()`?

cppforlife: i figured ill add this _temporarily_ for debugging any high urgency issues => does not become part of "official" interface, so did not want to invest into dependency injecting it. relatedly we should probably sort out some guidelines on flags/env/config.

PR #399, opened 2021-10-12
Releasing to ghcr

### LINKED ISSUE #319: Move kapp-controller Image from DockerHub to New Registry to Avoid Rate Limiting
As kapp-controller continues to gain wider adoption, an issue that has come up is [DockerHub's rate limiting](https://www.docker.com/increase-rate-limits). In order to assure users can successfully pull the kapp-controller image when deploying kapp-controller to a Kubernetes cluster without hitting rate limits, the kapp-controller image should be relocated to a new image registry. 

There are no hard requirements on which registry should be used. There should be some initial research into DockerHub alternatives with the main requirement being avoiding rate limiting errors for kapp-controller users. 

One option that has been suggested is GitHub's container registry, which is free to public images: https://github.com/features/packages

joe-kimmel-vmw:
moves our artifact hosting to ghcr.io / github container registry.   This is because some people were having issues with the dockerhub rate limits: if many employees of a single org were all coming from behind the same VPN-assigned IP then they all shared the same dockerhub quota.

Fixes #319

After this PR is merged, users will pull subsequent release's OCI artifact from github's container registry,  instead of dockerhub --  all old releases will stay where they were, but going forward this will switch our release host. 
```release-note
Going forward all release OCI artifacts will be hosted on github;  this unifies all of our release artifacts to be on github.
```

Feel free to suggest that i delete the alpha release yaml - in some sense it's only there in case anybody wants to try installing it to ensure that the image really is public,  but we've already done this at least internally to the team (and the reviewers are all on the team, so, uh...)

Followup work to this PR should include automating our release through github actions. 

NOTE my expectation is that pull access to the packages is fully public, while push access is limited to the carvel-writers group.   Would be helpful if a reviewer could check access settings for the packages (e.g. by visually inspecting what's checked,  or even better by confirming access/denial).

```docs
Until we have releases run via github actions we'll have to do them ourselves, which isn't really a change.

To run a release manually you'll have to be logged into ghcr (not dockerhub). 
1.) Ensure you have a PAT, with read/write/delete permissions on packages (selecting those will auto-select repo permissions, and you don't need any further permissions attached to this token): https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token

2.) docker login to ghcr with your pat (as shown here: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry ) 

3.) run hack/release.sh

```


hack/version-util.sh
```diff
@@ -7,7 +7,7 @@
 
 # git tag version extraction graciously provided by https://github.com/vmware-tanzu/carvel-imgpkg/blob/develop/hack/build-binaries.sh
 function get_latest_git_tag {
-  git describe --tags | grep -Eo 'v[0-9]+\.[0-9]+\.[0-9]+'
+  git describe --tags | grep -Eo 'v[0-9]+\.[0-9]+\.[0-9]+(-alpha\.[0-9]+)?'
```
danielhelfand, line 10: Should we include `(-alpha\.[0-9]+)?`? 

My thought is that we received feedback on our releasing of alpha releases that it was not the best process for releasing packaging work more frequently. 

I am curious if it would be a better pattern moving forward to use something like `rc` and actually publish an official GitHub release.

This is also a non-blocking comment. Please feel free to tell me it makes sense to keep supporting this.

danielhelfand (commented):
Some questions that are non blocking:

1. Should we clean up images that were published as part of testing the migration to ghcr.io?
2. Should we also change [k14stest](https://github.com/vmware-tanzu/carvel-kapp-controller/blob/6672b18701223b6f76967af832d77cb1cad65047/config/values.yml#L12) to ghcr.io?

Otherwise, lgtm

joe-kimmel-vmw:
> 1. Should we clean up images that were published as part of testing the migration to ghcr.io?

I deleted the two that had never been downloaded, and removed the yaml for the third,  but I actually don't see a delete option?  Either I've forgotten how to use the UI since yesterday,  or we're not allowed to delete packages once they've been downloaded? Both possibilities seem possible.

> 2. Should we also change [k14stest](https://github.com/vmware-tanzu/carvel-kapp-controller/blob/6672b18701223b6f76967af832d77cb1cad65047/config/values.yml#L12) to ghcr.io?

Curious what anyone else thinks about this ---  the "upshot" is that i think those are dev/test builds, and that repo is more ~private.  I can see an argument against having our test builds be visible from our github (though maybe it would be a separate package, and we could permission that package differently).        The downside of leaving it on dockerhub is of course that then we're still using dockerhub.      Opinions from @cppforlife  @neil-hickey ?   others?

cppforlife:
> Should we also change k14stest to ghcr.io?

i like the convenience of dockerhub for myself, but i also can easily use -v image_repo when necessary.

i think having alpha/rc builds is useful when we are iterating on particular larger featureset. we can revisit that whenever that happens.

PR #402, opened 2021-10-15
Automate release process for kapp-controller

danielhelfand:

This pull request automates the release process for kapp-controller via a GitHub workflow. 

It pushes the kapp-controller image to ghcr.io, uploads the release.yml/checksum file information to the newly available release, verifies the assets built in the CI process match what was uploaded to GitHub.

N/A

```release-note
NONE
```

For more information on developer  experience of this, see updates to release process in dev.md.


.goreleaser.yml
```diff
@@ -0,0 +1,51 @@
...
+
+  # You can disable this pipe in order to not upload any artifacts.
+  # Defaults to false.
+  disable: false
+
+  # This uploads the kapp-controller release.yml to the
+  # release goreleaser will create
+  extra_files:
+    - glob: ./tmp/release.yml
+    - glob: ./tmp/checksums.txt
```
danielhelfand: Wondering how signing could be involved for uploading checksums file. goreleaser does support signing but haven't looked very deeply into it yet.

https://goreleaser.com/customization/sign/

.github/workflows/release.yml
```diff
@@ -0,0 +1,109 @@
...
+      - name: Run release script
+        run: |
+          set -e -x
+
+          ./hack/build-release.sh
+
+      - name: Create Release with goreleaser
+        uses: goreleaser/goreleaser-action@5e15885530fb01d81d1f24e8a6f54ebbd0fed7eb
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
danielhelfand: How do we feel on using token with these permissions? Is there a more specific one for pushing to ghcr.io.

.github/workflows/release.yml
```diff
@@ -0,0 +1,109 @@
...
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run release script
+        run: |
+          set -e -x
+
+          ./hack/build-release.sh
+
+      - name: Create Release with goreleaser
```
cppforlife: why do we need goreleaser in this case?

cppforlife: ah i see, seems like for release.yml?

cppforlife: might be a simpler alternative here -- somethign along the lines https://github.com/FiloSottile/age/blob/main/.github/workflows/build.yml

danielhelfand: Creating the release as the trigger mechanism would be useful in removing use of goreleaser, but would also add a manual step of creating a published release. 

I do think goreleaser does do some nice things (e.g. commit history summary) that I am not quite sure how to replicate with that example. But can look into alternative tools for that as well.

danielhelfand: After discussing further with @cppforlife, it sounds like the best thing to do is move away from goreleaser and support a workflow primarily focused on the following things:

1. Trigger off of publishing a release, rather than off of pushing tag
2. Upload the release.yml to the published release
3. Publish the checksum in the body of the GitHub release
4. Verify the assets (i.e. release.yml) created match via checksum comparison

cppforlife: > Trigger off of publishing a release, rather than off of pushing tag

we should still keep the tag trigger imho

danielhelfand: Updated release process to not use goreleaser. Example of carrying out release here: https://github.com/danielhelfand/carvel-kapp-controller-fork1/releases/tag/v0.31.0

Logs of release here: https://github.com/danielhelfand/carvel-kapp-controller-fork1/runs/4246855321?check_suite_focus=true

.github/workflows/release-process.yml
```diff
@@ -0,0 +1,102 @@
+name: kapp-controller release
+
+on:
+  push:
+    tags:
+      - 'v*'
```
joe-kimmel-vmw, line 6: If Carvel lives for a million years eventually someone will push a tag like `vincent's-special-tag` and we'll automate a release off of it,  but this probably captures the 99% use-case :)

.github/workflows/release-process.yml
```diff
@@ -0,0 +1,102 @@
...
+
+jobs:
+  kapp-controller-release:
+    name: kapp-controller release
+    runs-on: ubuntu-latest
```
joe-kimmel-vmw, line 11: nit - i'd be tempted to pin to `ubuntu-20.04` just so that nobody has to scramble in two-ish years when ubuntu-latest suddenly points to 22.04 and some weird breaking change happens..

.github/workflows/release-process.yml
```diff
@@ -0,0 +1,102 @@
...
+        uses: actions/checkout@v1
+        with:
+          path: src/github.com/${{ github.repository }}
+
+      - name: Install Carvel Tools
+        uses: vmware-tanzu/carvel-setup-action@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          only: ytt, kbld
+          ytt: v0.37.0
```
joe-kimmel-vmw, line 23: I'd be tempted to generate these versions from a single source of truth, e.g. write a pre-commit hook that scrapes the versions from our Dockerfile and puts them in here,  as otherwise eventually someone will be confused about some behavior.   granted that's some extra machinery - can be left for the future, no need to block initial rollout of this automation.

docs/dev.md
```diff
@@ -94,33 +94,51 @@ Tag the release - it's necessary to do this first because the release process us
...
-kapp deploy -a kc -f ./tmp/release.yml
-```
+After the release process finishes successfully, you can view the newly drafted release via 
+the GitHub UI [here](https://github.com/vmware-tanzu/carvel-kapp-controller/releases). 
 
-After verifying, push the tag to GitHub.
-```
-git push --tags
-```
+The newly published release will be available as a Draft (i.e. not available to the public).
```
joe-kimmel-vmw, line 112: that's so cool!

joe-kimmel-vmw (commented):
few little nits / future-iteration notes but LGTM,  big improvement over current process!

hack/build-release.sh
```diff
@@ -9,6 +9,6 @@ source $(dirname "$0")/version-util.sh
 
 ytt -f config/ -f config-release -v kapp_controller_version="$(get_kappctrl_ver)" | kbld -f- > ./tmp/release.yml
 
-shasum -a 256 ./tmp/release*.yml
+shasum -a 256 ./tmp/release*.yml > ./tmp/checksums.txt
```
cppforlife: can we output and redirect?

danielhelfand: So something like `shasum -a 256 ./tmp/release*.yml | tee ./tmp/checksums.txt`?

danielhelfand: Let me know if you meant something different here, but tested on fork and ran successfully including output of shasum in logs.

.github/workflows/release-process.yml
```diff
@@ -0,0 +1,102 @@
...
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run release script
+        run: |
+          set -e -x
+          ./hack/build-release.sh
+
+      - name: Create release draft and upload release yaml
+        uses: softprops/action-gh-release@v1
```
cppforlife: assuming we have audited this action, lets lock it down by digest.

danielhelfand: Added digest for the action: `softprops/action-gh-release@17cd0d34deddf848fc0e7d9be5202c148c270a0a`

PR #414, opened 2021-10-27
Updated test cases in app directory

### LINKED ISSUE #393: Update tests to use testify/require instead of value comparisons and Fatal

Stuie:

Switching from conditions and `t.Fatalf` to testify assertions.

Fixes #393

No, unit tests only.

### DIFF
--- a/pkg/app/app_test.go
+++ b/pkg/app/app_test.go
@@ -4,6 +4,7 @@
 package app_test
 
 import (
+	"github.com/stretchr/testify/assert"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


Stuie:
The issue is labelled with `Hacktoberfest`, but the site tells me that the repository is not participating. If this PR gets accepted, could you please add the `hacktoberfest-accepted` label?

pkg/app/app_test.go
```diff
@@ -4,6 +4,7 @@
 package app_test
 
 import (
+	"github.com/stretchr/testify/assert"
```
cppforlife: nit: do you mind moving non-std-import to below import section

Stuie: Sorted the imports, sorry about that.

cppforlife:
🥳 thanks for contributing!

PR #420, opened 2021-11-09
Add ability to auto populate adding KUBERNETES_SERVICE_HOST to noProxy

### LINKED ISSUE #391: kapp-controller fails to come up when cluster is configured with proxy
When TMC attached cluster is configured with proxy and kapp-controller-config is created with the http and https proxy URL, kapp-controller fails to communicate with k8s API server.
```
{"level":"error","ts":1631181446.8004937,"logger":"controller-runtime.manager","msg":"Failed to get API Group-Resources","error":"Get \"https://10.96.0.1:443/api?timeout=32s\": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\tgithub.com/go-logr/zapr@v0.2.0/zapr.go:132\nsigs.k8s.io/controller-runtime/pkg/manager.New\n\tsigs.k8s.io/controller-runtime@v0.7.0/pkg/manager/manager.go:317\ngithub.com/vmware-tanzu/carvel-kapp-controller/cmd/controller.Run\n\tgithub.com/vmware-tanzu/carvel-kapp-controller/cmd/controller/run.go:61\nmain.main\n\tcommand-line-arguments/main.go:41\nruntime.main\n\truntime/proc.go:225"}
```

After adding $(KUBERNETES_SERVICE_HOST) which is 10.96.0.1 in the noProxy list of kapp-controller-config, the kapp-controller started working.
We would like the kapp-controller to populate the required noProxy URLs when it is configured with proxy.

As per the discussion with @cppforlife, kapp-controller can add config that enables auto-population of noProxy list.

kapp-controller version: https://github.com/vmware-tanzu/carvel-kapp-controller/releases/tag/v0.23.0

---
Vote on this request

danielhelfand:

Auto populates `KUBERNETES_SERVICE_HOST` by adding it to `noProxy` list of kapp-controller Secret config. 

Fixes #391 

Will add small note to carvel.dev once UX is solidified.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -31,6 +31,9 @@ const (
 	noProxyEnvVar    = "no_proxy"
 
 	skipTLSVerifyKey = "dangerousSkipTLSVerify"
+
+	kubernetesServiceHostEnvVar    = "KUBERNETES_SERVICE_HOST"
+	kubernetesServiceHostShorthand = "KAPPCTRL_KUBERNETES_SERVICE_HOST"
 )
 
 // Config is populated from the cluster's Secret or ConfigMap and sets behavior of kapp-controller.
@@ -104,7 +107,7 @@ func (gc *Config) Apply() error {
 		return fmt.Errorf("Adding trusted certs: %s", err)
 	}
 
-	gc.configureProxies(gc.httpProxy, gc.httpsProxy, gc.noProxy)
+	gc.configureProxies()
 
 	return nil
 }
@@ -149,20 +152,28 @@ func (gc *Config) addTrustedCerts(certChain string) (err error) {
 	return file.Close()
 }
 
-func (gc *Config) configureProxies(httpProxy, httpsProxy, noProxy string) {
-	if httpProxy != "" {
-		os.Setenv(httpProxyEnvVar, httpProxy)
-		os.Setenv(strings.ToUpper(httpProxyEnvVar), httpProxy)
+func (gc *Config) configureProxies() {
+	if gc.httpProxy != "" {
+		os.Setenv(httpProxyEnvVar, gc.httpProxy)
+		os.Setenv(strings.ToUpper(httpProxyEnvVar), gc.httpProxy)
 	}
 
-	if httpsProxy != "" {
-		os.Setenv(httpsProxyEnvVar, httpsProxy)
-		os.Setenv(strings.ToUpper(httpsProxyEnvVar), httpsProxy)
+	if gc.httpsProxy != "" {
+		os.Setenv(httpsProxyEnvVar, gc.httpsProxy)
+		os.Setenv(strings.ToUpper(httpsProxyEnvVar), gc.httpsProxy)
 	}
 
-	if noProxy != "" {
-		os.Setenv(noProxyEnvVar, noProxy)
-		os.Setenv(strings.ToUpper(noProxyEnvVar), noProxy)
+	if gc.noProxy != "" {
+		gc.addKubernetesServiceHostInNoProxy()
+		os.Setenv(noProxyEnvVar, gc.noProxy)
+		os.Setenv(strings.ToUpper(noProxyEnvVar), gc.noProxy)
+	}
+}
+
+func (gc *Config) addKubernetesServiceHostInNoProxy() {
+	if strings.Contains(gc.noProxy, kubernetesServiceHostShorthand) {
+		k8sSvcHost := os.Getenv(kubernetesServiceHostEnvVar)
+		gc.noProxy = strings.Replace(gc.noProxy, kubernetesServiceHostShorthand, k8sSvcHost, 1)
 	}
 }
 
--- a/config-test/secret-config.yml
+++ b/config-test/secret-config.yml
@@ -42,8 +42,8 @@ stringData:
   #! Uncomment and fill in a value to test with https proxy
   #! httpsProxy: ""
 
-  #! must match the value in proxy_test.go and service dns in test/e2e/assests/registry/registry.yml
-  noProxy: github.com,docker.io,registry-svc.registry.svc.cluster.local
+  #! must match the service dns in test/e2e/assests/registry/registry.yml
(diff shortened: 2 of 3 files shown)


pkg/config/config.go
```diff
@@ -161,11 +161,20 @@ func (gc *Config) configureProxies(httpProxy, httpsProxy, noProxy string) {
...
 	if noProxy != "" {
+		noProxy = addKubernetesServiceHostInNoProxy(noProxy)
 		os.Setenv(noProxyEnvVar, noProxy)
 		os.Setenv(strings.ToUpper(noProxyEnvVar), noProxy)
 	}
 }
 
+func addKubernetesServiceHostInNoProxy(noProxy string) string {
+	if strings.Contains(noProxy, "KUBERNETES_SERVICE_HOST") {
+		k8sSvcHost := os.Getenv("KUBERNETES_SERVICE_HOST")
```
cppforlife: do we need to augment this with any other "shorthands", like kubernetes-api.kube-system.svc.local. etc? might we worth asking folks that are raising this issue.

pkg/config/config.go
```diff
@@ -161,11 +161,20 @@ func (gc *Config) configureProxies(httpProxy, httpsProxy, noProxy string) {
 	}
 
 	if noProxy != "" {
+		noProxy = addKubernetesServiceHostInNoProxy(noProxy)
 		os.Setenv(noProxyEnvVar, noProxy)
 		os.Setenv(strings.ToUpper(noProxyEnvVar), noProxy)
 	}
 }
 
+func addKubernetesServiceHostInNoProxy(noProxy string) string {
```
cppforlife: lets stick on `func (gc *Config)`

pkg/config/config.go
danielhelfand: cc @ssamrit-vmw 

Anything else we should consider here via this shorthand approach? Any preferred UX for it?

neil-hickey (commented):
LGTM

joe-kimmel-vmw:
lgtm, fwiw :)

pkg/config/config.go
ssamrit-vmw: Shouldn't we first split the noProxy with comma delimiter and then check for KUBERNETES_SERVICE_HOST.
With current implementation if someone provides "$(KUBERNETES_SERVICE_HOST)" in noProxy then it will end up creating noProxy as "$(IP-address)", right?

>Anything else we should consider here via this shorthand approach? Any preferred UX for it?

In our testing of kapp-controller on cluster configured with proxy using TMC, we found that adding this ip was sufficient for kapp-controller to work.
We have package-deployment-extension deployed on cluster which will set this value in noProxy of kapp config if cluster is configured with proxy configuration using TMC, so we don't expect user to set it while creating proxy configuration in TMC.

neil-hickey: If someone provides "$(KUBERNETES_SERVICE_HOST)" it will be "$(10.x.y.z)" which I think is much clearer error / indicator that someone made a mistake than if they had put "$(KUBERNETES_SERVICE_HOST)" and got "$(KUBERNETES_SERVICE_HOST)" because it doesn't match 1:1 the string KUBERNETES_SERVICE_HOST

danielhelfand: @neil-hickey I think the challenge here is when is when is this information surfaced to the user. The user won't understand how kapp-controller is replacing `KUBERNETES_SERVICE_HOST` and it will never be displayed in the associated controller secret. We should maybe consider having some kind of more direct error returned saying the shorthand provided was not an exact match.

neil-hickey: makes sense. I figured this information would be surfaced in some way. If that's not really the case, and they supply "xyz-KUBERNETES_SERVICE_HOST" or anything in addition to just the string "KUBERNETES_SERVICE_HOST" then an error / warning seems reasonable, given we are manipulating the input in this particular case.

danielhelfand: So this is not perfect in the sense that this will only protect against extra characters, but it will provide some failure information to help a user navigate this scenario. Hopefully noticing the noProxy not being properly configured / being able to notice the typo will provide enough detail in the case of a typo, etc.

pkg/config/config.go
```diff
@@ -161,9 +166,28 @@ func (gc *Config) configureProxies(httpProxy, httpsProxy, noProxy string) {
...
+		if err != nil {
+			return err
+		}
 		os.Setenv(noProxyEnvVar, noProxy)
 		os.Setenv(strings.ToUpper(noProxyEnvVar), noProxy)
 	}
+	return nil
+}
+
+func addKubernetesServiceHostInNoProxy(noProxy string) (string, error) {
```
cppforlife: should be on `func (gc *Config)`

pkg/config/config.go
```diff
@@ -161,9 +166,28 @@ func (gc *Config) configureProxies(httpProxy, httpsProxy, noProxy string) {
...
+	noProxyList := strings.Split(noProxy, ",")
+	for _, noProxyItem := range noProxyList {
```
cppforlife: this loop seems a bit weird -- we call strings.Replace in case we match (but against a first match which may not even be the matched split item), and we can call it multiple times. on top of that it seems that no_proxy is somewhat weird in various formats https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/ so may be it's not even comma separated.

i would recommend abandoning this logic and just sticking with plain strings.Replace like it was before. if we have concerns of "collision" with actual no proxy info, then lets change name to KAPPCTRL_KUBERNETES_SERVICE_HOST. i dont think we should worry about xyz-KUBE... case.

pkg/config/config.go
danielhelfand: Thanks, it looks like `configureProxies` func was not using `gc *Config` in func, so I was thrown off. I updated configureProxies to use data on Config struct in func as well.

pkg/config/config.go
danielhelfand: Reverted back and now using `KAPPCTRL_KUBERNETES_SERVICE_HOST`.

PR #424, opened 2021-11-15
Add trivy scan daily build

### LINKED ISSUE #368: Enable trivy scan as part of nightly build
As described in https://github.com/vmware-tanzu/carvel-kapp/issues/315, it might be helpful for us to consider adding a security scan as part of a nightly build process. This would help us monitor and triage security alerts better for kapp-controller. In particular, conducting a scan of the kapp-controller image could be helpful.

neil-hickey:
- Done as part of https://github.com/vmware-tanzu/carvel-kapp-controller/issues/368
- Add slack notification to triyy scanner

Fixes #368

```release-note
NONE
```

The final step is to add a secret to github that contains the slack bot with write access to the slack channel.

```docs

```

### DIFF
--- a/.github/workflows/trivy-scan.yml
+++ b/.github/workflows/trivy-scan.yml
@@ -34,25 +34,17 @@ jobs:
           git clone --depth 1 https://github.com/aquasecurity/trivy
       - name: Run Trivy Reports
         run: |
-          # kapp-controller binary - sarif
-          trivy rootfs --ignore-unfixed --format template --template "@trivy/contrib/sarif.tpl" \
-            --output trivy-results-binary.sarif --severity "MEDIUM,HIGH,CRITICAL" \
-            "controller"
+          export TRIVY_IGNORE_UNFIXED=true
+          export TRIVY_SEVERITY="MEDIUM,HIGH,CRITICAL"
+          export TRIVY_TEMPLATE="@trivy/contrib/sarif.tpl"
 
-          # kapp-controller docker image - sarif
-          trivy image --ignore-unfixed --format template --template "@trivy/contrib/sarif.tpl" \
-            --output trivy-results-image.sarif --severity "MEDIUM,HIGH,CRITICAL" \
-            "docker.io/carvel/kapp-controller:${{ github.sha }}"
+          # kapp-controller binary - output in sarif and json
+          trivy rootfs --format template --output trivy-results-binary.sarif "controller"
+          trivy rootfs --format json --output trivy-results-binary.json "controller"
 
-          # kapp-controller docker image - json
-          trivy rootfs --ignore-unfixed --format json \
-            --output trivy-results-binary.json --severity "MEDIUM,HIGH,CRITICAL" \
-            "controller"
-
-          # kapp-controller docker image - json
-          trivy image --ignore-unfixed --format json \
-            --output trivy-results-image.json --severity "MEDIUM,HIGH,CRITICAL" \
-            "docker.io/carvel/kapp-controller:${{ github.sha }}"
+          # kapp-controller docker image - output in sarif and json
+          trivy image --format template --output trivy-results-image.sarif "docker.io/carvel/kapp-controller:${{ github.sha }}"
+          trivy image --format json --output trivy-results-image.json "docker.io/carvel/kapp-controller:${{ github.sha }}"
       - name: Upload Trivy scan results to GitHub Security tab
         uses: github/codeql-action/upload-sarif@v1
         with:
--- a/.github/workflows/trivy-scan.yml
+++ b/.github/workflows/trivy-scan.yml
@@ -1,0 +1,99 @@
+name: Trivy CVE Dependency Scanner
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+
+jobs:
+  scan-all:
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v1
+        with:
+          go-version: "1.17"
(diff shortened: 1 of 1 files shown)


neil-hickey:
For line: https://github.com/vmware-tanzu/carvel-kapp-controller/pull/424/files#diff-d2f2b520bb7bfb36f97c6ec62a237288585d8f9d4301d9b31cb0c3dfad98f1c9R74

I think we can re-use the same bot that the kapp-team are using (given it contains the correct access to our slack instance) - I don't have the ability to add this secret / copy the secret from  https://github.com/vmware-tanzu/carvel-kapp which I think will be the simplest move to get this working.

danielhelfand:
`Test_PackageInstalled_FromPackageInstall_DeletionFailureBlocks` is what [failed in CI](https://github.com/vmware-tanzu/carvel-kapp-controller/runs/4217966568?check_suite_focus=true). This test has been flaky, and we should look into it outside of this pr.

.github/workflows/trivy-scan.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Install Carvel Tools
+        uses: vmware-tanzu/carvel-setup-action@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          only: ytt, kapp, kbld, vendir
+          ytt: v0.37.0
+          kapp: v0.42.0
+          kbld: v0.31.0
+          vendir: v0.23.0
+      - name: Build image from Dockerfile
```
danielhelfand: I don't think we need this, right? Looks like all that is happening here is a `docker build` and trivy-related things. So Carvel tools shouldn't be installed.

neil-hickey: Right, I realized I am missing a commit which tests the kapp-controller binary in addition to the docker image. That's why I had it in there. I am going to pull this PR back to draft while I add that in and then re-open it up

.github/workflows/trivy-scan.yml
```diff
@@ -0,0 +1,88 @@
...
+
+          # docker image
+          docker build -t docker.io/carvel/kapp-controller:${{ github.sha }} .
+      - name: Install Trivy
+        run: |
+          brew install aquasecurity/trivy/trivy
+
+          # download the sarif format template
+          git clone --depth 1 https://github.com/aquasecurity/trivy
+      - name: Run Trivy Reports
```
danielhelfand, line 35: Can any of this be captured in the trivy config so there is less scripting?

.github/workflows/trivy-scan.yml
```diff
@@ -0,0 +1,88 @@
...
+          ./hack/build.sh
+
+          # docker image
+          docker build -t docker.io/carvel/kapp-controller:${{ github.sha }} .
+      - name: Install Trivy
+        run: |
+          brew install aquasecurity/trivy/trivy
+
+          # download the sarif format template
+          git clone --depth 1 https://github.com/aquasecurity/trivy
```
danielhelfand, line 34: How is this used? Is there a different way of getting this than a git clone?

.github/workflows/trivy-scan.yml
```diff
@@ -0,0 +1,88 @@
...
+        uses: actions/setup-go@v1
+        with:
+          go-version: "1.17"
+      - name: Build the kapp-controller artifacts
+        run: |
+          ./hack/install-deps.sh
+          ./hack/build.sh
+
+          # docker image
+          docker build -t docker.io/carvel/kapp-controller:${{ github.sha }} .
```
cppforlife, line 28: hmm... it seems that we are building this image from scratch instead of analyzing last release image, which i think would mean that we would always be getting latest updates to images -- so no cves caught.

.github/workflows/trivy-scan.yml
neil-hickey: Hmm, I am not sure! Its a good shout, maybe we can reduce this down with some config file. 👍

.github/workflows/trivy-scan.yml
neil-hickey: So this is used by the following line:

`--format template --template "@trivy/contrib/sarif.tpl"`

The sarif.tpl is a template file that exists in the trivy repository, I couldn't find another way to get it. It's not built into the trivy scanner binary itself, so this seemed simplest

.github/workflows/trivy-scan.yml
neil-hickey: The outcome from our conversation RE this comment:

We will create another daily scan that gets the latest released docker image and scans it.

.github/workflows/trivy-scan.yml
neil-hickey: Trivy doesn't have any sort of config / defaults. However they do have some ENV_VARs you can use. I made a new commit that extracts some of the duplication to variables. I think it makes it a fair bit clear 👍

.github/workflows/trivy-scan.yml
danielhelfand: Should we maybe check this file into the kapp-controller repo and update regularly? Not sure if we want to pull directly from main branch of trivy to get this.

neil-hickey: It felt cheaper to me for trivy folks to look after that template rather than us. Maybe we start with this, and if it ends up breaking a lot or they remove it, we can pull it into the repo? I am hesitant to take on the template ourselves, given a bunch of the carvel tools will be using the same template eventually and we would have to duplicate it into each repo. @danielhelfand

danielhelfand: Sounds good to me.

.github/workflows/trivy-scan.yml
neil-hickey: This ended up being more complicated, getting the scan results into the security tag based on a tag doesn't seem to work for me. So I am going to move this out into another issue. I think having a daily run against HEAD is good enough that I don't want to block on previously released scanning.

PR #438, opened 2021-11-30
dev-deploy script cuts all available corners

joe-kimmel-vmw:
for rapidly iterating on kapp-controller after it has already been
deployed.

Similar to the one for secretgen-controller, and also achieves an order of magnitude speedup in deploy cycles.

Note I changed our main Dockerfile slightly so that all the executable copying happens before copying over kapp controller,  which should result in using cached images slightly more often i think,  since the kapp-controller executable changes almost every time, but the binaries for kapp, git, helm, etc. change less often? 

time / not-quite-benchmarks:   I ran each of these twice (n=2) 
- deploy.sh from develop branch:   ~285 seconds (4.75 minutes)
- deploy.sh from this branch:  ~261 seconds (4.35 minutes)
- dev-deploy.sh:  ~21 seconds  (0.35 minutes)

This differs from the one for secretgen-controller because kc needs some additional executables (helm, kapp, etc.), and I'm choosing to get those by copying them from the most recent local container image tagged as kapp-controller, which means if you've torn down your dev environment you'd have to run `deploy.sh` the first time and then subsequent deploys could be done via this "dev-deploy" script.


Dockerfile
```diff
@@ -80,5 +77,9 @@ COPY --from=0 /usr/local/bin/age .
 # deployers
 COPY --from=0 /usr/local/bin/kapp .
 
+# Name it kapp-controller to identify it easier in process tree
+COPY --from=0 /go/src/github.com/vmware-tanzu/carvel-kapp-controller/controller kapp-controller
+
+
```
cppforlife: remove extra newline

hack/dev-deploy.sh
```diff
@@ -0,0 +1,26 @@
...
+echo "# -- this file is autogenerated by dev-deploy.sh from main Dockerfile and is not version controlled" > Dockerfile.dev
+
+run_image_start=`cat Dockerfile | grep -n "\- run image \-" | cut -d':' -f1`
+kc_latest_image=`docker image ls | grep kapp-controller | sed 's/ \{2,\}/ /g' | cut -d' ' -f3 | head -n 1`
+if [ -z "$kc_latest_image" ] ;
+then
+  echo "Error: unable to find tag for previous image of kapp-controller"
+  echo "For your first deploy please use hack/deploy.sh and then try re-running this script for subsequent deploys."
+  exit 1
+fi
```
cppforlife, line 16: how come we didnt have to do any of that in here: https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/develop/hack/dev-deploy.sh.

joe-kimmel-vmw: the run container for sg only has one `COPY` directive, and it's for the controller.exe : https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/develop/Dockerfile#L39

kapp-controller is more complicated bc it needs those other executables.  There's other ways of solving this but I was aiming for "a dockerfile that can be built from the main Dockerfile's run image via a series of simple substitutions."     

   So a different strategy (which I didn't pursue) could've been to combine the existing `build` and `run` stages into a single "dev-build" stage (e.g. starting from the install script and including all the wgets: https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/Dockerfile#L28 )

danielhelfand: I think it would be more ideal to have it be combined so that a user isn't confused about where to add new dependencies (e.g. AGE) during development. Needing to know to run `./hack/deploy.sh` sounds like it will lead to confusion in future. I don't think this should block the pr as the point of this was to quickly iterate on another issue, but something that should be changed eventually.

danielhelfand (commented):
Can we add some notes in dev.md on using this?

hack/dev-deploy.sh
joe-kimmel-vmw: re 
> where to add new dependencies

Hopefully it's clear that the `./Dockerfile` is still the only thing under version control and the sole source of truth.  

if we add
> ... during development

then I think I understand you're saying,  "adding new binary dependencies exposes a 'leaky abstraction' in the way that script works, which requires me to know how it works and perform a non-obvious series of operations."  Which is true.  
 I think because we're fundamentally generating a "fast but not robust" dockerfile from a "slow methodical production-grade" one, there's going to be a few foot-guns in that process almost regardless of the details of how it's done;  appreciate your support of merging something that's working and then iterating on it as we see signals for where it needs improvement.

PR #441, opened 2021-12-01
Supports Inline PathsFrom during template phase

### LINKED ISSUE #421: implement spec.template[].ytt.inline.pathsFrom[].secretRef.directoryPath
currently it seems that even though api has directoryPath, that functionality is not implemented (directory would need to be explicitly created).

repro:

```yaml
apiVersion: kappctrl.k14s.io/v1alpha1
kind: App
metadata:
  name: simple-app
  namespace: default
spec:
  serviceAccountName: default-ns-sa
  fetch:
  - inline:
      paths:
        file-from-fetch-paths: ""
        # something/blah.txt: ""
      pathsFrom:
        - secretRef:
            name: from-secret
            directoryPath: dir
  template:
  - ytt:
      inline:
        # pathsFrom:
        # - secretRef:
        #     name: foo
        #     # NOT IMPLEMENTED -- results in an error
        #     directoryPath: something
        paths:
          config.yml: |
            #@ load("@ytt:data", "data")
            #@ load("@ytt:yaml", "yaml")
            
            #@ print(data.list())

            kind: ConfigMap
            apiVersion: v1
            metadata:
              name: foo
            data:
              data: #@ yaml.encode(data.list())
  deploy:
  - kapp: {}
---
kind: Secret
apiVersion: v1
metadata:
  name: from-secret
stringData:
  file-from-secret: ""
```

---
Vote on this request

joe-kimmel-vmw:

Fixes #421


pkg/template/ytt.go
```diff
@@ -139,6 +139,11 @@ func (t *Ytt) addInlinePaths(args []string) ([]string, *memdir.TmpDir, error) {
 
 	inline := t.fetchFactory.NewInline(*t.opts.Inline, t.genericOpts.Namespace)
 
+	// TODO / question:should we be creating the path at this point rather than doing the o.stat and conditional create one layer deeper?
+	// TODO / question: does this suffice to get the path cleaned up later? is this important? do we need to do something else to get the path cleaned up later?
```
joe-kimmel-vmw: it's not clear to me whether it matters that we append the path to t.opts.Paths --  on the pod it seems like `/etc/kappctrl-mem-tmp/` gets wiped entirely.

danielhelfand: Based on test failures, assuming the path should not be appended. As @cppforlife mentioned, this gets added here: https://github.com/vmware-tanzu/carvel-kapp-controller/blob/cdb0a53753f0dea8b936e6f63f42e4580144a396/pkg/template/ytt.go#L152

danielhelfand (commented):
Should we add a test for this? This did break over course of its existence so maybe would be worthwhile to add one.

joe-kimmel-vmw:
> Should we add a test for this? This did break over course of its existence so maybe would be worthwhile to add one.

I was thinking of this change as the minimal coverage: 
https://github.com/vmware-tanzu/carvel-kapp-controller/pull/441/files#diff-eb66446d5a40c82f16b85b0973e506c11901331c6dfe0b6bb7b02e920dec594bR19

Because i updated one of the examples,  and we have a step that kapp-applies all the examples (right? i think that's true?), therefore if this breaks again (in a similar very gross way such that kapp apply fails entirely) then we'd be notified. 

So I think it's a fair question of: "should we add more of a test or more thorough coverage" but I want to see if you agree that the one line change linked above adds a tiny bit of coverage.

danielhelfand:
> Because i updated one of the examples, and we have a step that kapp-applies all the examples (right? i think that's true?), therefore if this breaks again (in a similar very gross way such that kapp apply fails entirely) then we'd be notified.

Yes, it would provide coverage.

> So I think it's a fair question of: "should we add more of a test or more thorough coverage" but I want to see if you agree that the one line change linked above adds a tiny bit of coverage.

The problem I see with using this as coverage is the example will still work if removed. I think it's also unclear that the property is added for coverage. I think having a test makes it more clear.

examples/simple-app-git/2.yml
```diff
@@ -16,6 +16,7 @@ spec:
         pathsFrom:
         - secretRef:
             name: simple-app-values
+            directoryPath: myFavoriteFolder
```
danielhelfand: Can we remove this now that there is a test?

test/e2e/kappcontroller/app_secret_configmap_reconcile_test.go
```diff
@@ -210,6 +210,95 @@ data:
...
+	}
+	cleanUp()
+	defer cleanUp()
+
+	logger.Section("deploy", func() {
+		kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, e2e.RunOpts{StdinReader: strings.NewReader(appYaml)})
+	})
+
+	logger.Section("check configmap is populated", func() {
+		retry(t, 10*time.Second, func() error {
```
danielhelfand: Do we need this retry logic? My thought is the kapp deploy should wait for the App to successfully reconcile but did you see the test was not succeeding in some cases?

test/e2e/kappcontroller/app_secret_configmap_reconcile_test.go
```diff
@@ -210,6 +210,95 @@ data:
...
+  serviceAccountName: kappctrl-e2e-ns-sa
+  fetch:
+  - inline:
+      paths:
+        file-from-fetch-paths: ""
+        something/blah.txt: ""
+      pathsFrom:
+        - secretRef:
+            name: from-secret
+            directoryPath: dir
```
danielhelfand: Maybe `fetchDir`?

test/e2e/kappcontroller/app_secret_configmap_reconcile_test.go
```diff
@@ -210,6 +210,95 @@ data:
...
+  template:
+  - ytt:
+      inline:
+        pathsFrom:
+        - secretRef:
+            name: from-secret
+            directoryPath: something
```
danielhelfand: Maybe `templateDir`?

danielhelfand (commented):
Would this file maybe be a better location for this test since the test is about templating? 

https://github.com/vmware-tanzu/carvel-kapp-controller/blob/aaef1dfec812c3172336471f64e6f2f43469354e/test/e2e/kappcontroller/template_test.go

test/e2e/kappcontroller/app_secret_configmap_reconcile_test.go
danielhelfand: I think the reason it's in the other test in this file is because the secret/configmap triggers the reconciliation of the App (i.e. the secret/configmap for the App is updated causing the App to reconcile). In the test you are writing, the App reconciling should I think be enough to know the configmap exists.

test/e2e/kappcontroller/app_secret_configmap_reconcile_test.go
cppforlife: also probably multiple levels.

joe-kimmel-vmw:
> Would this file maybe be a better location for this test since the test is about templating?
> 
> https://github.com/vmware-tanzu/carvel-kapp-controller/blob/aaef1dfec812c3172336471f64e6f2f43469354e/test/e2e/kappcontroller/template_test.go

I moved it over to that file.   I'm not really sure that this test is about templating,  but i'm not sure what else it would be about. I still am in the position of not understanding what in the real world this feature is actually for.  the user story is like, "for some reason i care about the directory name of some transient artifacts in between fetching and deploying my stuff."  and I'm still missing the "so that ..."

danielhelfand:
> I moved it over to that file. I'm not really sure that this test is about templating, but i'm not sure what else it would be about. I still am in the position of not understanding what in the real world this feature is actually for. the user story is like, "for some reason i care about the directory name of some transient artifacts in between fetching and deploying my stuff." and I'm still missing the "so that ..."

@cppforlife Any scenarios for this test that are more similar to how users would make use of the `directoryPath` for templating?

cppforlife:
> @cppforlife Any scenarios for this test that are more similar to how users would make use of the directoryPath for templating?

file layout is important to ytt, so it's not a "transient artifact". ytt for example has particular order for evaluation files. additionally if we have multiple secrets that have same data content, how would one separate them out without directoryPath?

joe-kimmel-vmw:
> file layout is important to ytt, so it's not a "transient artifact".  ytt for example has particular order for evaluation files. 

I think all the files are temporary/intermediate artifacts but the fact that YTT has an ordering is helpful for imaging why someone might want to specify folder names.  @cppforlife do you think we need to expand this test so that it's relying on the ordering of the files? 

> additionally if we have multiple secrets that have same data content, how would one separate them out without directoryPath?

I'm a little lost on this point: 
- why would all the multiple secrets have the same file name?
- why does it matter whether their data is the same?   do you mean if they all had the same metadata, so they had the same name? (they'd have to be in different namespaces though, right?)

test/e2e/kappcontroller/template_test.go
```diff
@@ -164,3 +164,89 @@ from_path: true
...
+  deploy:
+  - kapp: {}
+---
+kind: Secret
+apiVersion: v1
+metadata:
+  name: from-secret
+  namespace: kappctrl-test
+stringData:
+  file-from-secret: "thisIsData"
```
joe-kimmel-vmw, line 224: add a second secret that also has the `file-from-secret` key,   put that secret in a separate directoryPath, then at least the test is more motivated.

joe-kimmel-vmw:
> * why would all the multiple secrets have the same file name?
> * why does it matter whether their data is the same?   do you mean if they all had the same metadata, so they had the same name? (they'd have to be in different namespaces though, right?)

just to reply here in case anyone (including future me) stumbles across it:  the top level keys of the secret are what becomes the filename

test/e2e/kappcontroller/template_test.go
```diff
@@ -164,3 +164,102 @@ from_path: true
...
+	kubectl := e2e.Kubectl{t, env.Namespace, logger}
+	sas := e2e.ServiceAccounts{env.Namespace}
+
+	name := "inline-pathsfrom-directorypath"
+	appYaml := `---
+apiVersion: kappctrl.k14s.io/v1alpha1
+kind: App
+metadata:
+  name: simple-app2
+  namespace: kappctrl-test
```
cppforlife: should remove explicit ref to kappctrl-test. it's configurable so we shouldnt hard code "common" variation. see all other e2e tests.

test/e2e/kappcontroller/template_test.go
```diff
@@ -164,3 +164,102 @@ from_path: true
...
+        - secretRef:
+            name: from-secret
+            directoryPath: template/dir
+        paths:
+          config.yml: |
+            #@ load("@ytt:data", "data")
+            #@ load("@ytt:yaml", "yaml")
+
+            #@ print(data.list())
+
```
cppforlife: lets remove this since it's not being used by the test.

PR #446, opened 2021-12-08
pretty sure you-all implemented git a while ago

joe-kimmel-vmw:
In our CRDs' openApiSchema all the git fetch stages had a "TODO implement" in the description which I think has been stale for quite a while

only for users who read the openApiSchema

```release-note
removes stale "TODO" from spec.fetch.git.description of several CRDs
```

### DIFF
--- a/config/crds.yml
+++ b/config/crds.yml
@@ -153,7 +153,6 @@ spec:
                         items:
                           properties:
                             git:
-                              description: TODO implement git
                               properties:
                                 lfsSkipSmudge:
                                   type: boolean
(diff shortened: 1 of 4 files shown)


danielhelfand (approved):
lgtm

There are actually a number of other TODOs in comments that I think we should actually remove as well. They are related to dockerconfigjson secret format. 

We also very interestingly have jsonnet/kustomize options exposed through the schema which are not actually implemented. My personal preference would be to remove them until implemented in case users are using things like `kubectl explain`. 

These other miscellaneous items I can document in another issue and shouldn't block this.

PR #461, opened 2021-12-21
hack/test-e2e: tee test output also to a file

joe-kimmel-vmw:

tees test output to a file so that we can see it in stdout but also then retrospectively grep through it if we want to.

#### Which issue(s) this PR fixes

I have an issue where maybe 2-3 e2e tests fail and I wish i could track them, so in other words
I want to both see the whole e2e test output on the screen, and also be able to grep it for just the 
"FAILED" tests.  I think I can accomplish all my dreams by tee-ing the test output to a file.

only for users who run the e2e tests.
```release-note

```

```docs

```

### DIFF
--- a/hack/test-e2e.sh
+++ b/hack/test-e2e.sh
@@ -6,10 +6,10 @@ go clean -testcache
 
 # create ns if not exists because the `apply -f -` won't complain on a no-op if the ns already exists.
 kubectl create ns $KAPPCTRL_E2E_NAMESPACE --dry-run=client -o yaml | kubectl apply -f -
-go test ./test/e2e/kappcontroller -timeout 60m -test.v $@
+go test ./test/e2e/kappcontroller -timeout 60m -test.v $@ | tee tmp/e2eoutput.txt
 
 if [ "$KAPPCTRL_E2E_SECRETGEN_CONTROLLER" == "true" ]; then
-  go test ./test/e2e/secretgencontroller -timeout 60m -test.v $@
+  go test ./test/e2e/secretgencontroller -timeout 60m -test.v $@ | tee -a tmp/e2eoutput.txt
 fi
 
 echo E2E SUCCESS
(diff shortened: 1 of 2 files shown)


hack/test-e2e.sh
```diff
@@ -6,10 +6,10 @@ go clean -testcache
 
 # create ns if not exists because the `apply -f -` won't complain on a no-op if the ns already exists.
 kubectl create ns $KAPPCTRL_E2E_NAMESPACE --dry-run=client -o yaml | kubectl apply -f -
-go test ./test/e2e/kappcontroller -timeout 60m -test.v $@
+go test ./test/e2e/kappcontroller -timeout 60m -test.v $@ | tee tmp/e2eoutput.txt
```
cppforlife, line 9: lets rm this file first then

joe-kimmel-vmw: @cppforlife we can,  but i thought it was graceful enough to overwrite it each time.  I don't mind making the change but i'd appreciate understanding why you think that's better?

cppforlife: > graceful enough to overwrite it each time

doesnt tee just keep on appending? if it does indeed override, then your 2nd tee would always blow away first tee.

joe-kimmel-vmw: > doesnt tee just keep on appending

second one has a `-a`[ flag ](https://man7.org/linux/man-pages/man1/tee.1.html). default behavior is overwrite. 

I'll admit i didn't test very thoroughly and i'm just trusting the man pages.

danielhelfand (commented):
Can we add some notes to dev.md on this? Maybe under the e2e testing section?

joe-kimmel-vmw:
> Can we add some notes to dev.md on this? Maybe under the e2e testing section?

oh sorry- just realized i made this change but forgot to push it :-|

PR #458, opened 2021-12-16
rbac: add permissions for flowcontrol prioritylevelconfigurations and…

joe-kimmel-vmw:
Closes #460 

#### What this PR does / why we need it:

turns out when we upgraded k8s libraries something in the libraries really wanted to Watch these flowcontrol resources. I'm not sure what it is or why it needs to, and also we seem to ~work regardless, so there's definitely another angle that could be explored here of whether we could opt-out from that component. 

However, if you don't like seeing these errors on k8s clusters version > 1.20:
```
k8s.io/client-go@v0.22.4/tools/cache/reflector.go:167: Failed to watch *v1beta1.PriorityLevelConfiguration: failed to list *v1beta1.PriorityLevelConfiguration: prioritylevelconfigurations.flowcontrol.apiserver.k8s.io is forbidden: User "system:serviceaccount:kapp-controller:kapp-controller-sa" cannot list resource "prioritylevelconfigurations" in API group "flowcontrol.apiserver.k8s.io" at the cluster scope
E1215 21:54:59.994620      17 reflector.go:138] k8s.io/client-go@v0.22.4/tools/cache/reflector.go:167: Failed to watch *v1beta1.FlowSchema: failed to list *v1beta1.FlowSchema: flowschemas.flowcontrol.apiserver.k8s.io is forbidden: User "system:serviceaccount:kapp-controller:kapp-controller-sa" cannot list resource "flowschemas" in API group "flowcontrol.apiserver.k8s.io" at the cluster scope
```

but you DO like seeing these other errors on clusters version <= 1.19 : 
```
E1216 19:53:15.293473      17 reflector.go:138] k8s.io/client-go@v0.22.4/tools/cache/reflector.go:167: Failed to watch *v1beta1.PriorityLevelConfiguration: failed to list *v1beta1.PriorityLevelConfiguration: the server could not find the requested resource (get prioritylevelconfigurations.flowcontrol.apiserver.k8s.io)
E1216 19:53:36.083622      17 reflector.go:138] k8s.io/client-go@v0.22.4/tools/cache/reflector.go:167: Failed to watch *v1beta1.FlowSchema: failed to list *v1beta1.FlowSchema: the server could not find the requested resource (get flowschemas.flowcontrol.apiserver.k8s.io)
[...]

PR #462, opened 2021-12-21
Support newer and older k8s clusters via deployment config

### LINKED ISSUE #460: apiserver/client-go 0.22.4 produce noisy logs on <=1.19 k8s cluster
**What steps did you take:**

Deploy kapp-controller on a 1.19 cluster:

```
minikube start --kubernetes-version 1.19.0

kapp deploy -a kc -f https://github.com/vmware-tanzu/carvel-kapp-controller/releases/latest/download/release.yml
```

**What happened:**

The following logs are displayed:

```
E1216 19:53:15.293473      17 reflector.go:138] k8s.io/client-go@v0.22.4/tools/cache/reflector.go:167: Failed to watch *v1beta1.PriorityLevelConfiguration: failed to list *v1beta1.PriorityLevelConfiguration: the server could not find the requested resource (get prioritylevelconfigurations.flowcontrol.apiserver.k8s.io)
E1216 19:53:36.083622      17 reflector.go:138] k8s.io/client-go@v0.22.4/tools/cache/reflector.go:167: Failed to watch *v1beta1.FlowSchema: failed to list *v1beta1.FlowSchema: the server could not find the requested resource (get flowschemas.flowcontrol.apiserver.k8s.io)
E1216 19:54:12.208215      17 reflector.go:138] k8s.io/client-go@v0.22.4/tools/cache/reflector.go:167: Failed to watch *v1beta1.PriorityLevelConfiguration: failed to list *v1beta1.PriorityLevelConfiguration: the server could not find the requested resource (get prioritylevelconfigurations.flowcontrol.apiserver.k8s.io)
E1216 19:54:16.746485      17 reflector.go:138] k8s.io/client-go@v0.22.4/tools/cache/reflector.go:167: Failed to watch *v1beta1.FlowSchema: failed to list *v1beta1.FlowSchema: the server could not find the requested resource (get flowschemas.flowcontrol.apiserver.k8s.io)
```

**What did you expect:**

No logs related to v1beta1 flowcontrol resources since these resources don't exist in k8s 1.19. 

This problem seems to come from the k8s/apiserver dependency: https://github.com/vmware-tanzu/carvel-kapp-controller/blob/435d32dd191917c4575023c81ff6a53113687195/vendor/k8s.io/apiserver/pkg/server/options/recommended.go#L138

The apiserver is hardcoding in the use of the v1beta1 flowcontrol resources, meaning it is not compatible with 1.19. Documentation can be found here on the subject: https://kubernetes.io/docs/concepts/cluster-administration/flow-control/#enabling-disabling-api-priority-and-fairness

By disabling the APIPriorityAndFairness feature gate for the kapp-controller apiserver, the logs go away in 1.19. 

We have discussed making this configurable at the kapp-controller deployment level so users can opt out of this feature on 1.19 or at least opt to using v1alpha1 resources.

**Environment:**

- kapp Controller version (execute `kubectl get deployment -n kapp-controller kapp-controller -o yaml` and the annotation is `kbld.k14s.io/images`): develop branch / latest
- Kubernetes version (use `kubectl version`) 1.19.0

---
Vote on this request

joe-kimmel-vmw:

- adds rbac for APIPriorityAndFairness features so that we don't get log noise about not having permissions to watch them.

- Introduces a config deployment variable that is used to opt in/out of the v1.20 "APIPriorityAndFairness" features that we picked up when we upgraded our k8s libraries.   

So this way we'd recommend k8s clusters v1.19 and below set the new variable to false,  and newer versions of k8s can leave it as true (default)

Fixes #460

```release-note
For installing kapp-controller onto clusters with kubernetes version 1.19 and below we strongly recommend setting the config value `enable_APIPriorityAndFairness` to `false` as otherwise kapp-controller will attempt to watch resources that don't exist in your cluster. While kapp-controller will still be in a nominally working state, the excess log lines alone may create issues.
```

replaces https://github.com/vmware-tanzu/carvel-kapp-controller/pull/458

so we totally need to document this, but i think that level of docs lives in a different repo, right? I'll do that later today and link back.
```docs

```


pkg/apiserver/apiserver.go
```diff
@@ -174,6 +172,14 @@ func newServerConfig(aggClient aggregatorclient.Interface, bindPort int) (*gener
 		return nil, fmt.Errorf("error updating api service with generated certs: %v", err)
 	}
 
+	if !enableAPIPriorityAndFairness {
+		// this feature gate was not enabled in k8s <=1.19 as the
+		// APIs it relies on were in alpha.
+		// the apiserver library hardcodes the beta version of the resource
+		// so the best we can do for older k8s clusters is to allow it to be disabled.
+		feature.DefaultMutableFeatureGate.Set("APIPriorityAndFairness=false")
```
danielhelfand:
```go
err := feature.DefaultMutableFeatureGate.Set(fmt.Sprintf("%s=false", features.APIPriorityAndFairness))
if err != nil {
	return nil, err
}
```

pkg/apiserver/apiserver.go
```diff
@@ -74,13 +72,13 @@ type APIServer struct {
 	aggClient aggregatorclient.Interface
 }
 
-func NewAPIServer(clientConfig *rest.Config, coreClient kubernetes.Interface, kcClient kcclient.Interface, globalNamespace string, bindPort int) (*APIServer, error) {
+func NewAPIServer(clientConfig *rest.Config, coreClient kubernetes.Interface, kcClient kcclient.Interface, globalNamespace string, bindPort int, enableAPIPriorityAndFairness bool) (*APIServer, error) {
```
danielhelfand: Something we'll want to consider in the long term is a better place for the controller options (e.g. globalNamespace and enableAPIPriorityAndFairness). Ideally, we would just pass in the whole opts struct versus continuing to add more params to this func. I think we don't do that now cause of the circular import that comes up from importing controller.Options here.

cmd/main.go
```diff
@@ -29,6 +29,7 @@ func main() {
 	flag.BoolVar(&ctrlOpts.EnablePprof, "dangerous-enable-pprof", false, "If set to true, enable pprof on "+controller.PprofListenAddr)
 	flag.DurationVar(&ctrlOpts.APIRequestTimeout, "api-request-timeout", time.Duration(0), "HTTP timeout for Kubernetes API requests")
 	flag.BoolVar(&runController, controllerinit.InternalControllerFlag, false, "[Internal] run the controller code")
+	flag.BoolVar(&ctrlOpts.APIPriorityAndFairness, "enable-APIPriorityAndFairness", true, "gate for using flowcontrol beta APIs, must be false for clusters <= v1.19")
```
danielhelfand: Maybe `Enable/disable APIPriorityAndFairness feature gate for apiserver. Recommended to disable for <= k8s 1.19.`

pkg/apiserver/apiserver.go
joe-kimmel-vmw: yeah there's gotta be some german word for like, "feeling guilty because you added one more parameter to an already long function signature". right?   

maybe "schadenparameterverbreiten" which is roughly "shameful parameter spreading" ? 

anyways that's how i felt adding this param.  Agree it's getting a bit ugly.

joe-kimmel-vmw: hmmm so i tried the classic trick of "turn 3 arguments into one by making a new struct" and i dunno i'll push it up in my next diff but golang really hates looking pretty.

joe-kimmel-vmw:
docs PR: https://github.com/vmware-tanzu/carvel/pull/318

danielhelfand (approved):
lgtm

config/values.yml
```diff
@@ -4,6 +4,8 @@ namespace: kapp-controller
 packaging_global_namespace: kapp-controller-packaging-global
 create_namespace: true
 create_packaging_namespace: true
+#! clusters version 1.19 and below should disable APIPriorityAndFairness
+enable_APIPriorityAndFairness: true
```
cppforlife: is it weird casing for flag/values? should we have more general configuration for gates? eg `feature_gates: ["APIPriorityAndFairness=true"]` or somethign like that?

danielhelfand: It probably would be a better long term decision to make it more general to feature gates. I wouldn't be opposed to changing the format.

joe-kimmel-vmw: @cppforlife given we're using this flags lib currently https://pkg.go.dev/flag which doesn't support lists it seems like we'd have to do some bespoke string scraping,  doesn't feel totally warranted to set up for N=1 feature flag.   I do agree though that we're missing some feature-flag machinery and also what is up with the mixed camel/snake case variable (shouldn't that be a compiler error?).

joe-kimmel-vmw: @cppforlife we dug into a bit further and continue to feel like 
- there's only one flag currently
- to make it general is a non-trivial expansion of this PR (either we're parsing strings or we're passing unchecked user input directly to the k8s api, so we're parsing strings which is exactly what the flags parser is for already, but it doesn't support lists so we're back to using a single scalar flag.)
- the flag is only for the sake of k8s versions that are already out of their support window
- likely by the time we add a second flag, we'll be ready to remove this flag entirely (i.e. next year when we bump k8s lib versions again, we can entirely drop 1.19 support)

PR #463, opened 2021-12-22
exit early for trivy-scan and double-quote vars

aaronshurley:
_Hopefully_ improves the trivy-scan GH action.

None. It's ongoing maintenance.

No.


danielhelfand (approved):
We all say it's good

PR #464, opened 2021-12-22
PackageRepository: supports inline fetch step

joe-kimmel-vmw:
package repositories support inline fetch

- makes some kinds of e2e tests easier
- someone will probably find a good use for this

Fixes #

```release-note
package repositories support inline fetch
```

probably should add some related line in the docs / update the CRD on carvel.dev
separate PR pending.
```docs
package repositories support inline fetch
```

### DIFF
--- a/config/crds.yml
+++ b/config/crds.yml
@@ -1245,6 +1245,32 @@ spec:
                             type: object
                         type: object
                     type: object
+                  inline:
+                    properties:
+                      paths:
+                        additionalProperties:
+                          type: string
+                        type: object
+                      pathsFrom:
+                        items:
+                          properties:
+                            configMapRef:
+                              properties:
+                                directoryPath:
+                                  type: string
+                                name:
+                                  type: string
+                              type: object
+                            secretRef:
+                              properties:
+                                directoryPath:
+                                  type: string
+                                name:
+                                  type: string
+                              type: object
+                          type: object
+                        type: array
+                    type: object
                 type: object
               paused:
                 description: Paused when set to true will ignore all pending changes, once it set back to false, pending changes will be applied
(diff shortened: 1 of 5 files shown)


vendor/k8s.io/gengo/namer/namer.go.orig
```diff
@@ -1,394 +0,0 @@
-/*
```
cppforlife: hmm, how is this file getting deleted, but our CI scripts also dont consider this as "modified" (im assuming that CI is also trying to delete it?)

joe-kimmel-vmw: I've been confused about this file for a few weeks -- it seems like sometimes the `gen*.sh` scripts insist that it must exist, and other times they exist that it must not exist.  probably the second best thing (if we can't actually figure out what's going on) is to add it to .gitignore so that it stops toggling state and showing up as a diff.  I could add it in this PR if you think that's the right way out?

danielhelfand (changes requested):
Can we add a test for this? 

Can we add documentation at carvel.dev?

joe-kimmel-vmw:
> Can we add a test for this?
I think having it in the examples, which are tested, is a good test in this case, since it's a spec feature that re-uses existing machinery, it has an example named "inline fetch" specifically for this feature, and that example will break if we break the feature.  Do you think we need a further test at this time?

> Can we add documentation at carvel.dev?
https://github.com/vmware-tanzu/carvel/pull/326

danielhelfand:
> I think having it in the examples, which are tested, is a good test in this case, since it's a spec feature that re-uses existing machinery, it has an example named "inline fetch" specifically for this feature, and that example will break if we break the feature. Do you think we need a further test at this time?

Yeah, that's fair. I also forgot this is using same `*v1alpha1.AppFetchInline` which is covered elsewhere. 

> vmware-tanzu/carvel#326

Thanks for the docs. Will give a look.

danielhelfand (approved):
lgtm

PR #468, opened 2022-01-07
Disable kapp inspect for App CRs by default

### LINKED ISSUE #467: kapp inspect causes an increase in the number of api calls to cluster in the reconciliation stage
As discussed in [carvel-kapp#394](https://github.com/vmware-tanzu/carvel-kapp/issues/394), `kapp inspect` is one of the reason to cause a spike in the number of api calls made to the cluster when there are a lot of packages installed.

- [x] Disable inspect by default.
- [x] Provide AppCRs the ability to enable kapp inspect if required.

praveenrewar:

- Disable kapp inspect by default. 
- This is required to reduce the number of calls made by kapp in cases of large number of packages

Fixes #
https://github.com/vmware-tanzu/carvel-kapp-controller/issues/467

```release-note
kapp inspect will now be disabled by default for app cr deployments to reduce number of api calls.
```

ToDo
- [x] Add tests

### DIFF
--- a/pkg/app/app_deploy.go
+++ b/pkg/app/app_deploy.go
@@ -98,15 +98,17 @@ func (a *App) inspect() exec.CmdRunResult {
 	for _, dep := range a.app.Spec.Deploy {
 		switch {
 		case dep.Kapp != nil:
-			cancelCh, closeCancelCh := a.newCancelCh()
-			defer closeCancelCh()
+			if dep.Kapp.Inspect != nil {
+				cancelCh, closeCancelCh := a.newCancelCh()
+				defer closeCancelCh()
 
-			kapp, err := a.newKapp(*dep.Kapp, cancelCh)
-			if err != nil {
-				return exec.NewCmdRunResultWithErr(fmt.Errorf("Preparing kapp: %s", err))
+				kapp, err := a.newKapp(*dep.Kapp, cancelCh)
+				if err != nil {
+					return exec.NewCmdRunResultWithErr(fmt.Errorf("Preparing kapp: %s", err))
+				}
+
+				result = kapp.Inspect()
 			}
-
-			result = kapp.Inspect()
 
 		default:
 			result.AttachErrorf("%s", fmt.Errorf("Unsupported way to inspect"))
--- a/pkg/app/app_reconcile.go
+++ b/pkg/app/app_reconcile.go
@@ -198,6 +198,8 @@ func (a *App) reconcileInspect() error {
 			Error:     inspectResult.ErrorStr(),
 			UpdatedAt: metav1.NewTime(time.Now().UTC()),
 		}
+	} else {
+		a.app.Status.Inspect = nil
 	}
 
 	return a.updateStatus("marking inspect completed")
--- a/pkg/app/app_reconcile.go
+++ b/pkg/app/app_reconcile.go
@@ -84,9 +84,9 @@ func (a *App) reconcileDeploy() error {
 
 	// Reconcile inspect regardless of deploy success
 	// but don't inspect if deploy never attempted
-	if a.app.Status.Deploy != nil {
-		_ = a.reconcileInspect()
-	}
+	// if a.app.Status.Deploy != nil {
+	// 	_ = a.reconcileInspect()
+	// }
 
 	return a.updateStatus("marking reconcile completed")
 }
--- a/pkg/app/app_reconcile.go
+++ b/pkg/app/app_reconcile.go
@@ -84,9 +84,9 @@ func (a *App) reconcileDeploy() error {
 
 	// Reconcile inspect regardless of deploy success
 	// but don't inspect if deploy never attempted
-	// if a.app.Status.Deploy != nil {
-	// 	_ = a.reconcileInspect()
-	// }
+	if a.app.Status.Deploy != nil {
+		_ = a.reconcileInspect()
+	}
 
 	return a.updateStatus("marking reconcile completed")
 }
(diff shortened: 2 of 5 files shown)


pkg/app/app_deploy.go
```diff
@@ -97,6 +97,9 @@ func (a *App) inspect() exec.CmdRunResult {
 
 	for _, dep := range a.app.Spec.Deploy {
 		switch {
+		case dep.Kapp != nil && dep.Kapp.Inspect == nil:
```
cppforlife: lets turns this into a conditional under case dep.Kapp case (switch cases are only picking types of deployers eg kapp today)

pkg/app/app_reconcile.go
```diff
@@ -190,12 +190,14 @@ func (a *App) resetLastDeployStartedAt() {
 func (a *App) reconcileInspect() error {
 	inspectResult := a.inspect().WithFriendlyYAMLStrings()
 
-	a.app.Status.Inspect = &v1alpha1.AppStatusInspect{
-		Stdout:    inspectResult.Stdout,
-		Stderr:    inspectResult.Stderr,
-		ExitCode:  inspectResult.ExitCode,
-		Error:     inspectResult.ErrorStr(),
-		UpdatedAt: metav1.NewTime(time.Now().UTC()),
+	if inspectResult != (exec.CmdRunResult{}) {
```
cppforlife: would be nice to add IsEmpty() bool on CmdRunResult

pkg/app/app_deploy.go
praveenrewar: I thought of doing that but wanted to avoid the additional nesting, but it does makes sense to keep the behaviour consistent. Will do.

pkg/app/app_reconcile.go
praveenrewar: I was just looking for an alternative to do the comparison here. Thanks :)

pkg/app/app_deploy.go
praveenrewar: Done.

pkg/app/app_reconcile.go
praveenrewar: Done.

pkg/exec/cmd_run_result.go
```diff
@@ -62,3 +62,11 @@ func (r CmdRunResult) WithFriendlyYAMLStrings() CmdRunResult {
 		Finished: r.Finished,
 	}
 }
+
+// IsEmpty returns true if CmdRunResult is empty, otherwise returns false
```
neil-hickey: nit: Could this be turned into something like the following? And I think the comment is somewhat redundant, the function is so small and just does one thing

```
func (r CmdRunResult) IsEmpty() bool {
  return r == (CmdRunResult{})
}
```

neil-hickey:
Was the intention of this to just disable `kapp inspect` by default? Or also provide the option to enable it on the App CR ? I don't think I see that second part, but if it's coming in another PR that's fine 👍 

@praveenrewar

pkg/exec/cmd_run_result.go
joe-kimmel-vmw: agree with the implementation note;  i think the comment is b/c it's a public method and we have that blanket policy turned on.  i do think we can bump the SHA in the linter config rather than getting trapped in absurd/uninformative comments but imo something like `IsEmpty tells whether the receiver is equal to its zero value` is maybe a tiny bit more helpful?


I think  `empty` is overloaded because it can mean "has no fields"  but also maybe "is equivalent to what you'd get with a `{}` zero-value / default-constructor",  right?

praveenrewar: > nit: Could this be turned into something like the following?

Absolutely. Thanks for the suggestion :)

> And I think the comment is somewhat redundant, the function is so small and just does one thing

Not a big fan of such redundant comments myself, as @joe-kimmel-vmw mentioned I had to do it because of the blanket policy turned on😅 
If y'all agree that the comment is not needed, I would be happy to bump the SHA in linter config.

danielhelfand (commented):
For clarification here, would a user need to pass through an arg to `spec.deploy.kapp.inspect.rawoptions` to get the inspect information in the status?

I think it would be helpful to document this on the carvel.dev kapp-controller website to help users know how to enable this behavior since it is currently shown by default now.

joe-kimmel-vmw:
lgtm,  thank you!  Agree with Daniel re. documentation somehow/somewhere but that'll be a separate repo.

neil-hickey (approved):
LGTM

pkg/app/app_reconcile.go
```diff
@@ -190,12 +190,14 @@ func (a *App) resetLastDeployStartedAt() {
 func (a *App) reconcileInspect() error {
 	inspectResult := a.inspect().WithFriendlyYAMLStrings()
 
-	a.app.Status.Inspect = &v1alpha1.AppStatusInspect{
-		Stdout:    inspectResult.Stdout,
-		Stderr:    inspectResult.Stderr,
-		ExitCode:  inspectResult.ExitCode,
-		Error:     inspectResult.ErrorStr(),
-		UpdatedAt: metav1.NewTime(time.Now().UTC()),
+	if !inspectResult.IsEmpty() {
```
cppforlife, line 193: i would recommend to have an else here as well, that clears inspect status if it inspectresult is empty.

praveenrewar:
> For clarification here, would a user need to pass through an arg to spec.deploy.kapp.inspect.rawoptions to get the inspect information in the status?

Adding raw options would work, alternately having an empty inspect would also work. Something like this:
```yaml
   spec:  
      deploy:
      - kapp:
          inspect: {}
 ```
 
> I think it would be helpful to document this on the carvel.dev kapp-controller website to help users know how to enable this behavior since it is currently shown by default now.

Yes, totally makes sense to have this documented. I will create a PR for it.

pkg/app/app_reconcile.go
praveenrewar: Sure, although I am curious to know if there are any chances of inspect status not being empty before this stage.

cppforlife: it's more about enabling it, then disabling it.

praveenrewar: Done.

PR #471, opened 2022-01-12
set pipefail for e2es

joe-kimmel-vmw:


danielhelfand:
Closing since #472 is in

PR #472, opened 2022-01-12
Fix inspect tests

danielhelfand:

When #468 was introduced, there were test failures that were not surfaced by CI since there was no pipefail set for the test-e2e.sh script. As a result, the pr appeared to be passing all tests even though there were failures.

This fixes all tests in a way to preserve original behavior of the tests, which include inspect in the expected status.

It also adds the pipefail to the script.


joe-kimmel-vmw (approved):
LGTM

Issue #412, opened 2021-10-26
CLI for kapp-controller


renuy:
**Describe the problem/challenge you have**
There is no interface for users to interact with kapp-controller server-side. 

**Describe the solution you'd like**
a CLI which can be used to interact with kapp-controller  

**Milestone 1 -- Parity**
- [x] package available get
- [x] package available list
- [x] package install
  - [x] package installed create (alias to "package install")
  - [x] check values-file existence before trying to do things
- [x] package installed get
- [x] package installed list
- [x] package installed delete
- [x] package installed update
  - [x] check values-file existence before trying to do things
- [x] package repository get
- [x] package repository list
- [x] package repository create
- [x] package repository delete
- [x] package repository update
- [x] Clean up (In Progress)
  - [x] openapi v3 parsing (types)
- [x] Tests (In Progress)
- [x] Update root command name? (kapp -> kctrl???)
- [x] values in/out clean up (single values only supported)
- [x] remove wait for delete of package install
- [x] idempotency of install (already in downstream: https://github.com/vmware-tanzu/tanzu-framework/pull/983)
- [x] PR to develop
- [x] documentation PR listing out commands and workflows (Ongoing and updated [PR](https://github.com/vmware-tanzu/carvel/pull/315))

**Milestone 1 -- Feedback**
- [x] Error out when `--version` is not specified by user while installing package
- [x] Remove `--service-account-name` flag from `package installed update`
- [x] Ensure that if a values secret exists with a key other than `valuesFileKey`, update existing key
    - [x] Blow up if more than one keys already exist
- [x] Update only Fetch field instead of whole Spec when updating a package repository.
- [x] Update reconciliation messages to include resource description
- [x] Remove usage of openapi3 library?
- [x] Remove create namespace flag.
- [x] openapi schema fields should be sorted alpha
- [x] add github action test workflow
- [x] Add examples for commands
- [x] ensure that progress waiting output is not repeated (was prev problem with spinner impl)

**Milestone 2 -- Consuming as library**
- [x] importable command tree
- [x] UI Presentation
- [x] argument passing
- [x] global flags
- [x] Handle legacy Tanzu annotations and carvel annotations better
- [x] compile a list of differences from current tanzu package and kctrl (breaking changes)
- [x] sorting semvers via semver library
- [x] Support stdin input (e.g. -) for values file

**Milestone 3 -- Enhancements**
- [x] get, list, delete commands for app  
- [x] status, kick and pause commands for app 
[...]

PR #480, opened 2022-01-18
Enhancements/additions to kctrl CLI

100mik:
Ongoing PR with progress on #412 

So far:
- Flag for fetching kapp-controller version on the server
- StdIn input while passing a values file

### DIFF
--- a/cli/pkg/kctrl/cmd/package/available/get.go
+++ b/cli/pkg/kctrl/cmd/package/available/get.go
@@ -43,6 +43,7 @@ func NewGetCmd(o *GetOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Command 
 		Use:     "get",
 		Aliases: []string{"g"},
 		Short:   "Get details for an available package or the openAPI schema of a package with a specific version",
+		Args:    cobra.ExactArgs(1),
 		RunE:    func(_ *cobra.Command, args []string) error { return o.Run(args) },
 		Example: cmdcore.Examples{
 			cmdcore.Example{"Get details about an available package",
--- a/cli/pkg/kctrl/cmd/package/available/get.go
+++ b/cli/pkg/kctrl/cmd/package/available/get.go
@@ -30,12 +30,11 @@ type GetOptions struct {
 
 	ValuesSchema bool
 
-	binaryName        string
-	positionalNameArg bool
+	pkgCmdTreeOpts cmdcore.PackageCommandTreeOpts
 }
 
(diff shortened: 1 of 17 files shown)


cli/pkg/kctrl/cmd/version.go
```diff
@@ -26,11 +38,42 @@ func NewVersionCmd(o *VersionOptions, flagsFactory cmdcore.FlagsFactory) *cobra.
...
+
 	return cmd
 }
 
 func (o *VersionOptions) Run() error {
+	if o.controllerVersion {
+		err := o.showControllerVersion()
+		if err != nil {
+			return err
+		}
```
cppforlife: nit: just return above

cli/pkg/kctrl/cmd/version.go
```diff
@@ -26,11 +38,42 @@ func NewVersionCmd(o *VersionOptions, flagsFactory cmdcore.FlagsFactory) *cobra.
...
+func (o *VersionOptions) showControllerVersion() error {
+	coreClient, err := o.depsFactory.CoreClient()
+	if err != nil {
+		return err
+	}
+
+	controllerDeployment, err := coreClient.AppsV1().Deployments(kappControllerNamespace).Get(context.Background(), kappControllerDeployment, metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return fmt.Errorf("kapp-controller not installed on cluster")
```
cppforlife: nit: `kapp-controller not installed on cluster (or is not present in the default location namespace=kapp-controller deployment=kapp-controller)`

cli/pkg/kctrl/cmd/package/available/get.go
```diff
@@ -30,11 +30,12 @@ type GetOptions struct {
 
 	ValuesSchema bool
 
+	binaryName        string
 	positionalNameArg bool
```
cppforlife: should we turn these into a struct (it's used all over) and it gets pretty noisy to pass them around as individual args.

praveenrewar: Yep, I was thinking of doing the same, I was a little hesitant because then we would have opts inside other opts (GetOptions, ListOptions, etc,.), but I think it makes sense to have a struct, maybe we can name it something else.

cli/pkg/kctrl/cmd/version.go
praveenrewar: Done

cli/pkg/kctrl/cmd/version.go
praveenrewar: Done

PR #497, opened 2022-01-26
Fix trivy slack channel

benmoss:
Fixes the Slack channel to which Trivy notifications are sent for the latest release

cc @neil-hickey


neil-hickey (approved):
LGTM

we might want to change this whole action to using https://github.com/vmware-tanzu/carvel-release-scripts/#start-using-the-trivy-scanning-for-cli-tools to be consistent with the other tools now.

benmoss:
The problem with that script is it just scans source code, we're scanning images

PR #499, opened 2022-01-29
Dependency updater

neil-hickey:

This PR introduces a new github action workflow that runs daily around noon. It checks each carvel component kapp-controller uses and checks for a newer version

None

```release-note
N/A
```

I made a few decisions along the way:

1. I moved the dependency version and shas out into a `dependencies.yml` file. This has the consequence of: This file is now required to run the `install-deps.sh` script. I did this because inline editing `install-deps.sh` is too complicated. Too much sed / trying to key off variable names to edit in place. It was gross.
2. ~~`yq` is now required to run the `install-deps.sh` script~~
3. I changed `test-gh.yml` and `upgrade-testing.yml` to use `install-deps.sh` instead of the github action to download the tools. I did this because I wanted less places for this version to be proliferated

There's also a baked in assumption that all tools will release their binary shasums in the release notes. This is the only contract we have today for all the carvel tools, some use a `checksums.txt` and others dont. Fun!

##### Review Checklist:

  a link to that PR
  change


joe-kimmel-vmw:
I'm excited about (3) but I'm a little bummed about (2):
> yq is now required to run the install-deps.sh script

because yq is not something i expect pre-installed on docker base images etc., and the install-deps script is useful for e.g. getting a katacoda environment to work (that was why i wrote it originally).  Or more generally you could say, it feels gross to introduce a "dep" that the install-deps script can't install.   This is different from sha256sum, which is included in at least some base images (again, katacoda being the one i cared about at the time). 

  I see a few options:

- gross sed stuff 
  - could try to make sed-stuff less gross
  - could embrace the grossness
- accepting yq as a requirement 
  - could try write a cross-platform script to pre-install dependencies for the install-deps script. 
  - could embrace the grossness
- other...?

benmoss:
What about if we used Go modules? We could use Dependabot then too. I guess then Go would be a dependency for install-tools. Maybe we need to split the requirements for a kadacoda workflow and a build-the-Dockerfile workflow.

neil-hickey:
> What about if we used Go modules? We could use Dependabot then too. I guess then Go would be a dependency for install-tools. Maybe we need to split the requirements for a kadacoda workflow and a build-the-Dockerfile workflow.

I like this idea, I was thinking something along those lines also. But I don't have familiarity enough yet to know how big of a change that would be. This was just intended to help with some shorter term pain around manually pulling in the other dependencies with the setup we have today.

.github/workflows/dependency-updater.yml
```diff
@@ -0,0 +1,53 @@
...
+          
+          release_notes=$(curl https://api.github.com/repos/vmware-tanzu/carvel-${{ matrix.tool }}/releases/latest | jq .body -r)
+          new_darwin_amd64_sha=$( echo "$release_notes" | grep ${{ matrix.tool }}-darwin-amd64 | awk '{print $1;}')
+          new_linux_amd64_sha=$( echo "$release_notes" | grep ${{ matrix.tool }}-linux-amd64 | awk '{print $1;}')
+                    
+          yq eval '.${{ matrix.tool }}.version = "${{ steps.latest-version.outputs.version }}"' -i ./hack/dependencies.yml
+          yq eval ".${{ matrix.tool }}.checksum_darwin = \"${new_darwin_amd64_sha}\"" -i ./hack/dependencies.yml
+          yq eval ".${{ matrix.tool }}.checksum_linux = \"${new_linux_amd64_sha}\"" -i ./hack/dependencies.yml          
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3.12.0
```
cppforlife: if we are going to be using 3rd party dependency here... (options)
- minimum: review code in their repo + lock down by digest action used
- use git cli directly in the action
- find something officially published by github (still would want to lock down by digest)

neil-hickey:
hey @joe-kimmel-vmw I went ahead and created another PR that uses sed / cut / bash-y things to do the same as this one.

It's fairly brittle because - editing a non-structured `.sh` file will always be hacky

https://github.com/vmware-tanzu/carvel-kapp-controller/pull/505

hack/dependencies.yml
```diff
@@ -0,0 +1,21 @@
+---
+ytt:
+  version: v0.38.0
+  checksum_darwin: 1f84e26457a5efd2470df408072bf2f182c939414e4958f82e19e122d2ee120c
+  checksum_linux: 2ca800c561464e0b252e5ee5cacff6aa53831e65e2fb9a09cf388d764013c40d
+kbld:
+  version: v0.32.0
+  checksum_darwin: 5fc8a491327294717611974c6ab3da2bda3f3809ef3147c1e8472ac62af3ee18
```
joe-kimmel-vmw: if you duplicated the tool name in these leaf node names (e.g. here, `kbld_checksum_darwin`) then you get to have a structured file that you can use with yq outside of the install-deps script, and it's ~straightforward for the install-deps script to use sed to scrape the desired values.    do you think that's worth it?

neil-hickey: I like it :) I think it would work for sure, it's both structured where it needs to be, and not where its not. Shouldn't take long to change it to this. 👍

hack/install-deps.sh
```diff
@@ -19,32 +22,32 @@ install() {
...
-    ytt_checksum=2ca800c561464e0b252e5ee5cacff6aa53831e65e2fb9a09cf388d764013c40d
-    kbld_checksum=de546ac46599e981c20ad74cd2deedf2b0f52458885d00b46b759eddb917351a
-    kapp_checksum=9d2c0441729bebe49e5fb9a1a7b7088f0bee4092ba23e267339a9478948c5731
-    imgpkg_checksum=cfcfcb5afc5e3d28ce1f2f67971a4dcd18f514dadf8a63d70c864e49c9ddca7e
-    vendir_checksum=b7bfd227aa2e6df602f8e79edf725bb0a944b68d207005f42f46f061c4ecd55a
+    ytt_checksum=$(awk '/ytt_checksum_linux/{print $(NF-0)}' "${DEPENDENCIES_DIR}")
+    kbld_checksum=$(awk '/kbld_checksum_linux/{print $(NF-0)}' "${DEPENDENCIES_DIR}")
+    kapp_checksum=$(awk '/kapp_checksum_linux/{print $(NF-0)}' "${DEPENDENCIES_DIR}")
+    imgpkg_checksum=$(awk '/imgpkg_checksum_linux/{print $(NF-0)}' "${DEPENDENCIES_DIR}")
+    vendir_checksum=$(awk '/vendir_checksum_linux/{print $(NF-0)}' "${DEPENDENCIES_DIR}")
```
joe-kimmel-vmw: cool cool --  i check and awk is available on the stock katacoda image, but unsurprisingly doesn't seem like it's in photon.  Before merging can you verify real quick that the hack/deploy scripts still work?

.github/workflows/dependency-updater.yml
```diff
@@ -0,0 +1,58 @@
+name: dependency-updater
+
+on:
+  schedule:
+    - cron: '0 12 * * *'
```
danielhelfand, line 5: What happens if a pr is open with a dependency bump, but it's just not merged in yet?

neil-hickey: Nothing, the PR will be left open and the changes untouched (the step `Create Pull Request` will try to update the branch, see no changes and does nothing to the PR)

danielhelfand: Cool, thanks for clarification!

Issue #407, opened 2021-10-21
Expand automated testing to cover more k8s distributions and versions


joe-kimmel-vmw:
**Describe the problem/challenge you have**
Sometimes people run kapp-controller on something other than minikube. Weird, right?  I thought this whole kubernetes thing was for turning developer laptops into space heaters during the winter months, and then in the summer we shed our winter coats and run applications on bare metal again?   But anyways, whoever these people are running our tool in "the cloud", it turns out sometimes we have compatibility issues.  In order to support all clouds of digital troposphere, from the noble cirrus to the humble cumulonimbus, we should test against them!

**Describe the solution you'd like**
We should probably test against at least AWS and GC? 
There's something vaguely close to prior art in impgkg: https://github.com/vmware-tanzu/carvel-imgpkg/blob/develop/.github/workflows/gh-test-external-registry.yml#L59

**Anything else you would like to add:**
Definitely related to the discussion in https://github.com/vmware-tanzu/carvel-kapp-controller/issues/352 in spirit, - at some point there's a matrix of kubernetes versions x kubernetes providers. 

Likely a good first step on this issue would be to peel off a single deliverable,  like "we'll test kubernetes 1.22 on Google Cloud"; this is intended more as an "epic" or "parent issue". 

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

benmoss:
I always try to push back on these types of requests, there's an element of "let's test our thing on every Intel processor to make sure we're compatible". Kubernetes is not quite the same as x86-64, but to some extent it is.

danielhelfand:
One potential approach here is have a daily CI job that runs tests or examples on various k8s distributions like GKE/OpenShift. This way we have some level of coverage for kapp-controller in these environments but do not have to invest tons of time in maintaining this process. 

Another thing that could be helpful for this issue is to understand more on where support requests are coming from (i.e. which distributions). This would help us better identify what to have covered by this process.

PR #512, opened 2022-02-04
Support running as arbitrary non-root user

benmoss:
Make our Dockerfile and deployment manifest not specify a UID/GID, but instead just `runAsNonRoot` and some other low privilege specifications.

OpenShift in particular will pick a random UID to run as, so we can't be dependent on file permissions specified in the Dockerfile.

Related to #407

```release-note
Container will no longer run as a hard-coded kapp-controller user
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,6 +62,8 @@ RUN tdnf install -y git openssh-clients sed
 RUN echo "kapp-controller:x:1000:" > /etc/group && \
     echo "kapp-controller:x:1000:1000::/home/kapp-controller:/usr/sbin/nologin" > /etc/passwd
 
+RUN chmod a+w /etc/pki/tls/certs/ca-bundle.crt
+
 # fetchers
 COPY --from=0 /helm-v2-unpacked/linux-amd64/helm helmv2
 COPY --from=0 /helm-unpacked/linux-amd64/helm .
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,11 +57,10 @@ RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags="-X 'main.Version=$KC
 # --- run image ---
 FROM photon:4.0
 
-RUN tdnf install -y git openssh-clients shadow-tools sed
+RUN tdnf install -y git openssh-clients sed
 
-RUN groupadd -g 2000 kapp-controller && useradd -r -u 1000 --create-home -g kapp-controller kapp-controller
-RUN chmod g+w /etc/pki/tls/certs/ca-bundle.crt && chgrp kapp-controller /etc/pki/tls/certs/ca-bundle.crt
-USER kapp-controller
+RUN echo "kapp-controller:x:1000:" > /etc/group && \
+    echo "kapp-controller:x:1000:1000::/home/kapp-controller:/usr/sbin/nologin" > /etc/passwd
 
 # fetchers
 COPY --from=0 /helm-v2-unpacked/linux-amd64/helm helmv2
@@ -81,5 +80,6 @@ COPY --from=0 /usr/local/bin/kapp .
 # Name it kapp-controller to identify it easier in process tree
 COPY --from=0 /go/src/github.com/vmware-tanzu/carvel-kapp-controller/controller kapp-controller
 
+USER 1000:1000
 ENV PATH="/:${PATH}"
 ENTRYPOINT ["/kapp-controller"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,12 +57,14 @@ RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags="-X 'main.Version=$KC
 # --- run image ---
 FROM photon:4.0
 
+# Install openssh for git
+# TODO(bmo): why do we need sed?
 RUN tdnf install -y git openssh-clients sed
 
-RUN echo "kapp-controller:x:1000:" > /etc/group && \
-    echo "kapp-controller:x:1000:1000::/home/kapp-controller:/usr/sbin/nologin" > /etc/passwd
-
-RUN chmod a+w /etc/pki/tls/certs/ca-bundle.crt
+# Create the kapp-controller user in the root group, the home directory will be mounted as a volume
+RUN echo "kapp-controller:x:1000:0:/home/kapp-controller:/usr/sbin/nologin" > /etc/passwd
+# Give the root group write access to openssh's root bundle so we can append custom roots at runtime
+RUN chmod g+w /etc/pki/tls/certs/ca-bundle.crt
 
 # fetchers
 COPY --from=0 /helm-v2-unpacked/linux-amd64/helm helmv2
@@ -82,6 +84,7 @@ COPY --from=0 /usr/local/bin/kapp .
 # Name it kapp-controller to identify it easier in process tree
 COPY --from=0 /go/src/github.com/vmware-tanzu/carvel-kapp-controller/controller kapp-controller
 
-USER 1000:1000
+# Run as kapp-controller by default, will be overriden to a random uid on OpenShift
+USER 1000
 ENV PATH="/:${PATH}"
 ENTRYPOINT ["/kapp-controller"]
--- a/config/deployment.yml
+++ b/config/deployment.yml
@@ -51,7 +51,7 @@ spec:
           protocol: TCP
         securityContext:
           allowPrivilegeEscalation: false
-          readOnlyRootFilesystem: true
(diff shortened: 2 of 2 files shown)


config/deployment.yml
```diff
@@ -43,15 +43,17 @@ spec:
...
           mountPath: /etc/kappctrl-mem-tmp
-        securityContext:
-          runAsUser: 1000
-          runAsGroup: 2000
         ports:
         - containerPort: #@ data.values.api_port
           name: api
           protocol: TCP
-      securityContext:
-        fsGroup: 3000
```
cppforlife, line 54: any doc links on how k8s picks fsGroup if not specified?

benmoss: No links but it appears that it'll default to 0

benmoss:
I can't run the tests against an OpenShift environment since the Squid deployment fails, but kapp-controller starts up at least.

config/deployment.yml
benmoss: I think it'll be whatever group the files on volumes already were set to. I think `fsGroup` is applicable when you are mounting a volume with files that are not owned by your container's user, it basically does a `chmod/chown` on all the files/directories in the volume. We don't mount anything but emptyDir volumes so I think it's entirely irrelevant for our use-case.

Dockerfile
```diff
@@ -57,11 +57,12 @@ RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags="-X 'main.Version=$KC
 # --- run image ---
 FROM photon:4.0
 
-RUN tdnf install -y git openssh-clients shadow-tools sed
+RUN tdnf install -y git openssh-clients sed
 
-RUN groupadd -g 2000 kapp-controller && useradd -r -u 1000 --create-home -g kapp-controller kapp-controller
-RUN chmod g+w /etc/pki/tls/certs/ca-bundle.crt && chgrp kapp-controller /etc/pki/tls/certs/ca-bundle.crt
-USER kapp-controller
+RUN echo "kapp-controller:x:1000:" > /etc/group && \
```
benmoss: This gives us a user for normal, non OpenShift use-cases

Dockerfile
```diff
@@ -57,11 +57,12 @@ RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags="-X 'main.Version=$KC
...
+    echo "kapp-controller:x:1000:1000::/home/kapp-controller:/usr/sbin/nologin" > /etc/passwd
+
+RUN chmod a+w /etc/pki/tls/certs/ca-bundle.crt
```
benmoss: This sets the ca-bundle to world-writeable, so that no matter which user we are running as we'll be able to append certificates to it. I think ideally we would use [SSL_CERT_DIR](https://www.jvt.me/posts/2019/12/04/openssl-certs-dir-setup/) with an emptyDir volume, but that requires more work than I wanted to do here.

config/deployment.yml
```diff
@@ -43,19 +43,25 @@ spec:
         volumeMounts:
         - name: template-fs
           mountPath: /etc/kappctrl-mem-tmp
-        securityContext:
-          runAsUser: 1000
```
benmoss, line 47: This will default to the 1000:1000 set in the Dockerfile or be overridden on OpenShift

config/deployment.yml
```diff
@@ -43,19 +43,25 @@ spec:
...
-          runAsGroup: 2000
+        - name: home
+          mountPath: /home/kapp-controller
```
benmoss, line 47: Make this a volume so it is created with writeable permissions for whatever user we're running as

config/deployment.yml
```diff
@@ -43,19 +43,25 @@ spec:
...
         ports:
         - containerPort: #@ data.values.api_port
           name: api
           protocol: TCP
-      securityContext:
-        fsGroup: 3000
+        securityContext:
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: false #! TODO(bmo): don't write to the root ca-bundle
```
benmoss, line 54: Ideally this could be `true` except for the `/etc/pki/tls/certs/ca-bundle.crt` file I mentioned we're writing to.

neil-hickey:
hey @benmoss I noticed there's some assumptions of k8s behaviour RE defaults. Is it fair to assume that these are universal across all the k8s versions? Should we test this across a few versions just to validate?

Dockerfile
neil-hickey: How do we feel about commenting the dockerfile with these github comments? It feels like there's some magic going on that might be worth documenting for future us

joe-kimmel-vmw: agree --  from this diff what i'm guessing is that shadow-tools was providing groupadd and useradd but now we're using our 13117 h4x0rz skills to edit the corresponding files directly,  but i definitely don't know enough *nix sysadmin arcana to understand the effects of appending these lines to these files without the diff showing me what's being replaced.

Dockerfile
neil-hickey: Is there any sort of practical consequence of making the ca-bundle writable to all users?

joe-kimmel-vmw: is there a security implication here?   I guess now anyone with access to the container can tell us to trust some arbitrary source,  whereas before they would've needed access to the container and access to the kapp-controller group?  (which they could probably get pretty easily, so we haven't expanded the surface area of any potential attack in any interesting way?)

config/deployment.yml
joe-kimmel-vmw: i just want to keep saying a bunch of these comments in the PR could be comments in the Dockerfile.

config/deployment.yml
joe-kimmel-vmw: i just want to keep saying a bunch of these comments in the PR could be comments in the Dockerfile.

joe-kimmel-vmw (approved):
lgtm but maybe sprinkle some more comments throughout for additional context

Dockerfile
benmoss: Yeah, this is what I'm most uncertain about in this change. I couldn't think of a way to scope the permissions to a specific user/group.

benmoss:
> hey @benmoss I noticed there's some assumptions of k8s behaviour RE defaults. Is it fair to assume that these are universal across all the k8s versions? Should we test this across a few versions just to validate?

Not sure what RE stands for. What assumptions do you mean?

Dockerfile
```diff
@@ -57,11 +57,14 @@ RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags="-X 'main.Version=$KC
 # --- run image ---
 FROM photon:4.0
 
-RUN tdnf install -y git openssh-clients shadow-tools sed
+# Install openssh for git
+# TODO(bmo): why do we need sed?
```
cppforlife, line 61: why dont we remove it?

config/deployment.yml
```diff
@@ -43,19 +43,25 @@ spec:
         volumeMounts:
         - name: template-fs
           mountPath: /etc/kappctrl-mem-tmp
-        securityContext:
-          runAsUser: 1000
-          runAsGroup: 2000
+        - name: home
```
cppforlife, line 46: do we use home directory for anything?

config/deployment.yml
```diff
@@ -43,19 +43,25 @@ spec:
...
+            drop:
+            - all
       volumes:
       - name: template-fs
         emptyDir:
-          #! https://kubernetes.io/docs/concepts/storage/volumes/#emptydir
+          medium: Memory
+      - name: home
+        emptyDir:
           medium: Memory
```
cppforlife, line 65: dont think we need medium=Memory if we keep home dir.

Dockerfile
benmoss: Saving it for another time, I'm not sure if we need it

config/deployment.yml
benmoss: https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/pkg/apiserver/apiserver.go#L38-L39

config/deployment.yml
benmoss: Not sure what you mean by "if we keep home dir"

config/deployment.yml
cppforlife: can we just use a tmp directory for that?

config/deployment.yml
cppforlife: (per above comment when im trying to figure out why we have home dir)

config/deployment.yml
benmoss: Yeah, though it basically is a temp directory as it stands now, just mounted at `/home/kapp-controller`. I think we want to avoid writing to the rootfs and so we'll need a scratch volume to write these things to (unless they can just be held in-memory).

PR #527, opened 2022-02-18
Make all the tools a little more quiet 🤫 

benmoss:
- No more wget progress bars, use `curl --silent`
- No more go test -v


joe-kimmel-vmw (approved):
I've never read the rise and fall of the roman empire but don't many historians view Constantine's removal of ascii progress bars and verbose test output in 312 CE as one of primary causal factors in Rome's decline?  I'm approving this PR but let's carefully watch for butterfly effects in case the absence of all that test outputs start to erode the very fabric of our current society.

PR #529, opened 2022-02-24
Add basic cue support

### LINKED ISSUE #514: cue as a templating option
We've been hearing some excitement about [cue](https://cuelang.org/). It'd be a great opportunity to increase kapp-controller's extensibility by enabling cue as a templating option.

cue works with kapp-controller.

This is not a prescriptive issue. Feel free to work out the details in Slack.

---
Vote on this request

benmoss:
Adds basic support for cue as a templating option.

Not sure if we want to close #514 since we may want to introduce more functionality.

```release-note
Cue is supported as a templating language
```

What else do we want out of this? Package/PackageInstall samples?

##### Review Checklist:

  a link to that PR
  change

Will need to update docs to add this to the


pkg/template/cue.go
```diff
@@ -0,0 +1,59 @@
...
+// and boolean indicating whether subsequent operations
+// should operate on result, or continue operating on the directory
+func (c *cue) TemplateDir(dirPath string) (exec.CmdRunResult, bool) {
+	return c.template(dirPath, nil), true
+}
+
+// TemplateStream works on a stream returning templating result.
+// dirPath is provided for context from which to reference additonal inputs.
+func (c *cue) TemplateStream(stream io.Reader, dirPath string) exec.CmdRunResult {
+	panic("not implemented") // TODO: Implement
```
cppforlife: instead of panic:

```
return exec.NewCmdRunResultWithErr(fmt.Errorf("Templating stream is not supported"))
```

pkg/apis/kappctrl/v1alpha1/types_template.go
```diff
@@ -99,3 +100,9 @@ type AppTemplateSopsAge struct {
 type AppTemplateSopsPrivateKeysSecretRef struct {
 	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
 }
+
+// +k8s:openapi-gen=true
+type AppTemplateCue struct {
```
cppforlife, line 105: any other exciting options we should be offering based on what cue support?

relatedly, (we can do this in separate PR), we need to figure out how to inject values into cue (similar to --data-values-fie in ytt). how can we pass through PackageInstall's values to cue (values are represented as yaml there)? does cue project have an opinion on this?

benmoss: > any other exciting options we should be offering based on what cue support?

not that I know of, we could expose `--expression` to allow users to write cue expressions to transform their exported value, but that seemed like extra complexity without much justification right now.

>  how can we pass through PackageInstall's values to cue

I think we can have cue merge in yaml into the exported value, it works on the command-line if you explicitly specify the files like `cue export foo.yml foo.cue`, I don't see any way to automatically pickup files. The CLI is pretty confusing in its flags, it might make sense to just use the SDK to have more control.

pkg/template/cue.go
```diff
@@ -0,0 +1,85 @@
...
+// dirPath is provided for context from which to reference additonal inputs.
+func (c *cue) TemplateStream(stream io.Reader, dirPath string) exec.CmdRunResult {
+	return exec.NewCmdRunResultWithErr(fmt.Errorf("Templating stream is not supported")) // TODO: Implement
+}
+
+func (c *cue) template(dirPath string, input io.Reader) exec.CmdRunResult {
+	var stdoutBs, stderrBs bytes.Buffer
+	args := []string{"export", "--out", "yaml"}
+	if len(c.opts.Paths) == 0 {
+		paths, err := filepath.Glob(filepath.Join(dirPath, "*.cue"))
```
cppforlife, line 48: can we not give cue entire directory?

pkg/template/cue.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+		args = append(args, paths...)
+	} else {
+		args = append(args, c.opts.Paths...)
+	}
+
+	vals := Values{c.opts.ValuesFrom, c.genericOpts, c.coreClient}
+	paths, valuesCleanUpFunc, err := vals.AsPaths(dirPath)
+	if err != nil {
+		return exec.NewCmdRunResultWithErr(fmt.Errorf("writing values: %w", err))
```
cppforlife: nit: we typically start all errors with a capital letter -- this is useful for splitting error messages for better presentation.

pkg/template/cue.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	defer valuesCleanUpFunc()
+	if c.opts.InputField != "" {
+		args = append(args, "--path", fmt.Sprintf("%s:", c.opts.InputField))
```
cppforlife: if somebody does elect to use nested keys, would they say `inputField: "foo: bar"` or does --path only support one level?

pkg/template/cue.go
```diff
@@ -0,0 +1,85 @@
...
+	var stdoutBs, stderrBs bytes.Buffer
+	args := []string{"export", "--out", "yaml"}
+	if len(c.opts.Paths) == 0 {
+		paths, err := filepath.Glob(filepath.Join(dirPath, "*.cue"))
+		if err != nil {
+			return exec.NewCmdRunResultWithErr(fmt.Errorf("reading files: %w", err))
+		}
+		args = append(args, paths...)
+	} else {
+		args = append(args, c.opts.Paths...)
```
cppforlife, line 60: should use ScopedPath to limit paths to "working directory" (e.g. https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/pkg/template/ytt.go#L104-L118)

pkg/template/cue.go
benmoss: yeah, i was wondering about this because I get tons of warnings about this in my editor from this codebase: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/template/cue.go
benmoss: Yeah, `foo: bar` would work, a bit of a leaky implementation detail in that way

pkg/template/cue.go
benmoss: it appears not 
```
> cue eval cue/
cannot find package "cue"
```
but doing `cue eval cue/*.cue` works

pkg/template/cue.go
cppforlife: yeah, in this case carvel definitely deviated from that guideline. consistency is better though.

pkg/template/cue.go
cppforlife: if so, then i would recommend not adding : by default and just letting user enter it. not sure if error message is readable when : is not a suffix (or does it just work?)

benmoss:
```
❯ cue eval foo.cue --path 'config' values.yml
error evaluating label config: reference "config" not found
```

not the most helpful, but not the worst either... I'm ok with dropping the `:` and then we can just call it `InputExpression`

cppforlife: > I'm ok with dropping the : and then we can just call it InputExpression

👍 im for it.

PR #537, opened 2022-03-02
Dump app yaml when test-examples.sh fails

benmoss:
Dumps out all apps as YAML when a test app from `test-examples.sh` fails.

When this test [fails](https://github.com/vmware-tanzu/carvel-kapp-controller/runs/5395208957?check_suite_focus=true) it's pretty hard to diagnose what could be going wrong. 

```release-note
NONE
```

### DIFF
--- a/hack/test-examples.sh
+++ b/hack/test-examples.sh
@@ -2,6 +2,11 @@
 
 set -e -x -u
 
+function dump {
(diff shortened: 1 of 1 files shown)


joe-kimmel-vmw:
LGTM;  the most I've ever looked forward to getting dumped

Issue #151, opened 2021-04-01
Support Naming Directories Where Fetch Contents will be Stored


danielhelfand:
**Describe the problem/challenge you have**

As documented in this [code comment](https://github.com/vmware-tanzu/carvel-kapp-controller/blob/858ec7e7713c5185bf571cfd5af720a47c8f035c/pkg/app/app_fetch.go#L27), fetched contents are stored in subdirectories since vendir doesn't allow placing contents in the vendir root. 

The current approach is to use a number starting at 0 and increment for each fetch source included. This isn't documented anywhere for users, so it can be difficult to determine where fetched contents are being stored. 

**Describe the solution you'd like**

To supplement the default behavior of creating/naming subdirectories (i.e. incrementing from 0 to n-1 fetch sources), I would like to be able to specify what the directory name for each fetch source should be. 

```yaml
fetch:
  - git:
      url: https://github.com/k14s/kapp
      ref: origin/develop
      subPath: examples/gitops/guestbook
      subDirName: myDirNameGit
  - http:
      url: https://raw.githubusercontent.com/k14s/kapp/db2cc63e12e988235eb8815af8edc0ca0cfaa79c/examples/simple-app-example/config-1.yml
      sha256: 71b0a2dceb6bb6e7a519d0587abd9400a265cabbe2c084f783df94a92db6d980
      subDirName: myDirNameHttp
template:
  - ytt:
        paths:
          - myDirNameGit/foo.yml
          - myDirNameHttp/bar.yml
```

By default though, kapp-controller could keep its current approach of creating/naming sub directories for those not concerned with where sources are stored.

benmoss:
@neil-hickey is there anything more to do here? didn't #544 fix this?

neil-hickey:
Fixed by #544 - released in 0.33.1

PR #544, opened 2022-03-03
Add new optional field to fetch steps to allow specifying the local dir when fetching

neil-hickey:

Adds a new property to the `fetch` stage of `AppCRs` *only* (not PKGR as they appear to only have one fetchable item allowed) to support user specifying the destination directory for their fetched artifacts (not using our default incrementing integer scheme)

Fixes # https://github.com/vmware-tanzu/carvel-kapp-controller/issues/151

```release-note

```

##### Review Checklist:

  a link to that PR - https://github.com/vmware-tanzu/carvel/pull/387
  change

```docs

```


pkg/apis/kappctrl/v1alpha1/types_fetch.go
```diff
@@ -103,6 +112,9 @@ type AppFetchGit struct {
 	// Skip lfs download (optional)
 	// +optional
 	LFSSkipSmudge bool `json:"lfsSkipSmudge,omitempty" protobuf:"varint,5,opt,name=lfsSkipSmudge"`
+	// Specify the destination directory name for each fetch source
+	// +optional
+	DestDirName string `json:"destDirName,omitempty" protobuf:"bytes,7,opt,name=destDirName"`
```
joe-kimmel-vmw: we should probably have a brief RFC for names for this property that includes some of our more active community members - the biggest constraint was basically that for _some_ fetch types there's already a `subPath` which refers to a path in the source (git repo or http server), so it seemed best to at least attempt to indicate that this is a path that'll exist during templating, not a path that's part of the source.

test/e2e/kappcontroller/fetch_test.go
```diff
@@ -0,0 +1,168 @@
...
+      name: redis
+      # Chart version v1, DEPRECATED 
+      version: "11.3.4"
+      repository:
+        url: https://charts.bitnami.com/bitnami
+      destDirName: helm-dest
+  template:
+  - ytt:
+      paths:
+      - foo
```
joe-kimmel-vmw: i'm trying to imagine having a meaningful test of each of the fetches above.

i guess what if we had a series of ytt templating steps, each of which basically only included one of the named paths and then just put the contents (or part of the contents) of a specific named file from the fetched-contents into the body of a configmap?   that's only a little bit tedious to set up?

joe-kimmel-vmw (commented):
looks like you finished up where we left off - neat - thanks! 

I recommend we discuss the field name at least internally before merging as it'll be kind of fixed forever once we land it.

test/e2e/kappcontroller/fetch_test.go
neil-hickey: That would the fetched contents look like for something like helmTemplate? I did the check for the path being used by vendir for those cases where we don't really have a thing for ytt to read. Or at least not a common thing, I guess we could hardcode the contents of each test artifact?

Like if we know git repo has a file foo.yml in it we use that?

test/e2e/kappcontroller/fetch_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+	out := kapp.Run([]string{"inspect", "-a", name, "--raw", "--tty=false", "--filter-kind=App"})
+
+	var cr v1alpha1.App
+	err := yaml.Unmarshal([]byte(out), &cr)
+	if err != nil {
+		t.Fatalf("Failed to unmarshal: %s", err)
+	}
+
+	if !strings.Contains(cr.Status.Fetch.Stdout, "path: foo") {
```
joe-kimmel-vmw: nit  - let's use stretchr/testify and to `assert.Contains(t, cr.status.Fetch.Stdout, "path: foo", "Expected destDirName to be used by vendir")`

test/e2e/kappcontroller/fetch_test.go
joe-kimmel-vmw: > Like if we know git repo has a file foo.yml in it we use that?

That's what I was thinking.. ?   doesn't need to be that way.  You're right that those vendir assertions are a sort of minimal coverage,  maybe that's enough then.

neil-hickey: I added the tests, turns out using 
```
  template:
  - ytt:
      paths:
      - foo
```
tells ytt to evaluate the whole directory, so adding the configmap was kind of redundant. As ytt will fail if that directory doesn't exist.

neil-hickey:
@benmoss  / @cppforlife opinions on our new field name ? As @joe-kimmel-vmw mentioned, we wanna make sure it's sensible before we have to support it forever

benmoss:
> not PKGR as they appear to only have one fetchable item allowed

This will be on PackageRepositories too, I don't think that's a problem. 

> @benmoss / @cppforlife opinions on our new field name ? As @joe-kimmel-vmw mentioned, we wanna make sure it's sensible before we have to support it forever

I'd vote for `path` for being succinct and simple, it's what [actions/checkout](https://github.com/actions/checkout) uses, with the doc reading `# Relative path under $GITHUB_WORKSPACE to place the repository`

cppforlife:
i dont think we should put this field within individual fetcher (e.g. git). it should probably be at the higher level since it applies to any of them. im ok with `path` -- do we allow nesting in it?.

pkg/app/app_fetch.go
```diff
@@ -71,7 +71,7 @@ func (a *App) fetch(dstPath string) (string, exec.CmdRunResult) {
 	}
 
 	// if only one fetch, update dstPath for backwards compatibility
-	if len(a.app.Spec.Fetch) == 1 {
+	if len(a.app.Spec.Fetch) == 1 && a.app.Spec.Fetch[0].Path == "" {
```
neil-hickey, line 74: I verified that there must always be a len of 1 for the fetch step, therefore this `[0].Path` check should never fail

test/e2e/kappcontroller/fetch_test.go
```diff
@@ -0,0 +1,155 @@
...
+    kapp.k14s.io/change-group: kappctrl-e2e.k14s.io/apps
+spec:
+  serviceAccountName: kappctrl-e2e-ns-sa
+  fetch:
+  - inline:
+      paths:
+        file.csv: |
+            Year,Make,Model,Description,Price
+            1997,Ford,E350,"ac, abs, moon",3000.00
+    path: foo
```
cppforlife: should we expand this to also have a fetcher that has an empty path?

test/e2e/kappcontroller/fetch_test.go
```diff
@@ -0,0 +1,155 @@
...
+  template:
+  - ytt:
+      paths:
+      - foo
```
cppforlife: instead of using paths=[foo], lets not specify any paths explicitly and use ytt's data.list() to list out paths seen by ytt. then you can have an assertion that expected list is `[foo/file.csv, ...]`

test/e2e/kappcontroller/fetch_test.go
```diff
@@ -0,0 +1,155 @@
...
+	var cr v1alpha1.App
+	err := yaml.Unmarshal([]byte(out), &cr)
+	if err != nil {
+		t.Fatalf("Failed to unmarshal: %s", err)
+	}
+
+	assert.Contains(t, cr.Status.Fetch.Stdout, "path: foo", "Expected destDirName to be used by vendir")
+}
+
+func Test_App_FetchDirname_All(t *testing.T) {
```
cppforlife: not a fan of this test since it brings all these external dependencies, but your call if you want to keep it. since path is now "at the top level" risk of breaking it for some and not all cases seems very small.

neil-hickey: removed most of this, I kept two of these for the multiple case instead of all of them - as you mentioned theres no need now that it is a level above

PR #545, opened 2022-03-03
use checkedPath explicitly in cue

cppforlife:


cppforlife:
doesnt work in the following example:

```
root@w-7fcf8:/tmp/build/ary23/cue-test# cue export --out yaml ./foo ./another
bar: bar
foo: foo
---
another: 123

root@w-7fcf8:/tmp/build/ary23/cue-test# cue export --out yaml $(pwd)/foo ./another
cannot find package "/tmp/build/ary23/cue-test/foo"
```

PR #546, opened 2022-03-03
app status test assertions

joe-kimmel-vmw:
I think this test just happens to run first (`go test` has an undefined order, technically) but because the first thing that happens is a `kapp deploy` with an allowError=true, it means that you can get some funny errors if there's stale testing apps deployed in your test namespace. This is especially obscured if you happen to be making a change to the App Status object because then you'll really be wondering what you broke.  So these assertions will hopefully help someone recognize the root cause of the errors they're seeing faster than I did today.

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/test/e2e/kappcontroller/app_status_test.go
+++ b/test/e2e/kappcontroller/app_status_test.go
@@ -62,9 +62,9 @@ spec:
 		out, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, e2e.RunOpts{StdinReader: strings.NewReader(appYaml), AllowError: true})
 		// it's supposed to error, but it's also supposed to create the three supporting resources successfully:
 		assert.Error(t, err)
-		assert.Contains(t, out, "ok: reconcile role/kappctrl-e2e-ns-role (rbac.authorization.k8s.io/v1) namespace: kappctrl-test")
-		assert.Contains(t, out, "ok: reconcile rolebinding/kappctrl-e2e-ns-role-binding (rbac.authorization.k8s.io/v1) namespace: kappctrl-test")
-		assert.Contains(t, out, "ok: reconcile serviceaccount/kappctrl-e2e-ns-sa (v1) namespace: kappctrl-test")
+		assert.Contains(t, out, "ok: reconcile role/kappctrl-e2e-ns-role")
+		assert.Contains(t, out, "ok: reconcile rolebinding/kappctrl-e2e-ns-role-binding")
+		assert.Contains(t, out, "ok: reconcile serviceaccount/kappctrl-e2e-ns-sa")
 
 	})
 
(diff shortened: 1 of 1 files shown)


test/e2e/kappcontroller/app_status_test.go
```diff
@@ -58,10 +59,17 @@ spec:
...
 
 	logger.Section("deploy", func() {
-		kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, e2e.RunOpts{StdinReader: strings.NewReader(appYaml), AllowError: true})
+		out, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, e2e.RunOpts{StdinReader: strings.NewReader(appYaml), AllowError: true})
+		// it's supposed to error, but it's also supposed to create the three supporting resources successfully:
+		assert.Error(t, err)
+		assert.Contains(t, out, "ok: reconcile role/kappctrl-e2e-ns-role (rbac.authorization.k8s.io/v1) namespace: kappctrl-test")
+		assert.Contains(t, out, "ok: reconcile rolebinding/kappctrl-e2e-ns-role-binding (rbac.authorization.k8s.io/v1) namespace: kappctrl-test")
+		assert.Contains(t, out, "ok: reconcile serviceaccount/kappctrl-e2e-ns-sa (v1) namespace: kappctrl-test")
+
```
cppforlife: namespace cant be hard coded since it comes from kapp class. given below assertion im not sure it's worth adding this. alternatively you can may be assert on error message from below inside expectedStatus.

joe-kimmel-vmw: i'm just doing substring assertions so it's appealing to amend by shortening the string to exclude the namespace.   I'd still like to capture the information that those three resources should be successfully created either by asserting here or below in the output of the inspect.

test/e2e/kappcontroller/app_status_test.go
```diff
@@ -58,10 +59,17 @@ spec:
...
 
 	logger.Section("deploy", func() {
-		kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, e2e.RunOpts{StdinReader: strings.NewReader(appYaml), AllowError: true})
+		out, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, e2e.RunOpts{StdinReader: strings.NewReader(appYaml), AllowError: true})
+		// it's supposed to error, but it's also supposed to create the three supporting resources successfully:
+		assert.Error(t, err)
+		assert.Contains(t, out, "ok: reconcile role/kappctrl-e2e-ns-role")
+		assert.Contains(t, out, "ok: reconcile rolebinding/kappctrl-e2e-ns-role-binding")
+		assert.Contains(t, out, "ok: reconcile serviceaccount/kappctrl-e2e-ns-sa")
+
```
cppforlife: nit: newline

test/e2e/kappcontroller/app_status_test.go
cppforlife: not pretty looking, but ok.

benmoss (approved):

PR #551, opened 2022-03-05
fetch.vendir.shouldSkipTLSVerify passes the port

### LINKED ISSUE #549: extractImageRefHostname output doesn't container port number
**What steps did you take:**
Set `dangerousSkipTLSVerify=air-gap-scale.eng.vmware.com:8043` in kapp-controller's config map 

The requests to `air-gap-scale.eng.vmware.com:8043/registry` didn't skip TLS verify

**What happened:**
The [extractImageRefHostname function](https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/pkg/fetch/vendir.go#L359-L368) used to extract dangerousSkipTLSVerify hostname has a bug that it doesn't contain port name in the output. Even it used `hostnameAndPort` as var name.

As a result, requests to pull from `air-gap-scale.eng.vmware.com:8043/registry/xxxxx` will not skip tls even `dangerousSkipTLSVerify=air-gap-scale.eng.vmware.com:8043` is set, as `air-gap-scale.eng.vmware.com` doesn't match with `air-gap-scale.eng.vmware.com:8043`.

**What did you expect:**
The skiptls should happen as expected.

test snippet for kapp controller host name resolve
```
package main

import (
	"fmt"
	"strings"

	"github.com/google/go-containerregistry/pkg/name"
)

func main() {
	ref := "air-gap-scale.eng.vmware.com:8043/registry/packages/core/repo:v1.22.5_vmware.1-tkg.1"
	parsedRef, err := name.ParseReference(ref)
	if err != nil {
		return
	}
	hostnameAndPort := parsedRef.Context().RegistryStr()
	fmt.Println(strings.Split(hostnameAndPort, ":")[0])
}
```
result:
```
$ go run main.go
air-gap-scale.eng.vmware.com
```

**Environment:**
kapp-controller 0.30.0

---
Vote on this request

joe-kimmel-vmw:

Fixes #549 

note that this is a more strict version of https://github.com/vmware-tanzu/carvel-kapp-controller/pull/550

In particular the difference is whether we allow all ports to skipTLS (as in #550 ) or enforce the port specified (as in this PR) - see [line 150 of the unit test](https://github.com/vmware-tanzu/carvel-kapp-controller/blob/82c1d7451817a03bcc23b238638d3c898752c7f7/pkg/config/config_test.go#L150)

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/fetch/vendir.go
+++ b/pkg/fetch/vendir.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"strings"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
@@ -356,20 +355,15 @@ func (v *Vendir) configMapBytes(configMapRef vendirconf.DirectoryContentsLocalRe
 // expand this option to other fetch options, we will need to add hostname
 // extraction for those
 func (v *Vendir) shouldSkipTLSVerify(url string) bool {
-	hostname, hostAndPort := v.extractImageRefHostname(url)
-	skip := v.skipTLSConfig.ShouldSkipTLSForDomain(hostname)
-	if !skip && (hostname != hostAndPort) {
-		// sometimes people want to whitelist only a specific port of a host, so we'll try that as a fallback.
-		return v.skipTLSConfig.ShouldSkipTLSForDomain(hostAndPort)
-	}
-	return skip
+	hostAndPort := v.extractImageRefHostname(url)
+	return v.skipTLSConfig.ShouldSkipTLSForAuthority(hostAndPort)
 }
 
-func (v *Vendir) extractImageRefHostname(ref string) (string, string) {
+func (v *Vendir) extractImageRefHostname(ref string) string {
 	parsedRef, err := name.ParseReference(ref)
 	if err != nil {
-		return "", ""
+		fmt.Println("error extracting image ref hostname: ", err)
+		return ""
 	}
-	hostnameAndPort := parsedRef.Context().RegistryStr()
-	return strings.Split(hostnameAndPort, ":")[0], hostnameAndPort
+	return parsedRef.Context().RegistryStr()
 }
--- a/pkg/fetch/vendir.go
+++ b/pkg/fetch/vendir.go
@@ -351,18 +351,22 @@ func (v *Vendir) configMapBytes(configMapRef vendirconf.DirectoryContentsLocalRe
 // expand this option to other fetch options, we will need to add hostname
 // extraction for those
 func (v *Vendir) shouldSkipTLSVerify(url string) bool {
-	hostname := v.extractImageRefHostname(url)
+	hostname, hostAndPort := v.extractImageRefHostname(url)
 	skip := v.skipTLSConfig.ShouldSkipTLSForDomain(hostname)
+	if !skip && (hostname != hostAndPort) {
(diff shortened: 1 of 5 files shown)


pkg/fetch/vendir.go
```diff
@@ -65,6 +65,11 @@ func (v *Vendir) AddDir(fetch v1alpha1.AppFetch, dirPath string) error {
 	return nil
 }
 
+// GetConfig is just for accessing (a copy of) the internal config for testing; you probably don't want to call this IRL
+func (v *Vendir) GetConfig() vendirconf.Config {
```
cppforlife: nit: lets rename to Config() (it will be more consistent with ConfigReader; arguably we should move this next to that one.)

pkg/fetch/vendir.go
```diff
@@ -351,18 +356,20 @@ func (v *Vendir) configMapBytes(configMapRef vendirconf.DirectoryContentsLocalRe
 // expand this option to other fetch options, we will need to add hostname
 // extraction for those
 func (v *Vendir) shouldSkipTLSVerify(url string) bool {
-	hostname := v.extractImageRefHostname(url)
+	hostname, hostAndPort := v.extractImageRefHostname(url)
 	skip := v.skipTLSConfig.ShouldSkipTLSForDomain(hostname)
```
cppforlife: i think we should move all the logic responsible for checking domain/domain+port into ShouldSkipTLSForDomain. probably also would want to use https://pkg.go.dev/net#SplitHostPort there internally.

ShouldSkipTLSForDomain could be rename to ShouldSkipTLSForHostPort (or something along those lines).

joe-kimmel-vmw: IMO `SplitHostPort` is less ergonomic than checking for / splitting on `:` ourselves 

https://go.dev/play/p/1Ziio3WSeJ7

cppforlife: it covers other cases that strings.Split does not. take a look at the source for SplitHostPort in go docs.

joe-kimmel-vmw: these, right: https://cs.opensource.google/go/go/+/refs/tags/go1.17.8:src/net/ipsock.go;l=164

so the two error cases it covers is
const (
		[missingPort](https://cs.opensource.google/go/go/+/refs/tags/go1.17.8:src/net/ipsock.go;bpv=1;bpt=1;l=166?gsn=missingPort&gs=kythe%3A%2F%2Fgolang.org%3Flang%3Dgo%3Fpath%3Dnet%23const%2520%255Bipsock.go%2523166%255D.missingPort)   = "missing port in address"
		[tooManyColons](https://cs.opensource.google/go/go/+/refs/tags/go1.17.8:src/net/ipsock.go;bpv=1;bpt=1;l=167?gsn=tooManyColons&gs=kythe%3A%2F%2Fgolang.org%3Flang%3Dgo%3Fpath%3Dnet%23const%2520%255Bipsock.go%2523167%255D.tooManyColons) = "too many colons in address"
	)
	
of which, one of them is pretty explicitly not an error case (we're ok with missing ports).  and the other one we'd swallow/ignore anyway under the current behavior of the functions (this stack doesn't currently return or surface errors very well, but that's a larger change ... e.g. in testing it i realized that if i pass in `registry.com/myCamelCasedOrg/image:tag` then actually something in gocontainerregistry rejects the URL because it contains capital letters,  but we swallow that error. )

joe-kimmel-vmw: just noting here for reference,  i never knew this but i guess `host:port` is called [authority](https://www.rfc-editor.org/rfc/rfc3986#section-3)

PR #565, opened 2022-03-09
accept values over stdin in helm template

cppforlife:
is useful for mangling values via ytt before sending them to helm as values. this may be used to limit allowed configurations for example.

##### Review Checklist:

  a link to that PR
  change

### DIFF
--- a/pkg/template/helm_template.go
+++ b/pkg/template/helm_template.go
@@ -39,12 +39,22 @@ func NewHelmTemplate(opts v1alpha1.AppTemplateHelmTemplate,
 }
 
 func (t *HelmTemplate) TemplateDir(dirPath string) (exec.CmdRunResult, bool) {
+	return t.template(dirPath, nil), true
+}
+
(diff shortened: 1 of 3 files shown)


neil-hickey (approved):
LGTM

pkg/template/helm_template.go
```diff
@@ -74,12 +84,18 @@ func (t *HelmTemplate) TemplateDir(dirPath string) (exec.CmdRunResult, bool) {
...
 		if err != nil {
-			return exec.NewCmdRunResultWithErr(err), true
+			return exec.NewCmdRunResultWithErr(err)
 		}
 
 		defer valuesCleanUpFunc()
 
 		for _, path := range paths {
+			if path == stdinPath {
+				if input == nil {
```
neil-hickey: can these be folded into `if path == stdinPath && input == nil` ?

cppforlife: fixed

PR #567, opened 2022-03-10
Add a misspell linter

benmoss:
Adds a misspell linter using https://github.com/reviewdog/action-misspell

It will only complain about new misspellings, ones already in our codebase are grandfathered in :)

Example PR showing this here: https://github.com/benmoss/carvel-kapp-controller/pull/6

### DIFF
--- a/.github/workflows/misspell.yml
+++ b/.github/workflows/misspell.yml
@@ -1,0 +1,18 @@
+name: reviewdog
+on: [pull_request]
+jobs:
+  misspell:
+    name: runner / misspell
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code.
+        uses: actions/checkout@v2
+      - name: misspell
+        uses: reviewdog/action-misspell@e71015afcf6b31e78b03ac2bc15a8422a0fc13d5 # v1.1.0
+        with:
+          fail_on_error: true
+          locale: "US"
+          exclude: |
+            ./vendor/*
+            ./cli/vendor/*
+            *generated*


.github/workflows/misspell.yml
```diff
@@ -0,0 +1,18 @@
...
+on: [pull_request]
+jobs:
+  misspell:
+    name: runner / misspell
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code.
+        uses: actions/checkout@v2
+      - name: misspell
+        uses: reviewdog/action-misspell@v1
```
cppforlife: this seems somewhat unknown repo (https://github.com/reviewdog/action-misspell) and it seems to have github_token available to it.

- is there more "common" action for this
- can we lock action to digest so that it doesnt "change" under us without review
- can we eliminate/scope what the token can do? (what can it do today?)

danielhelfand:
I believe you can just add misspell in the .golangci.yml: https://github.com/vmware-tanzu/carvel-kapp-controller/blob/405528350a13e2995d7bc99705794c61f0e9fb43/.golangci.yml#L4.

.github/workflows/misspell.yml
benmoss: it's good enough for google: https://github.com/google/go-containerregistry/blob/main/.github/workflows/style.yaml#L33

Let me see about constraining permissions

benmoss:
>I believe you can just add misspell in the .golangci.yml:

i dont think that would do anything about non-Go files, would it?

danielhelfand:
> i dont think that would do anything about non-Go files, would it?

Ah, yeah, the linter does expose a way to include non go files, but I think golangci-lint doesn't allow it to be configured.

.github/workflows/misspell.yml
benmoss: It looks like we can get away without the token and it will log the errors instead of posting a review

Issue #300, opened 2021-07-19
Better UX for Including Prereleases in PackageInstall versionSelection 


danielhelfand:
**Describe the problem/challenge you have**

For PackageInstall's referencing a [prerelease package](https://github.com/vmware-tanzu/carvel-kapp-controller/blob/84fbfc9e426bfc9b4200319c38de29b0b7f82598/test/e2e/assets/kc-e2e-test-repo/packages/pkg.test.carvel.dev/pkg.test.carvel.dev.3.0.0-rc.1.yml), a user must provide the following as part of a PackageInstall `.spec.packageRef` to successfully install the prerelease version:

```yaml
versionSelection:
  constraints: 3.0.0-rc.1
  prereleases: {} # <----------
```

Prereleases must not be null as documented in #116. As of the current implementation in kapp-controller, the user is responsible for configuring this setting via the PackageInstall `.spec.packageRef.versionSelection` property.

For users specifying a specific version to install (like above), it might be a nice enhancement to just always consider prereleases as opposed to making users configure this prerelease behavior. 

**Describe the solution you'd like**

To install a prerelease when a specific version is passed, I should not need to specify the `prereleases: {}` property above to install a prerelease version like below:

```yaml
versionSelection:
  constraints: 3.0.0-rc.1
```

**Anything else you would like to add:**

We should consider if there are other cases should have prereleases considered by default. 

We also have a test for the current behavior in place which can be adapted to TDD this feature out: https://github.com/vmware-tanzu/carvel-kapp-controller/blob/84fbfc9e426bfc9b4200319c38de29b0b7f82598/pkg/packageinstall/packageinstall_test.go#L19

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

aaronshurley:
@danielhelfand I noticed that the GH actions bot closed this out but I think it's still relevant?

PR #578, opened 2022-03-17
do not require prereleases:{} to be set when setting explicit version…

cppforlife:
… in constraints

currently if user specifies prerelease version explicitly in constraints field, they need to remember to specify `prereleases: {}` without which version selection wont work. this PR makes it unnecessary to specify prereleases when explicit prerelease version is given.

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/packageinstall/packageinstall.go
+++ b/pkg/packageinstall/packageinstall.go
@@ -210,6 +210,16 @@ func (pi *PackageInstallCR) referencedPkgVersion() (datapkgingv1alpha1.Package, 
 		}
 	}
 
+	// If constraint is a single specified version, then we
+	// do not want to force user to manually set prereleases={}
+	if len(semverConfig.Constraints) > 0 && semverConfig.Prereleases == nil {
+		// Will error if it's not a single version
+		singleVer, err := versions.NewSemver(semverConfig.Constraints)
+		if err == nil && len(singleVer.Pre) > 0 {
+			semverConfig.Prereleases = &verv1alpha1.VersionSelectionSemverPrereleases{}
+		}
+	}
+
 	verConfig := verv1alpha1.VersionSelection{Semver: semverConfig}
 
 	selectedVersion, err := versions.HighestConstrainedVersion(versionStrs, verConfig)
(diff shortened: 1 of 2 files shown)


cppforlife:
seems to address https://github.com/vmware-tanzu/carvel-kapp-controller/issues/300

pkg/packageinstall/packageinstall.go
```diff
@@ -210,6 +210,16 @@ func (pi *PackageInstallCR) referencedPkgVersion() (datapkgingv1alpha1.Package,
 		}
 	}
 
+	// If constraint is a single specified version, then we
+	// do not want to force user to manually set prereleases={}
+	if len(semverConfig.Constraints) > 0 && semverConfig.Prereleases == nil {
+		// Will error if it's not a single version
+		singleVer, err := versions.NewSemver(semverConfig.Constraints)
```
neil-hickey, line 217: Are we ignoring the error because a following stage will error instead?

cppforlife: constraints could be various things: single version or version ranges. if we fail to parse it as single version then it's either malformed version or it's not a single version. in either one of those cases we dont want to mangle what the user provided since we do not know what it is.

neil-hickey: oh nice, I didn't realise it could be a range. Makes sense!

pkg/packageinstall/packageinstall_test.go
```diff
@@ -70,6 +70,46 @@ func Test_PackageRefWithPrerelease_IsFound(t *testing.T) {
...
+						Constraints: "3.0.0-rc.1",
+					},
+				},
+			},
+		},
+		pkgclient: fakePkgClient,
+	}
+
+	out, err := ip.referencedPkgVersion()
+	if err != nil {
```
neil-hickey: nit : I think we are moving towards using `testify/assert` for test assertions:

I think you could use `assert.NoErrorf(t, err, "desc")` and maybe theres a `reflect` equivalent?

cppforlife: fixed

joe-kimmel-vmw (approved):
lgtm

PR #579, opened 2022-03-17
Avoid retrying image-based app installs when nothing has changed

benmoss:
We introduced retrying for image-based apps in #331 to handle the race condition between image pull secrets getting populated and us running `vendir sync`. However this results in unnecessary churn/repository traffic when the configmaps and secrets haven't changed, since there's not going to be any different outcome.

### DIFF
--- a/pkg/pkgrepository/app_fetch.go
+++ b/pkg/pkgrepository/app_fetch.go
@@ -6,7 +6,6 @@ package pkgrepository
 import (
 	"bytes"
 	"fmt"
-	"io"
 	"os"
 	"path"
 	"strconv"
@@ -37,13 +36,13 @@ func (a *App) fetch(dstPath string) (string, exec.CmdRunResult) {
 		}
 	}
 
-	confReader, err := vendir.ConfigReader()
+	conf, err := vendir.ConfigBytes()
 	if err != nil {
 		result.AttachErrorf("Fetching: %v", err)
 		return "", result
 	}
 
-	result = a.runVendir(confReader, dstPath)
+	result = a.runVendir(conf, dstPath)
 	// retry if error occurs before reporting failure.
 	// This is mainly done to support private registry
 	// authentication for images/bundles since placeholder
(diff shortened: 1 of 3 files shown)


pkg/fetch/vendir_test.go
```diff
@@ -1,15 +1,14 @@
 // Copyright 2022 VMware, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-package fetch_test
+package fetch
```
cppforlife: generally i recommend to keep them isolated so that we have enough "public" interface to check on whatever we need to check.

cppforlife:
lgtm.

pkg/pkgrepository/app_fetch.go
```diff
@@ -55,12 +54,16 @@ func (a *App) fetch(dstPath string) (string, exec.CmdRunResult) {
 			// Sleep for 2 seconds to allow secretgen-controller
```
neil-hickey, line 54:
```
		time.Sleep(2 * time.Second)
		newConf, err := vendir.ConfigBytes()
		if err != nil {
			result.AttachErrorf("Fetching: %v", err)
			return "", result
		}

		for i := 0; i < 2 && !bytes.Equal(conf, newConf); i++ {
			time.Sleep(2 * time.Second)
			newConf, err := vendir.ConfigBytes()
			if err != nil {
				result.AttachErrorf("Fetching: %v", err)
				return "", result
			}
			
			result = a.runVendir(newConf, dstPath)
			if result.Error == nil {
				break
			}
		}
```

How do you feel about this? I think the `continue` on line 64 will add 4 seconds of sleeping (by looping another 2 times potentially) that don't need to happen. Right?

Also I might be the against the go way of using `continue` instead of a loop termination. I just find it hard to read `continue`'s and even `breaks`. Causes my brain to jump around, and I'm not a computer (or a slow one)

pkg/fetch/vendir_test.go
benmoss: Yeah I started to think about refactoring this out but I'm not going to bother

pkg/pkgrepository/app_fetch.go
cppforlife: > on line 64 will add 4 seconds of sleeping (by looping another 2 times potentially) that don't need to happen. Right?

i think we do want that so that there is enough time for any change to kick to the secret. this would only happen if fetch failed in a first place.

benmoss: The idea is that after the next sleep `vendir.ConfigBytes()` might have returned a new Vendir configuration (with image pull secrets).

The flow is basically:
- resolve all the `secretRef`/`configMapRef`s and turn them into the [multi-doc config format](https://github.com/vmware-tanzu/carvel-vendir/blob/develop/examples/secrets/vendir.yml) that vendir supports 
- run `vendir sync`
- if it fails and one of our fetches is an `image:` or `imgpkg:`
  - resolve all the secrets/configmaps again
  - if they're the same, sleep/retry
  - if they're different run `vendir sync` again
  - if it fails again, sleep/retry

neil-hickey: > The idea is that after the next sleep vendir.ConfigBytes() might have returned a new Vendir configuration (with image pull secrets).

gotcha! Coolio. Can disregard the comment so 👍

PR #580, opened 2022-03-17
Add `includedSoftware` field to `Package` spec

### LINKED ISSUE #212: Need Ability to Specify Package Contents/Version

Packages need the ability to specify/declare the software that they contain.

* Package contains 1 piece of software

    Consider the cases where I have multiple packages that bundle the cert-manager software at various versions. 

    | Package Name | Package Version | Contents | Contents Version |
    | --------------- | ---------------- | --------- | ----------------- |
    | cert-mgr | 1.0.0 | cert-manager | 1.1.0 |
    | cert-mgr | 1.0.1 | cert-manager | 1.1.1 |
    | cert-mgr | 1.1.0 | cert-manager | 1.2.0 |
    | cert-mgr | 1.2.0 | cert-manager | 1.3.0 |

    When I load these packages to a PackageRepository and retrieve the list of packages from my cluster, I'll get:

    ```
    NAME                             VERSION         DESCRIPTION
    cert-mgr                         1.0.0
    cert-mgr                         1.0.1
    cert-mgr                         1.1.0
    cert-mgr                         1.2.0
    ```
    There is no indication of what is actually in the package, unless I were to update that information into the description.

* The problem gets worse when there is multiple pieces of software in the package. Consider the case of Prometheus and Alert-Manager bundled together.

    | Package Name | Package Version | Contents | Contents Version |
    | --------------- | ---------------- | --------- | ----------------- |
    | tobs | 1.0.0 | prometheus, alert-manager | 2.25.0, 0.21.0 |
    | tobs | 1.1.0 | prometheus, alert-manager | 2.26.0, 0.22.0 |

    When I load these packages to a PackageRepository and retrieve the list of packages from my cluster, I'll get:

    ```
    NAME                             VERSION         DESCRIPTION
    tobs                             1.0.0
    tobs                             1.1.0
    ```

I would like the ability to list in the `packageVersion` CR what software is contained in the package and I would like to be able to see that information when retrieved from the cluster.

Modify the `packageVersion` CR:

```yaml
---
apiVersion: package.carvel.dev/v1alpha1
kind: Package
metadata:
  name: simple-app.corp.com.1.0.0
spec:
  refName: simple-app.corp.com
  version: 1.0.0
  software:
  - name: simple-app-controller
    version: 1.0.0
    description: the controller
  - name: foo-webhook
    version: 2.3.4
    description: a foo webhook
  - name: bar-webhook
    version: 3.4.5
    description: a bar webhook
  releaseNotes: |
        Initial release of the simple app package
...
```

I would then like to see that information through the command line

```
NAME                             VERSION         CONTENTS
simple-app.corp.com.1.0.0        1.0.0           simple-app-controller-1.0.0;foo-webhook-2.3.4;bar-webhook-3.4.5
```

---
Vote on this request

neil-hickey:
- Allows users to specify an array of included software within a package
  => it's name, version and description

Fixes #212

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```


neil-hickey:
1. I am not so sure showing the package contents as part of `kubectl get` is a great idea. ([all in this commit which can be reverted](https://github.com/vmware-tanzu/carvel-kapp-controller/pull/580/commits/046331748ec807ebbd221f9a953889d2e926ad5d))
2. `--field-selector=` as suggested in the issue is very limited to the things it can filter, and custom types are not part of that.
3. The only real way to see this is with some `-o jsonpath` filtering or using `kubectl describe`. OR finally relying on the new `kctrl` to show the info?

WDYT @cppforlife ?

Screenshot 2022-03-17 at 17 24 43

pkg/apiserver/registry/datapackaging/package_crd_rest.go
```diff
@@ -340,6 +347,7 @@ func (r *PackageCRDREST) ConvertToTable(ctx context.Context, obj runtime.Object,
 			{Name: "PackageMetadata Name", Type: "string", Format: "name", Description: "Associated PackageMetadata name"},
 			{Name: "Version", Type: "string", Description: "Version"},
 			{Name: "Age", Type: "date", Description: "Time since resource creation"},
+			{Name: "Included Software", Type: "array", Description: "Software Included in Package"},
```
cppforlife: i dont think we should show this info in the table:
- it becomes too wide for most screens
- most packages cover only single piece of software so package version is a good proxy (details can always be obtained via kubctl get or eventually in kctrl)

pkg/apiserver/apis/datapackaging/types.go
```diff
@@ -127,3 +132,10 @@ type ValuesSchema struct {
 	// +kubebuilder:pruning:PreserveUnknownFields
 	OpenAPIv3 runtime.RawExtension `json:"openAPIv3,omitempty"`
 }
+
+// IncludedSoftware contains the underlying Software Contents of a Package
+type IncludedSoftware struct {
+	Name        string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
```
cppforlife: should we call this displayName since im assuming we are planning to use this for human oriented names vs like cert-manager.x.x

pkg/apiserver/apis/datapackaging/types.go
```diff
@@ -127,3 +132,10 @@ type ValuesSchema struct {
...
+	Version     string `json:"version,omitempty" protobuf:"bytes,2,opt,name=version"`
```
cppforlife, line 139: i wonder if we can highlight that this version is a "third party version" somehow. eg doesnt have to follow strict semver format Package's spec.version needs to follow

neil-hickey:
Looking for opinions on field naming:
- @seemiller do you all have any concrete use cases for these fields? Would really help us narrow down the naming conventions.
- Does accessing through `kubectl describe` and eventually `kctrl` seem sufficient to you?

pkg/apiserver/registry/datapackaging/package_crd_rest.go
neil-hickey: I agree, leaving this comment open for @seemiller to take a look. But I have reverted the change for now.

seemiller:
@neil-hickey [App-toolkit](https://github.com/vmware-tanzu/community-edition/pull/3568) was just added to TCE as the first package that bundles other packages, so that's a  pretty concrete use case. This change will greatly enhance the ability of package authors to identify what software is included.

I agree that including a potentially large amount of information in a standard `kubectl get pkg/...` call would not be helpful. Using `describe`, or adding a `-o wide` option, would suffice for a first cut of this in my opinion.

seemiller:
Given that "a package of packages" is typically implemented as a package that contains PackageInstall CRDs, would it be possible for kapp-controller to introspect the package for this and display the include software name/version/description?

pkg/apiserver/apis/datapackaging/types.go
neil-hickey: I like `displayName` 👍

pkg/apiserver/apis/datapackaging/types.go
neil-hickey: Not that I can think of - not semantically through a name.

kartiklunkad26:
this is really valuable! I'm glad we are doing it. I know @jorgemoralespou will like it. thanks @seemiller for letting me know about this PR. 

> Looking for opinions on field naming: 

Could we just includedPackages given this is mostly going to be applicable for package of packages? Or is the use-case also for packages that contain multiple pieces of software? Not too strongly opinionated here. `includedSoftware` is fine too IMO. 

>Does accessing through kubectl describe and eventually kctrl seem sufficient to you?

I would definitely want it to be enabled by `tanzu package` eventually, but I understand as a starting point `kubectl` and `kctrl` are good places to start. 

>Given that "a package of packages" is typically implemented as a package that contains PackageInstall CRDs, would it be possible for kapp-controller to introspect the package for this and display the include software name/version/description?

I agree with this. One question: 

-  Do we plan to have a more concrete/first-class name for the concept of package of packages?

jorgemoralespou:
This should be for any package

cppforlife:
> This should be for any package

i can confirm that this is just plain metadatat -- applied to any package.

> Do we plan to have a more concrete/first-class name for the concept of package of packages?

no, we do not at this point.

neil-hickey:
Thanks for the input everyone! 

> Given that "a package of packages" is typically implemented as a package that contains PackageInstall CRDs, would it be possible for kapp-controller to introspect the package for this and display the include software name/version/description?

Not currently, and I don't expect this is something we could support. The only way for KC to "tell" that a Package contains Packages is to look at the output and then parse those yaml files to get info.

> Could we just includedPackages given this is mostly going to be applicable for package of packages? Or is the use-case also for packages that contain multiple pieces of software? Not too strongly opinionated here. includedSoftware is fine too IMO.

~~`includedPackages` makes sense to me.~~ Actually I think that brings up the idea of a software version being different from a package version which is what we are trying to highlight here.

pkg/apiserver/apis/datapackaging/types.go
neil-hickey: Any follow ups on this suggestion @cppforlife before I go ahead and mark this as ready to go without any changes to the `Version` field?

joe-kimmel-vmw: the thing that we wish we could convey through the name, i want to make sure i understand it:  from the package's perspective, this version field is "theirs" and not "ours" ?

joe-kimmel-vmw: if so,  options could include
- TheirVersion
- UnderlyingVersion
- PackagedVersion

(of those i'd vote for the last one,  but 🤷 )

jorgemoralespou: I personally think that Version is what users will understand as it relates to the Name of the IncludedSoftware. And to be honest, this is the one name I would change (but 🤷‍♂️ )

config/crds.yml
```diff
@@ -95,6 +95,19 @@ spec:
...
                 description: 'System requirements needed to install the package. Note: these requirements will not be verified by kapp-controller on installation. (optional; string)'
                 type: string
+              includedSoftware:
+                description: IncludedSoftware can be used to show the software contents of a Package. This is especially useful if the underlying versions do not match the Package version
+                items:
+                  description: IncludedSoftware contains the underlying Software Contents of a Package
+                  properties:
+                    description:
+                      type: string
+                    name:
```
benmoss:
```suggestion
                    displayName:
```

pkg/apiserver/apis/datapackaging/types.go
```diff
@@ -127,3 +132,10 @@ type ValuesSchema struct {
 	// +kubebuilder:pruning:PreserveUnknownFields
 	OpenAPIv3 runtime.RawExtension `json:"openAPIv3,omitempty"`
 }
+
+// IncludedSoftware contains the underlying Software Contents of a Package
+type IncludedSoftware struct {
+	DisplayName string `json:"name,omitempty" protobuf:"bytes,1,opt,name=displayName"`
```
benmoss:
```suggestion
	DisplayName string `json:"displayName,omitempty" protobuf:"bytes,1,opt,name=displayName"`
```

pkg/apiserver/apis/datapackaging/v1alpha1/types.go
```diff
@@ -101,3 +106,10 @@ type ValuesSchema struct {
 	// +kubebuilder:pruning:PreserveUnknownFields
 	OpenAPIv3 runtime.RawExtension `json:"openAPIv3,omitempty" protobuf:"bytes,1,opt,name=openAPIv3"`
 }
+
+// IncludedSoftware contains the underlying Software Contents of a Package
+type IncludedSoftware struct {
+	DisplayName string `json:"name,omitempty" protobuf:"bytes,1,opt,name=displayName"`
```
benmoss:
```suggestion
	DisplayName string `json:"displayName,omitempty" protobuf:"bytes,1,opt,name=displayName"`
```

benmoss (changes requested):
Will probably need to rerun codegen too

neil-hickey:
hi @seemiller / @kartiklunkad26 Do you all have any sense of urgency on this request? We are not planning on cutting a new KC immediately with this change, so if it is urgent we could do that.

kartiklunkad26:
When would a new KC be available with this change? If it is a few weeks/a month, then I think that's fine.

If it's much more than that, then that would feel very late.

PR #591, opened 2022-03-28
Allow dynamic KC config updates

### LINKED ISSUE #179: Watch kapp-controller-config configmap and apply updated config
Currently, kapp-controller loads and applies the kapp-controller-config configmap at startup. This means if the configmap is updated, the kapp-controller pod will need to be restarted in order to pick up that changes.

kapp-controller should dynamically reconfigure itself if there are updates to the configmap.

N/A

---
Vote on this request

neil-hickey:

Fixes #179

```release-note

```

See [this comment](https://github.com/vmware-tanzu/carvel-kapp-controller/issues/179#issuecomment-1078230688) for some concerns and questions before I implemented this feature

- I had a really hard time trying to test this. The reconciler is fairly dumb, so that felt unnecessary. And the config change I wanted an e2e test, but `kubectl exec` was failing from the test suite - always with errors like 

`Running 'kubectl exec -n kapp-controller deployment/kapp-controller -- bash -c ls'...
OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "bash -c ls": executable file not found in $PATH: unknown
`

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -80,6 +80,9 @@ COPY --from=0 /usr/local/bin/kapp .
 # Name it kapp-controller to identify it easier in process tree
 COPY --from=0 /go/src/github.com/vmware-tanzu/carvel-kapp-controller/controller kapp-controller
 
+# Copy the ca-bundle so we have an original
+RUN cp /etc/pki/tls/certs/ca-bundle.crt /etc/pki/tls/certs/ca-bundle.crt.orig
+
 # Run as kapp-controller by default, will be overridden to a random uid on OpenShift
 USER 1000
 ENV PATH="/:${PATH}"
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,8 +59,9 @@ RUN tdnf install -y git openssh-clients sed
 
 # Create the kapp-controller user in the root group, the home directory will be mounted as a volume
 RUN echo "kapp-controller:x:1000:0:/home/kapp-controller:/usr/sbin/nologin" > /etc/passwd
-# Give the root group write access to openssh's root bundle so we can append custom roots at runtime
-RUN chmod g+w /etc/pki/tls/certs/ca-bundle.crt
+# Give the root group write access to the openssh's root bundle directory
+# so we can rename the certs file with our dynamic config, and append custom roots at runtime
+RUN chmod g+w /etc/pki/tls/certs
 
 # fetchers
 COPY --from=0 /helm-unpacked/linux-amd64/helm .
--- a/cmd/controller/run.go
+++ b/cmd/controller/run.go
@@ -118,6 +118,26 @@ func Run(opts Options, runLog logr.Logger) error {
 	refTracker := reftracker.NewAppRefTracker()
 	updateStatusTracker := reftracker.NewAppUpdateStatus()
 
+	{ // add controller for config
+		reconciler := kcconfig.NewReconciler(coreClient, runLog.WithName("config"))
+
+		ctrl, err := controller.New("Config", mgr, controller.Options{
+			Reconciler: NewUniqueReconciler(&ErrReconciler{
+				delegate: reconciler,
+				log:      runLog.WithName("er"),
+			}),
+			MaxConcurrentReconciles: 1,
+		})
+		if err != nil {
+			return fmt.Errorf("Setting up Config reconciler: %s", err)
+		}
+
+		err = reconciler.AttachWatches(ctrl)
+		if err != nil {
+			return fmt.Errorf("Setting up Config reconciler watches: %s", err)
+		}
+	}
+
 	{ // add controller for apps
 		appFactory := app.CRDAppFactory{coreClient, kcClient, kcConfig, appMetrics}
 		reconciler := app.NewReconciler(kcClient, runLog.WithName("app"),
--- a/cmd/controller/run.go
+++ b/cmd/controller/run.go
@@ -122,10 +122,7 @@ func Run(opts Options, runLog logr.Logger) error {
 		reconciler := kcconfig.NewReconciler(coreClient, runLog.WithName("config"))
 
 		ctrl, err := controller.New("Config", mgr, controller.Options{
-			Reconciler: NewUniqueReconciler(&ErrReconciler{
-				delegate: reconciler,
-				log:      runLog.WithName("er"),
-			}),
+			Reconciler:              reconciler,
 			MaxConcurrentReconciles: 1,
 		})
 		if err != nil {
--- a/cmd/controller/run.go
+++ b/cmd/controller/run.go
@@ -129,7 +129,12 @@ func Run(opts Options, runLog logr.Logger) error {
 			return fmt.Errorf("Setting up Config reconciler: %s", err)
 		}
 
-		err = reconciler.AttachWatches(ctrl)
+		ns := os.Getenv("KAPPCTRL_SYSTEM_NAMESPACE")
+		if ns == "" {
+			return fmt.Errorf("Cannot get kapp-controller namespace")
+		}
+
+		err = reconciler.AttachWatches(ctrl, ns)
 		if err != nil {
 			return fmt.Errorf("Setting up Config reconciler watches: %s", err)
 		}
--- a/cmd/controller/run.go
+++ b/cmd/controller/run.go
@@ -121,7 +121,7 @@ func Run(opts Options, runLog logr.Logger) error {
 	{ // add controller for config
 		reconciler := kcconfig.NewReconciler(coreClient, runLog.WithName("config"))
 
-		ctrl, err := controller.New("Config", mgr, controller.Options{
+		ctrl, err := controller.New("config", mgr, controller.Options{
 			Reconciler:              reconciler,
 			MaxConcurrentReconciles: 1,
 		})
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -157,30 +157,36 @@ func (gc *Config) addTrustedCerts(certChain string) (err error) {
 	if err != nil {
 		return fmt.Errorf("Opening original certs file: %s", err)
 	}
+	defer backupFile.Close()
 
 	tmpFile, err := os.CreateTemp(os.TempDir(), "tmp-ca-bundle-")
 	if err != nil {
 		return fmt.Errorf("Creating tmp certs file: %s", err)
 	}
+	defer os.Remove(tmpFile.Name())
 
 	_, err = io.Copy(tmpFile, backupFile)
 	if err != nil {
+		_ = tmpFile.Close()
 		return fmt.Errorf("Copying certs file: %s", err)
 	}
 
 	_, err = tmpFile.Write([]byte("\n" + certChain))
 	if err != nil {
-		_ = backupFile.Close()
-		_ = os.Remove(tmpFile.Name())
+		_ = tmpFile.Close()
+		return err
+	}
+
+	if err = tmpFile.Close(); err != nil {
 		return err
 	}
 
 	err = os.Rename(tmpFile.Name(), systemCertsFilePath)
 	if err != nil {
-		return fmt.Errorf("renaming certs file: %s", err)
+		return fmt.Errorf("Renaming certs file: %s", err)
 	}
 
-	return backupFile.Close()
+	return nil
 }
 
 func (gc *Config) configureProxies() {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -22,8 +22,9 @@ import (
 const (
 	kcConfigName = "kapp-controller-config"
 
-	caCertsKey      = "caCerts"
-	systemCertsFile = "/etc/pki/tls/certs/ca-bundle.crt"
+	caCertsKey            = "caCerts"
+	systemCertsFile       = "/etc/pki/tls/certs/ca-bundle.crt"
+	backupSystemCertsFile = "/etc/pki/tls/certs/ca-bundle.crt.orig"
 
 	httpProxyKey     = "httpProxy"
 	httpsProxyKey    = "httpsProxy"
@@ -41,12 +42,13 @@ const (
 // Config is populated from the cluster's Secret or ConfigMap and sets behavior of kapp-controller.
 // NOTE because config may be populated from a Secret use caution if you're tempted to serialize.
 type Config struct {
-	caCerts       string
-	httpProxy     string
-	httpsProxy    string
-	noProxy       string
-	skipTLSVerify string
-	SkipCertCheck bool
+	caCerts            string
+	httpProxy          string
+	httpsProxy         string
+	noProxy            string
+	skipTLSVerify      string
+	BackupCaBundlePath string
+	SystemCaBundlePath string
 }
 
 // findExternalConfig will populate exactly one of its return values and the others will be nil.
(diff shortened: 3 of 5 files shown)


neil-hickey:
### Implementation Notes:

#### What happens if I update my ca bundle whilst KC is running?

- I added the PEM for https://self-signed.badssl.com/ to the `caCerts` field of our KC configmap. Verified I got an error with `curl https://self-signed.badssl.com/` before updating the configmap and then after updating I ran `curl https://self-signed.badssl.com/` again whilst `kubectl exec'd` on the container and it worked.

#### But what about KC process? Does it pick up the change?

It does not. I added in a `http.Get("https://self-signed.badssl.com/ "` into one of the controllers. After applying the secret and updating the config - the curl still fails with a bad cert error. Upon running `kubectl rollout restart deployment/kapp-controller -n kapp-controller` that `http.Get()` no longer fails. So any http calls within KC will not get the new certs dynamically until the next restart. This feels like a niche case though, as KC is mostly just a fancy `exec()` for the fetch / template / deploy tools.

> NOTE: If we do ever move to using the go library version of these tools instead of invoking binaries, we will likely need to revisit this implementation

#### What about GoRoutine race conditions whilst `os.Setenv`'ing?

I validated that the `os.Setenv` func and `os.Getenv` are [guarded by a r/w lock](https://github.com/golang/go/blob/master/src/syscall/env_unix.go#L115-L116) in golang. So only one goroutine can access it at once. I also printed out the environment variables in each controller to validate that they were picking up the new values dynamically.

#### Is `os.Getenv()` global / shared amongst go routines?

Yep - checkout my playground example https://go.dev/play/p/mJoS_uwD6yC

#### Do we need to have a `lock` around the `/etc/pki/tls/certs/ca-bundle.crt` file?

Probably not. Again golang has a `lock` around the [write function](https://github.com/golang/go/blob/master/src/internal/poll/fd_unix.go#L369-L402). So we should be good here also (given we are not expecting any other process to update the system file.) - which is the assumption the previous code here held

neil-hickey:
Hmm, I need to give some more thought on our `append to /etc/pki/tls/certs/ca-bundle.crt` approach. As every update of the config appends the certs to that file. So we get lots of duplication.

Dockerfile
```diff
@@ -87,6 +87,9 @@ COPY --from=0 /usr/local/bin/kapp .
 # Name it kapp-controller to identify it easier in process tree
 COPY --from=0 /go/src/github.com/vmware-tanzu/carvel-kapp-controller/controller kapp-controller
 
+# Copy the ca-bundle so we have an original
+RUN cp /etc/pki/tls/certs/ca-bundle.crt /etc/pki/tls/certs/ca-bundle.crt.orig
```
neil-hickey, line 85: I chose `RUN cp` over `COPY` because `COPY` was failing with a "file not found error". Seems like something to do with accessing the `/etc/` folder via the `COPY` syntax.

benmoss:
This seems like a bug:
- Create a config (I just copy/pasted the one from https://carvel.dev/kapp-controller/docs/v0.34.0/controller-config/#controller-configuration-spec)
- See that it gets applied
- Delete that secret
- Configuration doesn't change back

I think it's because we noop if we don't find a ConfigMap or Secret, when instead we need to somehow revert to defaults

benmoss:
This workflow seems problematic:
- Create a config (I just copy/pasted the one from https://carvel.dev/kapp-controller/docs/v0.34.0/controller-config/#controller-configuration-spec)
- See that it gets applied
- Delete that secret
- Configuration doesn't change back

neil-hickey:
Good find! Yeh, I wasn't really sure about that case. Like if you delete the config accidentally and then everything breaks (proxy config is revert to defaults etc), is that the UX we want?

Seems like you were expecting yes, I wasn't really so sure.

Dockerfile
benmoss: I think COPY is for copying from either another build stage or from the build context

neil-hickey:
> This seems like a bug:
> 
>     * Create a config (I just copy/pasted the one from https://carvel.dev/kapp-controller/docs/v0.34.0/controller-config/#controller-configuration-spec)
> 
>     * See that it gets applied
> 
>     * Delete that secret
> 
>     * Configuration doesn't change back
> 
> 
> I think it's because we noop if we don't find a ConfigMap or Secret, when instead we need to somehow revert to defaults

Good point @benmoss  ! I updated the code a bit to accommodate that case. Basically always apply the config (even if there doesn't exist one) -- therefore just set the default (empty string) values.

benmoss:
Nice, I was confused by that `populated` bool anyway, not needing it makes the code easier to understand

benmoss (approved):

pkg/config/config.go
```diff
@@ -45,7 +46,7 @@ type Config struct {
 	httpsProxy    string
 	noProxy       string
 	skipTLSVerify string
-	populated     bool
+	SkipCertCheck bool
```
benmoss: Not a big fan of this approach, I would just make it configurable where the certs get read/written to and then have a unit test that uses a temp dir or a Reader/Writer

benmoss (approved):

cmd/controller/run.go
```diff
@@ -114,6 +114,26 @@ func Run(opts Options, runLog logr.Logger) error {
 	refTracker := reftracker.NewAppRefTracker()
 	updateStatusTracker := reftracker.NewAppUpdateStatus()
 
+	{ // add controller for config
+		reconciler := kcconfig.NewReconciler(coreClient, runLog.WithName("config"))
+
+		ctrl, err := controller.New("Config", mgr, controller.Options{
+			Reconciler: NewUniqueReconciler(&ErrReconciler{
```
cppforlife: i dont think you want to be using `NewUniqueReconciler(&ErrReconciler{`, instead i think it should just be `reconciler`.

pkg/config/config.go
```diff
@@ -150,12 +145,36 @@ func (gc *Config) ShouldSkipTLSForAuthority(candidateAuthority string) bool {
...
+		systemCertsFilePath = gc.SystemCaBundlePath
+	}
+
+	src, err := os.OpenFile(backupCertsFilePath, os.O_RDONLY, os.ModeExclusive)
+	if err != nil {
+		return fmt.Errorf("Opening original certs file: %s", err)
+	}
+	defer src.Close()
+
+	dst, err := os.OpenFile(systemCertsFilePath, os.O_CREATE|os.O_RDWR|os.O_TRUNC, os.ModeExclusive)
```
cppforlife: is there more details on what exactly ModeExclusive does?

pkg/config/config.go
```diff
@@ -150,12 +145,36 @@ func (gc *Config) ShouldSkipTLSForAuthority(candidateAuthority string) bool {
...
+	if err != nil {
+		return fmt.Errorf("Opening certs file: %s", err)
+	}
+	defer dst.Close()
```
cppforlife: we should be checking on close errors. flushing might happen in close for writes.

pkg/config/config.go
```diff
@@ -150,12 +145,36 @@ func (gc *Config) ShouldSkipTLSForAuthority(candidateAuthority string) bool {
...
+		return fmt.Errorf("Copying certs file: %s", err)
+	}
+
 	if certChain == "" {
 		return nil
 	}
 
-	var file *os.File
-	file, err = os.OpenFile(systemCertsFile, os.O_CREATE|os.O_APPEND|os.O_WRONLY, os.ModeAppend)
+	file, err := os.OpenFile(systemCertsFilePath, os.O_APPEND|os.O_WRONLY, os.ModeAppend)
```
cppforlife: in general this seems racy to me in combination with above.

pkg/config/reconciler.go
```diff
@@ -0,0 +1,70 @@
...
+
+	err := controller.Watch(&source.Kind{Type: &v1.ConfigMap{}}, &handler.EnqueueRequestForObject{}, p)
+	if err != nil {
+		return fmt.Errorf("Watching Configmaps: %s", err)
+	}
+
+	err = controller.Watch(&source.Kind{Type: &v1.Secret{}}, &handler.EnqueueRequestForObject{}, p)
+	if err != nil {
+		return fmt.Errorf("Watching Secrets: %s", err)
+	}
```
cppforlife, line 49: are these somehow limited to kapp controller namespace?

pkg/config/reconciler.go
```diff
@@ -0,0 +1,70 @@
...
+
+	return nil
+}
+
+// nolint: revive
```
cppforlife: whats this for?

pkg/config/config.go
```diff
@@ -170,21 +189,15 @@ func (gc *Config) addTrustedCerts(certChain string) (err error) {
 }
 
 func (gc *Config) configureProxies() {
-	if gc.httpProxy != "" {
```
cppforlife, line 173: whats the thinking behind removing empty check? if it's to unset, shouldnt we be using Unsetenv?

pkg/config/config.go
```diff
@@ -99,10 +102,6 @@ func GetConfig(client kubernetes.Interface) (*Config, error) {
 }
 
 func (gc *Config) Apply() error {
-	if !gc.populated {
```
cppforlife, line 102: im curious what was trying to protect before?

pkg/config/reconciler.go
joe-kimmel-vmw: fwiw as a person who argued in favor of setting the linter to require comments, i fully support nolint directives like this one :)

joe-kimmel-vmw: @cppforlife we commented at the same time --   the nolint directive is saying "don't throw me a linter error for not having some terrible comment like `// Reconcile is the reconciler method for reconcilers`

cppforlife: i do not. i think we should either have linter or not have linter. sprinkling dont lint me comments feels dirty.

pkg/config/config.go
neil-hickey: The thinking behind it is: We want to always just set the variable to whatever we are given. So creates, updates and deletes are all just a simple `os.Set()`

If they delete the config, this will be set to an empty string. 

We *could* check if its empty and do the `os.Unset()` but that felt like more code for the same outcome ("" and unset are equivalent)

pkg/config/config.go
neil-hickey: It was a short-circuit. If the config that came back from the cluster was empty then we skip all of the file / system operations and just return early. I think it was an optimization rather than protecting anything in particular.

In our new case, an empty config coming back has meaning (we want to restore defaults) so this bool makes less sense now.

cmd/controller/run.go
neil-hickey: I opted for this because it handles alot of the `synchronization` problems that could arise by processing the requests one at a time. Would changing it back to the plain reconciler not introduce the potential for race conditions on requests in parallel?

pkg/config/reconciler.go
neil-hickey: no, they are currently not scoped by namespace at all. The config is searched for based on whatever the kubeconfig on file says for a namespace, but not filtered in the reconciler by namespace.

https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/pkg/config/config.go#L78-L80

Given this namespace could be dynamic (I'm thinking `tkg-system` or `kapp-controller`) we have two choices, listen for that config across all namespaces (current option) or load the ns from the kubeconfig and filter on that.

pkg/config/config.go
neil-hickey: I think this should just be `0600` instead of the exclusiveMode.

cmd/controller/run.go
cppforlife: MaxConcurrentReconciles=1 should take care of this.

pkg/config/config.go
cppforlife: i dont think we have assurance from various tools that http_proxy= (empty) is equivalent to no http_proxy at all.

pkg/config/config.go
cppforlife: (specifically, initial write or append are not atomic to readers of the file so they may actually see corrupt view of the file while we are modifying it)

pkg/config/config.go
neil-hickey: hmm, fair enough. Perhaps that assumption was misinformed. It's probably safer to unset the variable if there is a doubt on the behaviour.

pkg/config/reconciler.go
neil-hickey: I changed this so that it follows the same logic as `GetConfig()`, i.e it loads the namespace from the kubeconfig.

cmd/controller/run.go
```diff
@@ -114,6 +114,23 @@ func Run(opts Options, runLog logr.Logger) error {
 	refTracker := reftracker.NewAppRefTracker()
 	updateStatusTracker := reftracker.NewAppUpdateStatus()
 
+	{ // add controller for config
+		reconciler := kcconfig.NewReconciler(coreClient, runLog.WithName("config"))
+
+		ctrl, err := controller.New("Config", mgr, controller.Options{
```
cppforlife: nit: Config => config

pkg/config/reconciler.go
```diff
@@ -0,0 +1,78 @@
...
+
+// NewReconciler constructs new Reconciler.
+func NewReconciler(coreClient kubernetes.Interface, log logr.Logger) *Reconciler {
+	return &Reconciler{coreClient, log}
+}
+
+var _ reconcile.Reconciler = &Reconciler{}
+
+// AttachWatches configures watches needed for reconciler to reconcile the kapp-controller Config.
+func (r *Reconciler) AttachWatches(controller controller.Controller) error {
```
cppforlife: instead of loading kubeConfig here again, may be we should just pass namespace into this method from outside?

pkg/config/reconciler.go
```diff
@@ -0,0 +1,78 @@
...
+	return nil
+}
+
+// Reconcile gets the current config from the cluster and applies any changes.
+func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	log := r.log.WithValues("request", request)
+
+	kcConfig, err := GetConfig(r.coreClient)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("getting kapp-controller config: %s", err)
```
cppforlife: if Reconcile returns an error, next reconcile is immediately scheduled. i assume we do not want to immediately schedule another one.

pkg/config/config.go
cppforlife: @neil-hickey let me know if you want to talk about this.

pkg/config/reconciler.go
neil-hickey: Wow, deep in the reconciler it says:

```
// The Controller will requeue the Request to be processed again if an error is non-nil or
// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
```

The is somewhat contradictory (though technically not) to the `reconcile.Result{}` struct which indicated that you need to set the bool to true to get a new reconcile.

```
type Result struct {
    // Requeue tells the Controller to requeue the reconcile key.  Defaults to false.
    Requeue bool
```

I suspect you hit this before, I didn't even think to look that far down the code path to find that hidden comment on the Reconciler interface

pkg/config/reconciler.go
neil-hickey: Hmm, the namespace isn't available outside the method. `Options` struct contains the optional ns to watch, but this could be `""` which would be weird and also possible different from `GetConfig()` which uses the kubeconfig on disk to get the NS.

I guess to guarantee it's the same, I was opting for loading it again in `AttachWatches`. I'm hesitant to change the `GetConfig()` signature to return the ns it's checking in as it's used in a few places.

cppforlife: how about using https://github.com/vmware-tanzu/carvel-kapp-controller/blob/2833f23be56ee9c028d5638bca94bf06f5aabec3/pkg/apiserver/apiserver.go#L44. seems like it's set in kc deployment "dynamically".

Dockerfile
```diff
@@ -65,8 +65,9 @@ RUN tdnf install -y git openssh-clients sed
 
 # Create the kapp-controller user in the root group, the home directory will be mounted as a volume
 RUN echo "kapp-controller:x:1000:0:/home/kapp-controller:/usr/sbin/nologin" > /etc/passwd
-# Give the root group write access to openssh's root bundle so we can append custom roots at runtime
-RUN chmod g+w /etc/pki/tls/certs/ca-bundle.crt
+# Give the root group write access to the openssh's root bundle directory
+# so we can rename the certs file with our dynamic config, and append custom roots at runtime
+RUN chmod g+w /etc/pki/tls/certs
```
neil-hickey, line 64: This was changed to allow os.Rename()

- `w` permission is needed on the entire directory to create files / rename files within
- This is something we will need to change in TCE's overlay files when we release or else this breaks!

PR #593, opened 2022-03-30
Make minor changes to kctrl for integration with tanzu-framework

praveenrewar:
*kctrl*
- Use binary name to configure namespace envVar key
- Use binary name to configure kubeconfig, kubeconfig context and kubeconfig yaml envVar key
- Use status as column name instead of description in pkgr get and pkgi get to show the current reconciliation status of repository or package installation.
- Return an error in package available get if package is not present
- Add --default-values-file-output-flag to generate default values file for an available package version.

Fixes #

```release-note
package available get now has a --default-values-file-output flag to generate default-values file for an available package version (with all the values commented).
```

##### Review Checklist:

  a link to that PR
  change

```docs

```


cli/pkg/kctrl/cmd/package/available/get.go
```diff
@@ -65,6 +68,7 @@ func NewGetCmd(o *GetOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Command
 	}
 
 	cmd.Flags().BoolVar(&o.ValuesSchema, "values-schema", false, "Values schema of the package (optional)")
+	cmd.Flags().BoolVar(&o.GenerateDefaultValues, "generate-default-values-file", false, "Generate default values from schema of the package (optional)")
```
cppforlife: we have a convention that flags that end with -output output something, so for this case, `default-values-file-output file.yml` would be better.

cli/pkg/kctrl/cmd/package/available/generate-default-values.go
```diff
@@ -0,0 +1,136 @@
...
+	"gopkg.in/yaml.v3"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	structuralschema "k8s.io/apiextensions-apiserver/pkg/apiserver/schema"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const Object = "object"
+
+// SchemaDefault returns a yaml byte array with values populated according to schema
+func SchemaDefault(schema []byte) ([]byte, error) {
```
cppforlife: lets turn this into class with all of these methods private, except one.

cli/pkg/kctrl/cmd/package/available/generate-default-values.go
```diff
@@ -0,0 +1,136 @@
+// Copyright 2020 VMware, Inc.
```
cppforlife: file names do not have dashes.

cli/pkg/kctrl/cmd/package/available/generate-default-values.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return nil, err
+	}
+	unstructured := make(map[string]interface{})
+
+	schemaDefault(unstructured, s)
+	var b bytes.Buffer
+	yamlEncoder := yaml.NewEncoder(&b)
+	yamlEncoder.SetIndent(2)
+	if err := yamlEncoder.Encode(unstructured); err != nil {
```
cppforlife: challenge with us generating "default values" file is that if user just takes that yaml and uses it as values.yml, then user effectively says "override everything" but in general user just needs to override just a few values. i believe @jorgemoralespou proposed an idea that we should generate a file that has all the fields commented. so that user can just uncomment fields and set values to what they want. additionally it would be nice if we can include descriptions on top of values and indicate what type.

cli/pkg/kctrl/cmd/package/available/generate-default-values.go
praveenrewar: Hmm, I thought I was using underscore. Good catch 👍🏻

cli/pkg/kctrl/cmd/package/available/generate-default-values.go
praveenrewar: > we should generate a file that has all the fields commented. so that user can just uncomment fields and set values to what they want.

I really like the idea. 

> additionally it would be nice if we can include descriptions on top of values and indicate what type.

That would be great. Let me see if I can add that as part of the same PR (the title minor changes doesn't feel suitable now 😁 )

cli/pkg/kctrl/cmd/package/available/get.go
praveenrewar: Just to make sure that I have understood this correctly, we want to update the flag to `default-values-file-output` and it should accept a file name where the values should be saved, right?
I have been thinking about the latter part as well (user should have the flexibility to choose file name).

cli/pkg/kctrl/cmd/package/repository/list.go
```diff
@@ -113,7 +113,8 @@ func NewSourceValue(pkgr v1alpha1.PackageRepository) uitable.Value {
 		switch {
 		case pkgr.Spec.Fetch.ImgpkgBundle != nil:
 			source = "(imgpkg) " + pkgr.Spec.Fetch.ImgpkgBundle.Image
-			if pkgr.Spec.Fetch.ImgpkgBundle.TagSelection != nil && pkgr.Spec.Fetch.ImgpkgBundle.TagSelection.Semver != nil {
+			if pkgr.Spec.Fetch.ImgpkgBundle.TagSelection != nil && pkgr.Spec.Fetch.ImgpkgBundle.TagSelection.Semver != nil &&
+				pkgr.Spec.Fetch.ImgpkgBundle.TagSelection.Semver.Constraints != "" {
```
praveenrewar, line 117: Does this make sense or should we have a default value (something like v0.0.0+) if constraint is empty?

cppforlife: package repository logic in kapp-controller is responsible for picking version if it's not provided by a human.

cli/pkg/kctrl/cmd/package/available/get.go
```diff
@@ -269,3 +287,25 @@ func (o *GetOptions) showValuesSchema(client pkgclient.Interface, pkgName, pkgVe
...
+		return nil
+	}
+
+	s := PackageSchema{pkg.Spec.ValuesSchema.OpenAPIv3.Raw}
+	defaultValues, err := s.DefaultValues()
+	if err != nil {
+		return err
+	}
+
+	os.WriteFile(o.DefaultValuesFile, defaultValues, 0600)
```
cppforlife: no err check

praveenrewar: Updated.

PR #608, opened 2022-04-12
API server startup does not timeout

### LINKED ISSUE #607: kapp-controller shouldn't exit if the packaging API is unavailable
**What steps did you take:**
If you have an issue like #479 where the Kubernetes API server can't reach kapp-controller's aggregated apiservice, kapp-controller will go into a crash loop, making the troubleshooting/diagnostics process even harder.

I think we should probably just loop forever instead of 60 times as we currently do, and log a message about the broken state we're in.

benmoss:
This prevents the confusing situation of kapp-controller crashing because the k8s apiserver can't reach us. kapp-controller has a dependency on its own apiserver for the packaging API reconcilers, so if the apiserver fails to come up we will be stuck in a degraded state.

We can't use readiness probes or anything like that to signal this degraded state since Kubernetes Services wont route traffic to pods until they are ready.

Fixes #607

```release-note
The aggregated API server will now wait forever if it is not registered with the Kubernetes API server instead of exiting after 60 seconds.
```


pkg/apiserver/apiserver.go
```diff
@@ -119,23 +120,17 @@ func NewAPIServer(clientConfig *rest.Config, coreClient kubernetes.Interface, kc
...
+	}()
 
-		if ready {
-			return nil
-		}
-		time.Sleep(1 * time.Second)
-	}
-	return fmt.Errorf("timed out after %s waiting for api server to become healthy. Check the status by running `kubectl get apiservices v1alpha1.data.packaging.carvel.dev -o yaml`", retries*time.Second)
+	return wait.PollInfinite(time.Second, func() (bool, error) {
+		klog.Info("waiting for API service to become ready")
```
cppforlife: would be nice to to add: "(hint: check the status via `kubectl get apiservices v1alpha1.data.packaging.carvel.dev -o yaml`)"

pkg/apiserver/apiserver.go
```diff
@@ -119,23 +120,17 @@ func NewAPIServer(clientConfig *rest.Config, coreClient kubernetes.Interface, kc
...
-	)
-	go as.server.PrepareRun().Run(as.stopCh)
-
-	for i := 0; i < retries; i++ {
-		ready, err := as.isReady()
+	go func() {
+		err := as.server.PrepareRun().Run(as.stopCh)
 		if err != nil {
-			return fmt.Errorf("checking readiness: %v", err)
+			klog.Warningf("API service stopped: %v", err)
```
cppforlife: i think we should pass in log `logr.Logger` into NewAPIServer (similar to other places).

benmoss: yeah, i started to do this and then got annoyed at needing to change the signature of `newServerConfig` and the private methods it calls where we use klog as well. Probably ought to do it though.

cppforlife: just need it on *APIServer, right?

PR #624, opened 2022-04-14
Add 'yanked' field to PackageCR

### LINKED ISSUE #201: Taint old/faulty Packages from being installed
I need a way to reject attempts to install an older version of a package because it has a critical bug, and there is a newer version of the package available w/the fix. 
When I have a package that has a bug I will probably create a new version that fixes it, and I want to prevent any new consumer of the package to install the faulty version.

I would like a way to define in a Package definition that the package is obsolete/faulty so that it would show in the output of the CLI. Alternatively I would desire that kapp-controller rejects installing Packages marked as faulty.

neil-hickey:
Add new status condition `PackageYanked` to indicate a package is marked as 'yanked'. This allows package authors to mark a package as being unfit for production - it contains a CVE / critical bug.

https://adamj.eu/tech/2021/09/20/how-to-fix-pip-yanked-version-warnings/

Fixes #201

```release-note
Add new field to `Package` - `spec.yanked` where users can mark a package as yanked by giving a `Reason` for the package yank
```

##### Review Checklist:

  a link to that PR
  change

```docs

```


pkg/apiserver/apis/datapackaging/types.go
```diff
@@ -139,3 +144,8 @@ type IncludedSoftware struct {
 	Version     string `json:"version,omitempty" protobuf:"bytes,2,opt,name=version"`
 	Description string `json:"description,omitempty" protobuf:"bytes,3,opt,name=description"`
 }
+
+// Revoked contains the reason for the package revokation
+type Revoked struct {
+	Reason string `json:"reason,omitempty" protobuf:"bytes,1,opt,name=reason"`
```
cppforlife: i think we should also have a date field to record when it got revoked.

pkg/apiserver/apis/datapackaging/types.go
```diff
@@ -96,6 +96,11 @@ type PackageSpec struct {
 	// This is especially useful if the underlying versions do not match the Package version
 	// +optional
 	IncludedSoftware []IncludedSoftware `json:"includedSoftware,omitempty"`
+
+	// Revoked indicates that this package is no longer recommended for use
+	// This package likely contains unsupported software, a CVE or a critical bug.
+	// +optional
+	Revoked Revoked `json:"revoked,omitempty"`
```
cppforlife: lets make this a pointer (*Revoked) since we should be able to revoke without specifying a reason.

pkg/reconciler/status.go
```diff
@@ -64,6 +64,16 @@ func (s *Status) IsDeleteFailed() bool {
 	return false
 }
 
+// IsPackageRevoked checks the status.conditions for the presence of the PackageRevoked condition
+func (s *Status) IsPackageRevoked() bool {
+	for _, cond := range s.S.Conditions {
+		if cond.Type == kcv1alpha1.PackageRevoked {
```
cppforlife: should we check for status==ConditionTrue?

config/crds.yml
```diff
@@ -124,6 +124,12 @@ spec:
                 format: date-time
                 nullable: true
                 type: string
+              revoked:
+                description: Revoked indicates that this package is no longer recommended for use This package likely contains unsupported software, a CVE or a critical bug.
```
joe-kimmel-vmw: nit: looks like you're missing a period between "use. This"

pkg/apiserver/apis/datapackaging/types.go
```diff
@@ -96,6 +96,11 @@ type PackageSpec struct {
 	// This is especially useful if the underlying versions do not match the Package version
 	// +optional
 	IncludedSoftware []IncludedSoftware `json:"includedSoftware,omitempty"`
+
+	// Revoked indicates that this package is no longer recommended for use
```
joe-kimmel-vmw: oh i see, the period would go at the end of this line. :)

pkg/apiserver/apis/datapackaging/v1alpha1/types.go
```diff
@@ -79,6 +79,11 @@ type PackageSpec struct {
 	// This is especially useful if the underlying versions do not match the Package version
 	// +optional
 	IncludedSoftware []IncludedSoftware `json:"includedSoftware,omitempty" protobuf:"bytes,9,opt,name=includedSoftware"`
+
+	// Revoked indicates that this package is no longer recommended for use
```
joe-kimmel-vmw: and/or, for reasons none of us quite gets yet, at the end of this line.

pkg/packageinstall/packageinstall.go
```diff
@@ -97,6 +97,10 @@ func (pi *PackageInstallCR) reconcile(modelStatus *reconciler.Status) (reconcile
 		return reconcile.Result{Requeue: true}, err
 	}
 
+	if pkg.Spec.Revoked.Reason != "" {
```
cppforlife: assuming revoked would be changed to be a pointer, we can do a nil check here.

cppforlife: i think your else case it way down at the bottom of this function. i think it would be better to keep it here: 

else {
   modelStatus.UnsetPackageRevoked(pi.model.ObjectMeta)
}

pkg/reconciler/status.go
cppforlife: nm, we dont do this for any other one (above).

pkg/reconciler/status.go
```diff
@@ -116,6 +126,24 @@ func (s *Status) SetReconcileCompleted(err error) {
...
+		Type:   kcv1alpha1.PackageRevoked,
+		Status: corev1.ConditionTrue,
+		Reason: reason,
+	})
+
+	s.UpdateFunc(s.S)
+}
+
+// UnsetPackageRevoked removes the PackageRevoked status from the conditions
+func (s *Status) UnsetPackageRevoked(meta metav1.ObjectMeta) {
```
cppforlife: seems like meta is not necessary here.

pkg/apiserver/apis/datapackaging/types.go
joe-kimmel-vmw: i'll argue vaguely in favor of boolean + string.
first, because i secretly like python, i'll quote PEP 20 "Explicit is better than implicit."  - booleans are explicit, but in the code below you're inferring a boolean from whether or not a string is empty.   when/why could that be problematic?  Well what if i put something in the string like,  "oops we actually meant to revoke a different version, this one's fine, sorry" .    anyway it's a corner case and this is a weakly held opinion,  but if there was no strong argument against a boolean i'd make an explicit flag and let the string be supplemental metadata.

pkg/apiserver/apis/datapackaging/types.go
joe-kimmel-vmw: the description of the CRD above winds up missing a period between "use. This" because the generator doesn't insert punctuation on newlines i think.   TBH it's a mildly interesting experiment to me whether you need to add the period here or in the other file where this is duplicated.

pkg/packageinstall/packageinstall.go
joe-kimmel-vmw: again just vaguely arguing that `if pkg.Spec.Revoked.Revoked` is vaguely easier to read than `if ... != ""`

pkg/packageinstall/packageinstall_test.go
```diff
@@ -454,3 +454,100 @@ func Test_PlaceHolderSecretCreated_WhenPackageInstallUpdated(t *testing.T) {
 
 	assert.Equal(t, "instl-pkg-fetch-0", app.Spec.Fetch[0].ImgpkgBundle.SecretRef.Name)
 }
+
+func Test_CreatePackage_RevokedStatus(t *testing.T) {
```
joe-kimmel-vmw: I was expecting to see a test that shows that if you have versions 1.0.0 and 1.1.0 and you do the version selection logic, it returns 1.1.0.    but then if you revoke 1.1.0 and run the same version selection function again,  it returns 1.0.0. 

i think similarly you could cover the "don't upgrade to a revoked version" by testing that same version selection function and showing that if you then add 1.2.0  package it'll return that one (skipping the revoked package).

joe-kimmel-vmw: I was expecting to see a test that shows that if you have versions 1.0.0 and 1.1.0 and you do the version selection logic, it returns 1.1.0.    but then if you revoke 1.1.0 and run the same version selection function again,  it returns 1.0.0. 

i think similarly you could cover the "don't upgrade to a revoked version" by testing that same version selection function and showing that if you then add 1.2.0  package it'll return that one (skipping the revoked package).

joe-kimmel-vmw: I was expecting to see a test that shows that if you have versions 1.0.0 and 1.1.0 and you do the version selection logic, it returns 1.1.0.    but then if you revoke 1.1.0 and run the same version selection function again,  it returns 1.0.0. 

i think similarly you could cover the "don't upgrade to a revoked version" by testing that same version selection function and showing that if you then add 1.2.0  package it'll return that one (skipping the revoked package).

joe-kimmel-vmw: I was expecting to see a test that shows that if you have versions 1.0.0 and 1.1.0 and you do the version selection logic, it returns 1.1.0.    but then if you revoke 1.1.0 and run the same version selection function again,  it returns 1.0.0. 

i think similarly you could cover the "don't upgrade to a revoked version" by testing that same version selection function and showing that if you then add 1.2.0  package it'll return that one (skipping the revoked package).

neil-hickey: This PR doesn't cover any upgrading patterns, it just allows and then shows the revoked status. The next iteration is the one where this fun comes into play. If I read your comment correctly!

pkg/apiserver/apis/datapackaging/types.go
cppforlife: in k8s and carvel apis typically to enable you use presence of the object:

without reason
```
revoked: {}
```
with reason
```
revoked:
  reason: "foo"
```

i left a comment about making Revoked a pointer.

jorgemoralespou:
I see `revoked` as a confusing term (maybe because I'm no english native speaker).

neil-hickey:
> I see `revoked` as a confusing term (maybe because I'm no english native speaker).

Hey @jorgemoralespou ! I am open to suggestions on the naming. We initially veered away from something like `deprecated` or `unsupported`. Naming is hard for sure :)

Here's a link to the original team chat around the right word to use to represent this concept. [link](https://github.com/k14s/upgrade-dependency-management/pull/5/commits/d4babf68aea9995b5057f9b6705ac7f8c43507d8#r743231294)

jorgemoralespou:
> Here's a [link](https://github.com/k14s/upgrade-dependency-management/pull/5/commits/d4babf68aea9995b5057f9b6705ac7f8c43507d8#r743231294) to the original team chat around the right word to use to represent this concept. link

That seems to be a private conversation. In any case, not sure why deprecated, forbidden, unsupported, faulty, broken, not-recommended are not better options. To me, revoke means something that was granted to me but is taken away. I understand that this is not the intent here, but rather something that should not be installed.

neil-hickey:
Here's a summary, wasn't aware that was a private repo / comment thread.

- Author: cppforlife (dmitriy)
> any other alternative terms to deprecated. coming from api consumption land, deprecate does not typically mean "wont install/upgrade", but rather be aware that it will be removed.
 
- Author: neil-hickey
> Some potential words:
discontinued, unsupported, unstable, removed, revoked

- Author: cppforlife
> removed/revoked to me aligns more if you want to "skip" installing them for new installs. e.g. v5 (revoked) v4 v3
new users asking for latest version will only get v4. existing users who already have v5 installed will continue to be on v5.
 
- Author: joe-kimmel-vmw
> obsoleted, abandoned, orphaned, disallowed, blocked, undesirable, do-not-install, rated-0-on-rotten-tomatoes, 0-out-of-10-stars, ...
 
- Author: neil-hickey
> I am liking revoked, It semantically aligns with our idea of "this was explicitly marked as undesired by the author" and also "I should skip over something that was revoked". If no-one objects, I will re-word the PR to reflect "revoked". I think it also conveys that, "I am on a revoked version, that's probably not good and I should move off of it"

cppforlife:
i propose we call it `yanked` -- it's a term ive seen in ruby, python and other communities -- https://adamj.eu/tech/2021/09/20/how-to-fix-pip-yanked-version-warnings/

jorgemoralespou:
Revoked was not intuitive, yanked is even less, but I guess you have done good UX research, so won't complain anymore.

GrahamDumpleton:
In PyPi labelling it as yanked means the installer will not install it if using floating version, but will still install if is pinned to specific version. Will kapp-controller behave the exact same way if going to use same naming?

GrahamDumpleton:
Can you also explain how this would work in practice?

In the case of PyPi, because it is a centralised package repository (ignoring mirrors), then when a package is marked as yanked, the effect of that is immediate for everyone. This is in part possible because the metadata which holds that the package was yanked is in the index data of the package repository and the package itself doesn't need to change or a new build released.

What this PR really does I can't really tell since `InternalPackage` custom resource isn't documented in `kapp-controller` docs, so have no idea how that relates to actual `Package` definitions and package repositories.

Is this yanked flag something that can only be applied locally in a specific Kubernetes cluster and isn't actually a way of being able to flag upstream in a package repository that a package should no longer be used so everyone knows? Thus, a cluster admin would have to manually themselves work out that an upstream package shouldn't be used and somehow maintain this flag themselves? Confused....

neil-hickey:
> In PyPi labelling it as yanked means the installer will not install it if using floating version, but will still install if is pinned to specific version. Will kapp-controller behave the exact same way if going to use same naming?

We haven't quite ironed out the details of the system limitations based on this flag.

> Is this yanked flag something that can only be applied locally in a specific Kubernetes cluster and isn't actually a way of being able to flag upstream in a package repository that a package should no longer be used so everyone knows?

The idea behind this change is: 

- I am a package author and I want to indicate that my package is no longer supported / contains a bug. I regardless of whether my package is distributed by a PackageRepo or not.
- In the case where my Package is brought into a cluster via a PackageRepo, all consumers will get this notice. 

Hopefully this clears up some of the confusion. This PR is a first step, indicating via metadata that this Package is perhaps unwanted / not good. Subsequently we could have the system make decisions based on this. But for now, this would just  be a warning that shows up in the PackageInstall status.

pkg/apiserver/apis/datapackaging/types.go
```diff
@@ -139,3 +144,9 @@ type IncludedSoftware struct {
 	Version     string `json:"version,omitempty" protobuf:"bytes,2,opt,name=version"`
 	Description string `json:"description,omitempty" protobuf:"bytes,3,opt,name=description"`
 }
+
+// Yanked contains details of the rationale of why this version was yanked
+type Yanked struct {
+	Reason string `json:"reason,omitempty" protobuf:"bytes,1,opt,name=reason"`
+	Date   string `json:"date,omitempty" protobuf:"bytes,2,opt,name=date"`
```
cppforlife: i think we should use something like this (date+time), not sure about the name though, may be just `At`. it might be important for a user to understand when within the day package was yanked (useful when trying to understand production impact).

> ReleasedAt metav1.Time `json:"releasedAt,omitempty"`

pkg/reconciler/status.go
```diff
@@ -116,9 +122,21 @@ func (s *Status) SetReconcileCompleted(err error) {
 	s.UpdateFunc(s.S)
 }
 
-func (s *Status) SetDeleteCompleted(err error) {
-	s.removeAllConditions()
+// SetPackageYanked appends a new condition to indicate a package is yanked
+func (s *Status) SetPackageYanked(meta metav1.ObjectMeta, reason string) {
+	s.markObservedLatest(meta)
```
cppforlife: dont think we should mark latest generation here since this is not part of primary reconciliation logic (e.g. Reconcile*)

pkg/reconciler/status.go
```diff
@@ -116,9 +122,21 @@ func (s *Status) SetReconcileCompleted(err error) {
...
+
+	s.S.Conditions = append(s.S.Conditions, kcv1alpha1.Condition{
+		Type:   kcv1alpha1.PackageYanked,
+		Status: corev1.ConditionTrue,
+		Reason: reason,
+	})
 
+	s.UpdateFunc(s.S)
```
cppforlife: i would also try to avoid making an additional api request for non-terminal conditions

neil-hickey: I think we have to make the api request to update the status of the PackageInstall right? I'm a little confused by how we can get it to update without the api call

neil-hickey: @cppforlife

neil-hickey:
superceded by https://github.com/vmware-tanzu/carvel-kapp-controller/pull/936

PR #629, opened 2022-04-16
pkgr benchmark test

joe-kimmel-vmw:

- introduces benchmarking test suite
- introduces [benchmark action ](https://github.com/vmware-tanzu/carvel-kapp-controller/runs/6056404533?check_suite_focus=true) 
  - to run on changes to develop branch or manual trigger.
  - action has the ability to alert if we increase the benchmarked metrics by X (200% by default)

*open questions* : 
- the github action has the ability to write a[ timeseries graph of the metrics to github pages](https://github.com/benchmark-action/github-action-benchmark#charts-on-github-pages-1) which imo is the best reason for using it,  but are we cool with enabling github pages and letting it do its thing?
- the benchmarks default failure is if a single run drifts by 200% -- imo we should tighten that to more like 150% or even 130% but it might take a couple failures to learn how much variability we get naturally - any opinions?

This is related to the "layering" tract of work - I thought it'd be cool to introduce a benchmark suite and then we can see the impact of other work on the benchmarks as they get merged.

### DIFF
--- a/.github/workflows/benchmark-action.yaml
+++ b/.github/workflows/benchmark-action.yaml
@@ -4,7 +4,16 @@ on:
   push:
     branches:
       - develop
+  pull_request:
+    types: [opened, reopened, synchronize]
   workflow_dispatch:
+
+permissions:
+  # deployments permission to deploy GitHub pages website
+  deployments: write
+  # contents permission to update benchmark contents in gh-pages branch
+  contents: write
+
 
 # see https://github.com/benchmark-action/github-action-benchmark
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/benchmark-action.yaml
```diff
@@ -0,0 +1,76 @@
...
+          go test ./test/bench/... -bench=. | tee output.txt
+      # Download previous benchmark result from cache (if exists)
+      - name: Download previous benchmark data
+        uses: actions/cache@v1
+        with:
+          path: ./cache
+          key: ${{ runner.os }}-benchmark
+      # Run `github-action-benchmark` action
+      - name: Store benchmark result
+        uses: benchmark-action/github-action-benchmark@v1
```
cppforlife: since this is an external action (and it writes to our repo), i would recommend locking it down by sha.

joe-kimmel-vmw: just a note that to test this and ensure it's working i commented out the branch conditional so it'll need another commit/push before its ready to merge.  i'll ping and resolve this conversation when it's done.

joe-kimmel-vmw: confirmed I pinned to hashes correctly and restored the conditionals.

PR #635, opened 2022-04-20
apiserver: custom QPS and Burst to allow high throughput of packages

joe-kimmel-vmw:

Benchmarks on current develop (sidescroll to see all metrics):
```
Benchmark_pkgr_with_500_packages-2   	       1	94443823631 ns/op	        63.52 DeleteSeconds	        30.88 DeploySeconds
Benchmark_pkgr_with_100_packages-2   	       1	23855612927 ns/op	        15.56 DeleteSeconds	         8.249 DeploySeconds
Benchmark_pkgr_with_50_packages-2    	       1	14642181389 ns/op	         9.423 DeleteSeconds	         5.172 DeploySeconds
```

configuring high QPS and Burst only for packages gives us:
Benchmark run on this branch:
```
Benchmark_pkgr_with_500_packages-2   	       1	36399893748 ns/op	        18.88 DeleteSeconds	        17.48 DeploySeconds
Benchmark_pkgr_with_100_packages-2   	       1	9438283825 ns/op	         5.262 DeleteSeconds	         4.143 DeploySeconds
Benchmark_pkgr_with_50_packages-2    	       1	6437118537 ns/op	         4.277 DeleteSeconds	         2.126 DeploySeconds
```

Note that we also previously tested this change on GKE.  Ran with slightly different PKGR sizes, but similar results:

kapp controller 34 on GKE:

- pkgr w. 200 packages:  15 DeploySeconds   34 DeleteSeconds
- pkgr w. 1000 packages: 65 DeploySeconds.  132 DeleteSeconds

kc 0.34 plus these changes, on GKE: 

- pkgr w. 200 packages:  10 DeploySeconds   15 DeleteSeconds
- pkgr w. 1000 packages: 38 DeploySeconds.  43 DeleteSeconds

Fixes #

```release-note
PackageRepositories, especially those with hundreds of packages, now reconcile faster
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/cmd/controller/run.go
+++ b/cmd/controller/run.go
@@ -100,7 +100,15 @@ func Run(opts Options, runLog logr.Logger) error {
 		return fmt.Errorf("Expected to find %s env var", kappctrlAPIPORTEnvKey)
 	}
 
-	server, err := apiserver.NewAPIServer(restConfig, coreClient, kcClient, apiserver.NewAPIServerOpts{
+	// to facilitate creation of many packages at once from a larger PKGR
+	pkgRestConfig := config.GetConfigOrDie()
+	pkgRestConfig.QPS = 100
+	pkgRestConfig.Burst = 100
+	pkgKcClient, err := kcclient.NewForConfig(pkgRestConfig)
+	if err != nil {
+		return fmt.Errorf("Building pkg kappctrl client: %s", err)
+	}
+	server, err := apiserver.NewAPIServer(pkgRestConfig, coreClient, pkgKcClient, apiserver.NewAPIServerOpts{
 		GlobalNamespace:              opts.PackagingGloablNS,
 		BindPort:                     bindPort,
 		EnableAPIPriorityAndFairness: opts.APIPriorityAndFairness,
--- a/cmd/controller/run.go
+++ b/cmd/controller/run.go
@@ -102,8 +102,8 @@ func Run(opts Options, runLog logr.Logger) error {
 
 	// to facilitate creation of many packages at once from a larger PKGR
 	pkgRestConfig := config.GetConfigOrDie()
-	pkgRestConfig.QPS = 100
-	pkgRestConfig.Burst = 100
+	pkgRestConfig.QPS = 60
+	pkgRestConfig.Burst = 90
 	pkgKcClient, err := kcclient.NewForConfig(pkgRestConfig)
 	if err != nil {
 		return fmt.Errorf("Building pkg kappctrl client: %s", err)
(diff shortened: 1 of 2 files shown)


cmd/controller/run.go
```diff
@@ -100,7 +100,15 @@ func Run(opts Options, runLog logr.Logger) error {
 		return fmt.Errorf("Expected to find %s env var", kappctrlAPIPORTEnvKey)
 	}
 
-	server, err := apiserver.NewAPIServer(restConfig, coreClient, kcClient, apiserver.NewAPIServerOpts{
+	// to facilitate creation of many packages at once from a larger PKGR
+	pkgRestConfig := config.GetConfigOrDie()
+	pkgRestConfig.QPS = 100
+	pkgRestConfig.Burst = 100
```
neil-hickey: Why 100?

joe-kimmel-vmw: it's definitely a little arbitrary -   default values are`QPS:  20`  and `Burst:  30`. 

some longer notes below but TLDR 100/100 was a pretty fair guess --   I'm going to update the values to 3x multipliers (60/90) as that seems to be the smallest values that achieve the same performance as 100/100: 

  In some very informal experiments I also tried just a 2x multiplier on the defaults and found that was slower: QPS 40, Burst 60: pkgr with 500 packages had took 28 seconds to Deploy, as opposed to ~19 seconds above.

  Then I tried setting both values to 200, and at first I got a reconcile failure for the PKGR with 500 packages.  When I tried to figure out what went wrong I couldn't reproduce the error,  but I ran it twice and got Deploy/Delete metrics that were no faster than with those values at 100.   This is kind of reassuring since it suggests there really is some other throttling that keeps us from totally destroying the cluster, though maybe at the expense of failing our own reconciles. 
  
  So finally i tried setting both values to 3x multipliers (QPS 60 and Burst 90) and I got basically the same results as posted above:
  ```
  Benchmark_pkgr_with_500_packages-16    	       1	37331688769 ns/op	        19.52 DeleteSeconds	        17.73 DeploySeconds
Benchmark_pkgr_with_100_packages-16    	       1	10132445350 ns/op	         5.756 DeleteSeconds	         4.291 DeploySeconds
Benchmark_pkgr_with_50_packages-16     	       1	6981010533 ns/op	         4.637 DeleteSeconds	         2.261 DeploySeconds
```

that's pretty good , right ?  but is that QPS setting doing anything at all? 
QPS back to 20 with burst still at 90 took 50 seconds to deploy 500 packages,  so the answer is yes!

PR #637, opened 2022-04-21
`kctrl`: Adding commands kick, status and pause to `kctrl package installed` command tree

100mik:
Also includes some minor formatting/output changes for the app commands!


cli/pkg/kctrl/cmd/package/installed/status.go
```diff
@@ -0,0 +1,79 @@
...
+		return err
+	}
+
+	app, err := client.KappctrlV1alpha1().Apps(o.NamespaceFlags.Name).Get(context.Background(), o.Name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	if app.OwnerReferences[0].Kind != "PackageInstall" || app.OwnerReferences[0].Name != o.Name {
+		return fmt.Errorf("Could not find app associated with app '%s' in namespace '%s'", o.Name, o.NamespaceFlags.Name)
```
praveenrewar:
```suggestion
		return fmt.Errorf("Could not find app associated with package install '%s' in namespace '%s'", o.Name, o.NamespaceFlags.Name)
```

cli/pkg/kctrl/cmd/package/installed/pause_or_kick.go
```diff
@@ -0,0 +1,258 @@
...
+			o.statusUI.PrintMessagef("Error tailing or reconciling app: %s", err.Error())
+			*tailErrored = true
+		}
+	}
+	go tailAppStatusOutput(&appStatusTailErrored)
+
+	if err := wait.Poll(o.WaitFlags.CheckInterval, o.WaitFlags.Timeout, func() (done bool, err error) {
+
+		resource, err := client.PackagingV1alpha1().PackageInstalls(o.NamespaceFlags.Name).Get(context.Background(), o.Name, metav1.GetOptions{})
+		//resource, err := p.kappClient.GetPackageInstall(name, namespace)
```
praveenrewar: nit: Remove commented code

cli/pkg/kctrl/cmd/package/installed/pause_or_kick.go
```diff
@@ -0,0 +1,258 @@
...
+			IgnoreNotExists: true,
+		})
+
+		err := appWatcher.TailAppStatus()
+		if err != nil {
+			o.statusUI.PrintMessagef("Error tailing or reconciling app: %s", err.Error())
+			*tailErrored = true
+		}
+	}
+	go tailAppStatusOutput(&appStatusTailErrored)
```
praveenrewar, line 225: *question:* I am just trying to think why we need to do this in parallel. Maybe we could call TailAppStatus() first and if it errors out then do the polling like we used to, otherwise return from there itself?

100mik: That means we stop execution till TailAppStatus exits. 
Polling in the background means we are ensuring that the package install isn't erroring out at the same time. 

One case i recall where I recall this was useful is when a wrong version is supplied

praveenrewar: I haven't taken a look at it, but I hope the TailAppStatus will error in such scenarios, the only scenario where we should go ahead with the polling is when app is not found (maybe due to permissions), so maybe one option would be to check for IsNotFound error and then continue or return based on that.

praveenrewar: > the only scenario where we should go ahead with the polling is when app is not found

I could be wrong here. Need to think this throughly.

praveenrewar: Resolving it based on slack discussion:
TailAppStatus doesn't error out in all the scenarios.

cli/pkg/kctrl/cmd/package/installed/status.go
100mik: Adopted

cli/pkg/kctrl/cmd/package/installed/pause_or_kick.go
100mik: Resolved!

PR #651, opened 2022-04-27
Build / download deps in golang image

### LINKED ISSUE #644: Update Golang version to latest
Hello! 

I'm wondering if folks are already planning to update consumed golang version to latest which could fix a couple of CVEs. Opening this issue just to track the golang version update story in `kapp-controller` and also its affiliated tools (`k14s kapp`, `carvel vendir`, `helm`).

Thanks!

benmoss:
Switches our build step to use [library/golang](https://hub.docker.com/_/golang) instead of Photon.
 
Fixes #644

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,25 +1,7 @@
-FROM photon:4.0
+FROM golang:1.17.9 AS deps
 
 ARG KCTRL_VER=development
-
-# adapted from golang docker image
-ENV PATH /usr/local/go/bin:$PATH
-ENV GOLANG_VERSION 1.17.6
-ENV GO_REL_ARCH linux-amd64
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -46,10 +30,6 @@ RUN curl -sLo cue.tgz https://github.com/cue-lang/cue/releases/download/v0.4.2/c
   tar -xf cue.tgz -C /usr/local/bin cue && cue version
 
 # kapp-controller
-COPY . .
-# helpful ldflags reference: https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications
-RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags="-X 'main.Version=$KCTRL_VER' -buildid=" -trimpath -o controller ./cmd/main.go
```
cppforlife: still want to keep this last to avoid redownloading binaries when only source changes.

Issue #428, opened 2021-11-19
Support layering of package repositories


danielhelfand:
**Describe the problem/challenge you have**

For a user of kapp-controller, the possibility exists to create two PackageRepositories which may contain a Package with the same name. It could be the case that these are the exact same Package, but it could also be that the names are the same but there are differences between the Packages.

Currently, due to kapp-controller's use of `kapp` as its method for creating Packages, kapp-controller will fail to create a Package with the same name as another in the same namespace. Since PackageRepositories are often created around software products that can be independent of one another (e.g. Tanzu Application Platform and Tanzu Community Edition), it makes sense for this duplication to occur where both products want to both make the same Package available (e.g. cert-manager). 

kapp-controller should support a way of dealing with these conflicts to provide a better user experience for package consumers working with multiple repositories. 

**Describe the solution you'd like**

There currently exist some workarounds for this issue:
* Coming up with a unique name for a conflicting Package (e.g. maybe including the product name in the package name)
* Keep same package name but use slightly different version. The challenge with this workaround is that a PackageMetadata resource can only be owned by one PackageRepository.

It's not immediately clear what the long term way to address this problem is, but some ideas that have been considered are:
* Having kapp-controller create Packages without using `kapp`. Maybe using the k8s client to develop a custom experience for kapp-controller around handling conflicts.
* Supporting a way to define precedence and which Package kapp-controller should prefer. An example of this is a PackageRepo for product 1 has a cert-manager Pacakge that should be preferred over PackageRepo for product 2's cert-manager Package. What this experience would look like and how a user could configure this have yet to be determined.
* Could kapp-controller more gracefully handle the case of conflicts instead of showing a PackageRepository as failing to reconcile in the case where a conflict exists.
* Is there anything that could be directly implemented in `kapp` to address this issue?

**Anything else you would like to add:**

Are there any additional cases to consider beyond two Packages conflicting within the same namespace?

One question to still be answered is if kapp-controller will still create all Packages except the conflicting Package in the case of a conflict.

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

joe-kimmel-vmw:
[...]

PR #657, opened 2022-04-28
rebase rule and e2e tests for pkgrs with identical pkgs

joe-kimmel-vmw:
https://github.com/vmware-tanzu/carvel-kapp-controller/issues/428


hack/dev-deploy.sh
```diff
@@ -21,5 +22,7 @@ tail -n +$run_image_start Dockerfile | \
    sed 's/usr\/local\/bin\///' \
    >> Dockerfile.dev
 
+echo "COPY kapp ." >> Dockerfile.dev
```
joe-kimmel-vmw: obv. will not be included in final version / just for dev.

pkg/pkgrepository/app_deploy.go
```diff
@@ -30,13 +30,105 @@ func (a *App) deploy(tplOutput string) exec.CmdRunResult {
 			return exec.NewCmdRunResultWithErr(fmt.Errorf("Preparing kapp: %s", err))
 		}
 
-		return kapp.Deploy(tplOutput, a.startFlushingAllStatusUpdates, func(exec.CmdRunResult) {})
+		// TODO: should we inject this somewhere else? - there's some other rules that might interact with it...
```
joe-kimmel-vmw: @cppforlife   i left this in the same spot as the prototype for now but curious what your thoughts are --  iirc there's a trick that it can't be inserted too early so it's slightly delicate?

pkg/pkgrepository/app_deploy.go
```diff
@@ -30,13 +30,105 @@ func (a *App) deploy(tplOutput string) exec.CmdRunResult {
...
+        #@   fail(msg)
+        #@ end
+
+        #@ def annotations_are_identical(existing_anns, new_anns):
+        #@   carvel_anns = {"kapp.k14s.io/disable-original": True,
+        #@                   "kapp.k14s.io/disable-wait": True,
+        #@                   "packaging.carvel.dev/package-repository-ref": True,
+        #@                   "kapp.k14s.io/identity": True,
+        #@                   "kapp.k14s.io/create-strategy": True}
+        #@   return {x: v for x,v in existing_anns.items() if x not in carvel_anns} == {x: v for x,v in new_anns.items() if x not in carvel_anns}
```
joe-kimmel-vmw: i think i'll write a `filter` func to reduce the `{x: v for x,v in .... }`

pkg/pkgrepository/app_deploy.go
```diff
@@ -30,13 +30,105 @@ func (a *App) deploy(tplOutput string) exec.CmdRunResult {
...
+
+        #@ def is_identical(existing, new):
+
+        #@   print( "\nrefnames? " + str(existing.spec.refName == new.spec.refName) + \
+        #@           "\nversion? " + str(existing.spec.version == new.spec.version) + \
+        #@           "\ntemplate? " + str(struct.decode(existing.spec.template) == struct.decode(new.spec.template)) + \
+        #@           "\nlabels? " + str(labels_are_identical(struct.decode(existing.metadata.labels), struct.decode(new.metadata.labels))) +\
+        #@           "\nannotations? " + str(annotations_are_identical(struct.decode(existing.metadata.annotations), struct.decode(new.metadata.annotations))))
+
+        #@ print ("\n existing template: \n" + str(struct.decode(existing.spec.template)) + "\n new template: " + str(struct.decode(new.spec.template)))
```
joe-kimmel-vmw: these are debug prints,  but imo might be helpful to thread a debug flag through to here so that when someone gets stuck they're able to ask "why" in a meaningful way --- any thoughts on how best to enable introspection / debugging @cppforlife  ?

pkg/pkgrepository/app_deploy.go
```diff
@@ -30,13 +30,105 @@ func (a *App) deploy(tplOutput string) exec.CmdRunResult {
...
+    overlayContractV1:
+      overlay.yml: |
+        #@ load("@ytt:data", "data")
+        #@ load("@ytt:json", "json")
+        #@ load("@ytt:overlay", "overlay")
+        #@ load("@ytt:struct", "struct")
+
+        #@ def get_rev(annotations):
+        #@  if hasattr(annotations, "packaging.carvel.dev/repo-rev"):
+        #@    return int(annotations["packaging.carvel.dev/repo-rev"])
```
cppforlife: the revision itself is not connected to a repo, but rather revision of a particular resource => `packaging.carvel.dev/revision`

pkg/pkgrepository/app_deploy.go
neil-hickey: yeh, some kind of filter would probably make it more readable also.

pkg/pkgrepository/app_deploy.go
```diff
@@ -30,13 +30,105 @@ func (a *App) deploy(tplOutput string) exec.CmdRunResult {
...
+- ytt:
+    overlayContractV1:
+      overlay.yml: |
+        #@ load("@ytt:data", "data")
+        #@ load("@ytt:json", "json")
+        #@ load("@ytt:overlay", "overlay")
+        #@ load("@ytt:struct", "struct")
+
+        #@ def get_rev(annotations):
+        #@  if hasattr(annotations, "packaging.carvel.dev/repo-rev"):
```
neil-hickey: What is the `rev` in `packaging.carvel.dev/repo-rev` stand for?

joe-kimmel-vmw: @neil-hickey good question,  especially as this is a place where the rebase rule is ahead of the tests -- 

it's a "revision counter" 

there's a related TODO in the list of test cases i haven't written yet:
https://github.com/vmware-tanzu/carvel-kapp-controller/pull/657/files#diff-3c7cc881b74c8228f1e705cf8b9ea60b146d078f33a94009a19ac70ea4c75fe3R374

but both are definitely in a shorthand only sensible to someone who was working on the prototype. :)

pkg/pkgrepository/app_deploy.go
cppforlife: add where we already add other stuff -- https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/pkg/pkgrepository/package_repo_app.go#L82

pkg/pkgrepository/app_deploy.go
```diff
@@ -30,13 +30,111 @@ func (a *App) deploy(tplOutput string) exec.CmdRunResult {
...
+            kapp.k14s.io/noop: ""
+        #@ elif get_rev(data.values.existing.metadata.annotations) < get_rev(data.values.new.metadata.annotations):
+        #@ print("replacing existing older rev with newer rev")
+        #@ else:
+        #@   msg = "Error: Conflicting Packages with name " + data.values.existing.spec.refName + "; Package is already present but not identical (mismatch in " + reason +")"
+        #@   print(msg)
+        #@   fail(msg)
+        #@ end
+  resourceMatchers:
+  - apiVersionKindMatcher: {apiVersion: data.packaging.carvel.dev/v1alpha1, kind: Package}
```
cppforlife: this should also be applying to PackageMetadata (try adding 2 tap repos for example)

pkg/pkgrepository/app_deploy.go
```diff
@@ -30,13 +30,111 @@ func (a *App) deploy(tplOutput string) exec.CmdRunResult {
...
+    overlayContractV1:
+      overlay.yml: |
+        #@ load("@ytt:data", "data")
+        #@ load("@ytt:yaml", "yaml")
+        #@ load("@ytt:json", "json")
+        #@ load("@ytt:overlay", "overlay")
+        #@ load("@ytt:struct", "struct")
+
+        #@ def get_rev(annotations):
+        #@  if hasattr(annotations, "packaging.carvel.dev/repo-pkg-revision"):
```
cppforlife: annotation naming should not be Package specific, this also applies to PackageMetadata.

pkg/pkgrepository/app_deploy.go
```diff
@@ -30,13 +30,111 @@ func (a *App) deploy(tplOutput string) exec.CmdRunResult {
...
+        #@ def is_identical(existing, new):
+        #@   if existing.spec.refName != new.spec.refName:
+        #@     return False, "spec.refName"
+        #@   end
+        #@   if existing.spec.version != new.spec.version:
+        #@     return False, "spec.version"
+        #@   end
+        #@   if struct.decode(existing.spec.template) != struct.decode(new.spec.template):
+        #@    return False, "spec.template"
+        #@   end
```
cppforlife: how come we arent checking entire spec section?

pkg/pkgrepository/app_deploy.go
```diff
@@ -30,13 +30,111 @@ func (a *App) deploy(tplOutput string) exec.CmdRunResult {
...
+        #@ elif get_rev(data.values.existing.metadata.annotations) > get_rev(data.values.new.metadata.annotations):
+        #@overlay/match by=overlay.all
+        ---
+        metadata:
+          #@overlay/match missing_ok=True
+          annotations:
+            #@overlay/match missing_ok=True
+            kapp.k14s.io/noop: ""
+        #@ elif get_rev(data.values.existing.metadata.annotations) < get_rev(data.values.new.metadata.annotations):
+        #@ print("replacing existing older rev with newer rev")
```
cppforlife: nit: indent by 2 spaces after `@`

cppforlife: are these print statements always showing up in kapp deploy output?

pkg/pkgrepository/app_deploy.go
cppforlife: nit: indention should be 2 spaces like in other places below

cppforlife: (im still favoring `packaging.carvel.dev/revision` as a name)

test/e2e/kappcontroller/package_repo_test.go
```diff
@@ -287,6 +288,414 @@ spec:
...
+		assertPkgOwnedBy(pkgr1Name, "shirt-mgr.co.uk.5.5.5")
+		assertPkgOwnedBy(pkgr2Name, "contooor.co.uk.0.22.0")
+	})
+
+	// pkgr3 should fail to reconcile because the contooor package will conflict in the spec
+	pkgr3 := fmt.Sprintf(
+		"%s%s%s",
+		fmt.Sprintf(pkgrPreamble, 3),
+		fmt.Sprintf(pkgTemplate, "shirt-mgr.co.uk", "5.5.5"),
+		strings.Replace(fmt.Sprintf(pkgTemplate, "contooor.co.uk", "0.22.0"), "k8slt/kctrl-example-pkg:v1.0.0", "k8slt/some-other-image:latest", -1))
```
cppforlife: string building throughout tests seem really brittle and quite hard to piece together. lets come up with a different way to adjust inputs.

test/e2e/kappcontroller/package_repo_test.go
```diff
@@ -287,6 +288,414 @@ spec:
...
+	cleanUp3 := func() {
+		kapp.Run([]string{"delete", "-a", appName3})
+	}
+	cleanUp3()
+	defer cleanUp3()
+
+	logger.Section("deploy and check pkgr3", func() {
+		_, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", appName3},
+			e2e.RunOpts{StdinReader: strings.NewReader(pkgr3), AllowError: true})
+		if err != nil {
```
cppforlife: shouldnt have conditionals in the test. assert that error is set, and then assert that error contains info.

test/e2e/kappcontroller/package_repo_test.go
```diff
@@ -287,6 +288,414 @@ spec:
...
+	}
+	cleanUp()
+	defer cleanUp()
+
+	assertPkgOwnedBy := func(pkgrName string) {
+		out, err := kubectl.RunWithOpts(
+			[]string{"get", "package", pkgName, "-oyaml"},
+			e2e.RunOpts{AllowError: true})
+
+		assert.NoError(t, err)
```
cppforlife: if we dont expect error, then do not set AllowError=true above.

test/e2e/kappcontroller/package_repo_test.go
```diff
@@ -287,6 +288,414 @@ spec:
 	}
 }
 
+func Test_PackageReposWithOverlappingPackages_identicalPackagesReconcile(t *testing.T) {
```
cppforlife, line 418: lets also add a test where same pkgr is added in global packaging namespace and in local one.

pkg/pkgrepository/app_deploy.go
joe-kimmel-vmw: to be honest I thought it was your idea.

test/e2e/kappcontroller/package_repo_test.go
joe-kimmel-vmw: I agree it's a little ugly &verbose -- the trouble is that if it does fail, you have to put this logic back in and run it again to find out why -- all you get is "see status.usefulerrormessage for details" , but once the test has terminated the useful error message is gone. 

should we add support into the e2e.runopts,  like `kubectlOnError: []string{"get", "package" , pkgName, "-oyaml"}` ?

cli/test/e2e/package_repository_test.go
```diff
@@ -53,9 +55,24 @@ func TestPackageRepository(t *testing.T) {
 	})
 
 	logger.Section("adding a repository", func() {
-		kappCtrl.RunWithOpts([]string{"package", "repository", "add", "-r", pkgrName, "--url", pkgrURL}, RunOpts{})
-
-		kubectl.Run([]string{"get", "packagerepository", pkgrName})
+		out, err := kappCtrl.RunWithOpts([]string{"package", "repository", "add", "-r", pkgrName, "--url", pkgrURL}, RunOpts{AllowError: true})
+		if err != nil {
+			fmt.Println(out, err)
```
cppforlife: nits: remove fmt.Print...

pkg/pkgrepository/app_template.go
```diff
@@ -18,30 +27,116 @@ func (a *App) template(dirPath string) exec.CmdRunResult {
 
 	genericOpts := ctltpl.GenericOpts{Name: a.app.Name, Namespace: a.app.Namespace}
 
+	// first templating pass is to read all the files in
 	var result exec.CmdRunResult
-	var isStream bool
+	var template ctltpl.Template
+	template = a.templateFactory.NewYtt(*(a.app.Spec.Template[0]).Ytt, genericOpts)
+	result, _ = template.TemplateDir(dirPath)
```
cppforlife: assuming _ is actually an error type, we should check on it as well.

pkg/pkgrepository/app_template.go
```diff
@@ -18,30 +27,116 @@ func (a *App) template(dirPath string) exec.CmdRunResult {
 
 	genericOpts := ctltpl.GenericOpts{Name: a.app.Name, Namespace: a.app.Namespace}
 
+	// first templating pass is to read all the files in
 	var result exec.CmdRunResult
-	var isStream bool
+	var template ctltpl.Template
+	template = a.templateFactory.NewYtt(*(a.app.Spec.Template[0]).Ytt, genericOpts)
```
cppforlife: instead of `*(a.app.Spec.Template[0]).Ytt` lets pull this configuration into this file. right now it's quite decoupled so no way to know what it does.

pkg/pkgrepository/app_template.go
```diff
@@ -18,30 +27,116 @@ func (a *App) template(dirPath string) exec.CmdRunResult {
...
-			return exec.NewCmdRunResultWithErr(fmt.Errorf("Unsupported way to template"))
+	// intermediate phase deserializes and reserializes all of the resources
+	// which guarantees that they only include fields this version of kc knows about.
+	streamForSomeReason := strings.NewReader(result.Stdout)
+	buf := make([]byte, streamForSomeReason.Len())
+	_, err := io.ReadFull(streamForSomeReason, buf)
+	if err != nil {
+		result.Error = err
+		return result
+	}
```
cppforlife: you can just convert string to []byte via type cast: `[]byte(result.Stdout)`

pkg/pkgrepository/app_template.go
```diff
@@ -18,30 +27,116 @@ func (a *App) template(dirPath string) exec.CmdRunResult {
...
+
+	// third templating inserts a kapp rebase rule to allow noops on identical resources provided by multiple pkgrs
+	template = a.templateFactory.NewYtt(*(a.app.Spec.Template[2]).Ytt, genericOpts)
+	result = template.TemplateStream(strings.NewReader(resources), dirPath)
+
+	a.log.Info("returning from the great yaml caper")
```
cppforlife: nit: not a helpful log.

pkg/pkgrepository/app_template.go
```diff
@@ -18,30 +27,116 @@ func (a *App) template(dirPath string) exec.CmdRunResult {
...
+	return result
+}
+
+// FilterResources takes a multi-doc yaml byte array of the templated
+// contents of a PKGR, and filters each resource by deserializing and re-serializing
+func FilterResources(yamls []byte) (string, error) {
```
cppforlife: nit: lets use the same type for input/output -- either string or []byte.

pkg/pkgrepository/app_template.go
```diff
@@ -18,30 +27,116 @@ func (a *App) template(dirPath string) exec.CmdRunResult {
...
+	}
+	decoder := serializer.NewCodecFactory(sch).UniversalDeserializer().Decode
+
+	filteredYamls := []string{}
+	docs, err := yamlDocs(yamls)
+	if err != nil {
+		return "", err
+	}
+	for _, resourceYAML := range docs {
+
```
cppforlife, line 282: nit: newline should probably be above for loop.

pkg/pkgrepository/app_template.go
```diff
@@ -18,30 +27,116 @@ func (a *App) template(dirPath string) exec.CmdRunResult {
...
+		obj, gKV, err := decoder(resourceYAML, nil, nil)
+		if err != nil {
+			return "", err
+		}
+		kind := gKV.Kind
+		if gKV.Group != "data.packaging.carvel.dev" {
+			return "", fmt.Errorf("got unexpected group: %s", gKV.Group)
```
cppforlife: `got unexpected group` => `Expected group '%s' but was '%s'`

pkg/pkgrepository/app_template.go
```diff
@@ -18,30 +27,116 @@ func (a *App) template(dirPath string) exec.CmdRunResult {
...
+			return "", fmt.Errorf("got unexpected version: %s", gKV.Version)
 		}
-		if result.Error != nil {
+
+		switch kind {
+		case "Package":
+			p := obj.(*datapackagingv1alpha1.Package)
+			yp := printers.YAMLPrinter{}
+			buffy := new(bytes.Buffer)
+			yp.PrintObj(p, buffy)
```
cppforlife: i think Print is for "pretty printing". we should just be able to use yaml.Marshal() (https://github.com/kubernetes-sigs/yaml/blob/master/yaml.go#L31)

pkg/pkgrepository/app_template.go
```diff
@@ -18,30 +27,116 @@ func (a *App) template(dirPath string) exec.CmdRunResult {
...
+	// third templating inserts a kapp rebase rule to allow noops on identical resources provided by multiple pkgrs
+	template = a.templateFactory.NewYtt(*(a.app.Spec.Template[2]).Ytt, genericOpts)
+	result = template.TemplateStream(strings.NewReader(resources), dirPath)
+
+	a.log.Info("returning from the great yaml caper")
+	return result
+}
+
+// FilterResources takes a multi-doc yaml byte array of the templated
+// contents of a PKGR, and filters each resource by deserializing and re-serializing
```
cppforlife: should probably mention why we are doing deserializing and re-serializing.

test/e2e/kappcontroller/helm_test.go
```diff
@@ -226,6 +233,13 @@ spec:
...
-	kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name},
-		e2e.RunOpts{StdinReader: strings.NewReader(config)})
+	out, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name},
+		e2e.RunOpts{StdinReader: strings.NewReader(config), AllowError: true})
+	// but it's helpful to actually see the error message...
+	if err != nil {
+		fmt.Println(out)
+		kubectl := e2e.Kubectl{t, env.Namespace, logger}
+		fmt.Println(kubectl.Run([]string{"get", "app", "-oyaml"}))
+	}
```
cppforlife: can we solve this generically instead of adding this stanzas all over?

pkg/pkgrepository/package_repo_app.go
```diff
@@ -108,7 +115,132 @@ metadata:
...
+        #@   for attr in ex.keys():
+        #@     if not attr in nw:
+        #@       return False, "missing spec." + attr
+        #@     end
+        #@     if nw[attr] != ex[attr]:
+        #@       return False, "mismatch in spec." + attr
+        #@     end
+        #@   end
+        #@   return True, ""
+        #@ end
```
cppforlife: note from earlier convo: can we switch to using ==, and only do special casing when hinting at differences?

test/e2e/kappcontroller/package_repo_test.go
cppforlife: > should we add support into the e2e.runopts, like kubectlOnError: []string{"get", "package" , pkgName, "-oyaml"} ?

nice idea!

pkg/pkgrepository/app_template.go
joe-kimmel-vmw: it doesn't seem to be an error type.  For example,  [the OG line in appCR](https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/pkg/app/app_template.go#L46) - it's a flag indicating whether the return value should be treated as a stream,  which gets used when we loop generically through an unknown series of fetch stages, but here it's effectively hardcoded / known.

cppforlife: ah ok. never mind.

test/e2e/kappcontroller/package_repo_test.go
joe-kimmel-vmw: happy to talk about it but I didn't immediately see other options I thought were better.  it's all compromises, imo,  between hardcoding each one fully redundantly on the one end and pumping test artifacts through ytt on the other end.

cli/test/e2e/package_repository_test.go
```diff
@@ -53,9 +55,23 @@ func TestPackageRepository(t *testing.T) {
...
+
+		for i := 1; i < 5; i++ {
+			out = kubectl.Run([]string{"get", "packagerepository", pkgrName})
+			fmt.Println(out)
+			if strings.Contains(out, "failed") {
+				fmt.Println(kubectl.Run([]string{"get", "packagerepository", pkgrName, "-oyaml"}))
+			}
+			require.NotContains(t, out, "failed")
+			time.Sleep(time.Duration(i) * time.Second)
+		}
```
cppforlife: note: no need for this.

joe-kimmel-vmw:
docs: https://github.com/vmware-tanzu/carvel/pull/453

PR #665, opened 2022-05-05
Rename KC owned apps from `x-ctrl` to `x.app` or `x.pkgr`

### LINKED ISSUE #376: PackageRepository trying to delete when PackageInstall has same name and namespace
**What steps did you tak and What happened:**
We deployed a PackageRepository (foo) into the default namespace (with tanzu cli repository add command) that provided a Package and PackageMetadata named foo as well.
We then run `tanzu package install foo --packagename foo.example.com --version 1.0.0 -f my-values.yaml` and then, all of a sudden, the Package and PackageMetadata foo got uninstalled, and the PackageRepository foo printed an error like this:
```
kapp: Error: Applying delete namespace/foocustom (v1) cluster:
  Deleting resource namespace/foocustom (v1) cluster: namespaces "foocustom" is forbidden:
    User "system:serviceaccount:tkg-system:kapp-controller-sa" cannot delete resource "namespaces" in API group "" in the namespace "foocustom" (reason: Forbidden)
``` 

foocustom is a value we provided in the values-file to the namespace where foo bits should be installed.

If we install the PackageRepository on the global namespace and we do the same command to install the package on default namespace, everything works fine.

**What did you expect:**
That the Package was properly installed

**Environment:**
- kapp Controller version (execute `kubectl get deployment -n kapp-controller kapp-controller -o yaml` and the annotation is `kbld.k14s.io/images`): v0.25.0 (upstream kapp install) and v0.23.0 (tkg 1.4 kapp-controller install)
- Kubernetes version (use `kubectl version`): 1.22.1

---
Vote on this request

neil-hickey:
- PackageRepo will be suffixed with .pkgr
- Apps will be suffixed with .app (pkgi, appcr)

Fixes #376

```release-note
The managed app name suffix is being renamed from `-ctrl` to `.app` and `.pkgr` for AppCRs and PackageRepositoryCRs respectivefully
```

##### Review Checklist:

  a link to that PR
  change

```docs

```


pkg/deploy/interfaces.go
```diff
@@ -8,13 +8,13 @@ import (
 )
 
 type Deploy interface {
-	Deploy(tplOutput string, startedApplyingFunc func(),
+	Deploy(tplOutput string, appName string, startedApplyingFunc func(),
```
cppforlife: i think it would make sense to maintain Deploy interface as agnostic to appname. can this be injected into the constructor?

test/e2e/kappcontroller/migrate_managed_name_test.go
```diff
@@ -0,0 +1,133 @@
...
+	kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", oldName}, e2e.RunOpts{IntoNs: true, StdinReader: strings.NewReader(c)})
+
+	// kapp-controller takes over the managing of this resource as an PackageRepository
+	kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, e2e.RunOpts{StdinReader: strings.NewReader(pkgrYaml)})
+
+	_, err := kapp.RunWithOpts([]string{"inspect", "-a", name + ".pkgr", "--raw", "--tty=false", "--filter-kind=PackageRepository"}, e2e.RunOpts{AllowError: true})
+	assert.NoError(t, err, fmt.Sprintf("expected %s to exist but does not", name))
+
+	_, err = kapp.RunWithOpts([]string{"inspect", "-a", oldName, "--raw", "--tty=false", "--filter-kind=PackageRepository"}, e2e.RunOpts{AllowError: true})
+	assert.Error(t, err, fmt.Sprintf("expected %s not to exist", oldName))
```
cppforlife, line 137: would be nice to have an assertion that error is indicating "thing" is not found.

pkg/deploy/interfaces.go
neil-hickey: @cppforlife I opted for in-place editing of the appName ==> see https://github.com/vmware-tanzu/carvel-kapp-controller/pull/665/files#diff-33a2bb066bb47c93866134d1ee796e9bcad9ed87d6248e10428c21dd1183d245R34

WDYT? I'm not a huge fan of mutating state though, which this kinda does.

We can change the `newKapp` arguments to accept a different name, and use that (still overwriting the name in `GenericOpts`). Or the alternative is a new key in our constructor. I like the idea of the calling functions having the power over the name, so I went with this edit-in-place model for now.

cppforlife: > WDYT? I'm not a huge fan of mutating state though, which this kinda does.

agreed, mutation is not great. here is what im thinking of:

- NewKapp constructor can take a "suffix": https://github.com/vmware-tanzu/carvel-kapp-controller/blob/58ce5c8bd33e9b930e0d58a5299b2d563b24c0dc/pkg/deploy/kapp.go#L37. deploy.Factory.NewKapp method can instantiate NewKapp with ".app" suffix: https://github.com/vmware-tanzu/carvel-kapp-controller/blob/58ce5c8bd33e9b930e0d58a5299b2d563b24c0dc/pkg/deploy/factory.go#L64

- deploy.Factory.NewKappPrivileged can be renamed to NewKappPrivelegedForPackageRepository which would construct NewKapp with ".pkgr" suffix. https://github.com/vmware-tanzu/carvel-kapp-controller/blob/58ce5c8bd33e9b930e0d58a5299b2d563b24c0dc/pkg/deploy/factory.go#L70

this way this stuff is hidden from the rest of the code and the interface to deployment remains the same.

neil-hickey: Cool, I actually had it this way at some point. I ended up changing because I didn't want to make the `newKapp` vs `NewKappPrivileged` encode the `.app` vs `.pkgr` differences. But sounds like `NewKappPrivileged` will only EVER be used by package repositories, so this assumption is fine now. Thanks for the clarity

test/e2e/kappcontroller/migrate_managed_name_test.go
```diff
@@ -0,0 +1,135 @@
...
+	name := "simple-app"
+
+	cleanUp := func() {
+		kapp.Run([]string{"delete", "-a", name})
+		kapp.Run([]string{"delete", "-a", oldName})
+	}
+	cleanUp()
+	defer cleanUp()
+
+	c := `---
```
cppforlife: nit: typically in golang single letter names are only used for method receivers. lets call it a bit longer e.g. `config`

test/e2e/kappcontroller/migrate_managed_name_test.go
```diff
@@ -0,0 +1,135 @@
...
+              name: configmap
+            data:
+              key: value
+  template:
+    - ytt: {}
+  deploy:
+    - kapp:
+        inspect: {}`, name) + sas.ForNamespaceYAML()
+
+	// deploy a simple resource with the managed name
```
cppforlife: instead of comments lets use `logger.Section` similar to other e2e examples

voor:
How's this going to work for existing apps?

neil-hickey:
> How's this going to work for existing apps?

Existing KC managed apps (ending in `-ctrl`) will be renamed to the new pattern on next reconciliation.

PR #680, opened 2022-05-10
Add arm64 builds

### LINKED ISSUE #573: Create arm64 image for `kapp-controller`
Currently, `kapp-controller` cannot run on an ARM64 host, as there is no image[0].

Could we switch to doing multi-arch builds and produce arm64 alongside amd64?

[0]:

```
$ crane manifest ghcr.io/vmware-tanzu/carvel-kapp-controller@sha256:a97d9d43b8b5d996f6510ae88ee761ac0f1d883d04fa7402b32f8ce6f9f875bd
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
   "config": {
      "mediaType": "application/vnd.docker.container.image.v1+json",
      "size": 5471,
      "digest": "sha256:57a03345e2ece35c67f3e1ede6cfa10008af841d04a1370e9491475cb31d072b"
   },
   "layers": [
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 17535726,
         "digest": "sha256:259970e9133ac11d271009754f24a11f5f6dd553a4fd7ce214f36a605be71adb"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 102329116,
         "digest": "sha256:38fa5ae806a2cba3ece15743ee641b6ad21c3f2fad4452c1259d759f224d778d"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 177,
         "digest": "sha256:544dd9b3e981ed95d9d45c19ab24ef223d73e1f2302bee286975b000597eed84"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 336772,
         "digest": "sha256:d6c4922f95f2bdf246883dec707a07438ada5bf8ceedd4bc13197abb3648ff02"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 12563601,
         "digest": "sha256:e140a67716edcf8efb267689ef6d9bc800bb30c2b87b7adf8e1802e9b4878f13"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 13737243,
         "digest": "sha256:3c616adb79a7fc3b98ef79a5844290daa8ded835027acd3cf26cd16cfe6bc400"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 10912475,
         "digest": "sha256:b853756a6369238b2b2f94e657b7597f63b0cc917b99c9316582edcea1a4ba3d"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 6437530,
         "digest": "sha256:70ec834fcf4241aea9634bb05119afe861ab51fd4b7a1b2127e26b17205db578"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 7039747,
         "digest": "sha256:82b1d1a8eae1bc1d7a7bce092fd60d4243fb39ae88253b76252f6808f8b50cef"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 8371356,
         "digest": "sha256:4e87932c7ca35afbaa99b438fbdb84ead1f4b62a2a16639c4ed3c041c814b0f1"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 13522494,
         "digest": "sha256:433dd29b575264c4847b163177f6b765db11df38c35a9598360be91f04e031e8"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 2442718,
         "digest": "sha256:95525999d6dbd611b77fabda007fa3129ab8b8a4cbf7e19a5a39c0cf91c34465"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 5510531,
         "digest": "sha256:6fde91471c4ff553a35eee899bed1c01ebac1f9ce309447c0e1ceed5d274b10a"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 21091266,
         "digest": "sha256:3d3bef3596e24879faf108e6d1b3b9b0dfa99445c1483f2882470df490182007"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 29198052,
         "digest": "sha256:ebeb9d7977366aac485ef35389a24f1d9673d1b38ce4097e076d04a28c25e18e"
      }
   ]
}
```

benmoss:
- Switch to using [docker buildx](https://docs.docker.com/buildx/working-with-buildx) for building through kbld for multi-arch support

Fixes #573

```release-note
Adds support for arm64 in the kapp-controller image
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,8 +13,8 @@ RUN go run ./hack/dependencies.go install -d out --arch ${TARGETARCH} --os ${TAR
 # kapp-controller
 COPY . .
 # helpful ldflags reference: https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications
-RUN CGO_ENABLED=0 GOOS=linux GOOS=${TARGETOS} GOSARCH=${TARGETARCH} go build -mod=vendor -ldflags="-X 'main.Version=$KCTRL_VER'" -trimpath -o out/kapp-controller ./cmd/controller
-RUN CGO_ENABLED=0 GOOS=linux GOOS=${TARGETOS} GOSARCH=${TARGETARCH} go build -mod=vendor -ldflags="-X 'main.Version=$KCTRL_VER'" -trimpath -o out/kapp-controller-sidecarexec ./cmd/sidecarexec
+RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOSARCH=${TARGETARCH} go build -mod=vendor -ldflags="-X 'main.Version=$KCTRL_VER'" -trimpath -o out/kapp-controller ./cmd/controller
+RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOSARCH=${TARGETARCH} go build -mod=vendor -ldflags="-X 'main.Version=$KCTRL_VER'" -trimpath -o out/kapp-controller-sidecarexec ./cmd/sidecarexec
 
 # --- run image ---
 FROM photon:4.0
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
-FROM golang:1.18.3 AS deps
+FROM --platform=$BUILDPLATFORM golang:1.18.3 AS deps
 
-ARG KCTRL_VER=development
+ARG TARGETOS TARGETARCH KCTRL_VER=development
 WORKDIR /workspace
 
 # dependencies
@@ -8,13 +8,13 @@ COPY ./hack/dependencies.* ./hack/
 COPY go.* ./
 COPY vendor vendor
 RUN mkdir out
-RUN go run ./hack/dependencies.go install -d out
+RUN go run ./hack/dependencies.go install -d out --arch ${TARGETARCH} --os ${TARGETOS}
 
 # kapp-controller
 COPY . .
 # helpful ldflags reference: https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications
-RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags="-X 'main.Version=$KCTRL_VER'" -trimpath -o out/kapp-controller ./cmd/controller/...
-RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags="-X 'main.Version=$KCTRL_VER'" -trimpath -o out/kapp-controller-sidecarexec ./cmd/sidecarexec/main.go
+RUN CGO_ENABLED=0 GOOS=linux GOOS=${TARGETOS} GOSARCH=${TARGETARCH} go build -mod=vendor -ldflags="-X 'main.Version=$KCTRL_VER'" -trimpath -o out/kapp-controller ./cmd/controller
+RUN CGO_ENABLED=0 GOOS=linux GOOS=${TARGETOS} GOSARCH=${TARGETARCH} go build -mod=vendor -ldflags="-X 'main.Version=$KCTRL_VER'" -trimpath -o out/kapp-controller-sidecarexec ./cmd/sidecarexec
 
 # --- run image ---
 FROM photon:4.0
--- a/config/build.yml
+++ b/config/build.yml
@@ -5,14 +5,17 @@ kind: Sources
 sources:
 - image: kapp-controller
   path: .
-  #@ if/end not data.values.image_cache:
   docker:
-    build:
-      #! Always rebuild image
+    buildx:
       pull: true
-      noCache: true
+      noCache: #@ not data.values.image_cache
       #! pass kapp_controller_version into Dockerfile.
-      rawOptions: #@ ["--build-arg", "KCTRL_VER="+data.values.kapp_controller_version]
+      rawOptions:
+      - #@ "--build-arg=KCTRL_VER="+data.values.kapp_controller_version
+      #@ if/end data.values.push_secret != "":
+      - #@ "--registry-secret="+data.values.push_secret
+      #@ if/end data.values.platform != "":
+      - #@ "--platform="+data.values.platform
 
 #@ if/end data.values.push_images:
 ---
--- a/config/build.yml
+++ b/config/build.yml
@@ -12,8 +12,6 @@ sources:
       #! pass kapp_controller_version into Dockerfile.
       rawOptions:
       - #@ "--build-arg=KCTRL_VER="+data.values.kapp_controller_version
-      #@ if/end data.values.push_secret != "":
-      - #@ "--registry-secret="+data.values.push_secret
       #@ if/end data.values.platform != "":
       - #@ "--platform="+data.values.platform
 
--- a/hack/dependencies.yml
+++ b/hack/dependencies.yml
@@ -3,6 +3,7 @@
       amd64: c370fab22081b17b780d4ec37fb664edb58fe6f28d635ad4beccbac7997586f9
     linux:
       amd64: 65dbc4f3a4a2ed84296dd1b323e8e7bd77e488fa7540d12dd36cf7fb2fc77c03
+      arm64: 3eaa8fcb3526a2a305a4845196bbffe7e1b0f3ea57fcd90daad94c855139df86
   dev: true
   name: ytt
   repo: vmware-tanzu/carvel-ytt
@@ -13,6 +14,7 @@
       amd64: 5fd3afbe09a66fc9e91252c57448045eeeab1ee3f6d305da6f9885186b744e67
     linux:
       amd64: 67c86ece94a3747b2e011a5b72044b69f2594ca807621b1e1e4c805f6abcaeef
+      arm64: 658725c5dab5349dd4cb826ff9278749f093ed2275aadf51dc994265e6599b2c
   dev: true
   name: kbld
   repo: vmware-tanzu/carvel-kbld
@@ -23,6 +25,7 @@
       amd64: 51e209f903b17c9e330151c41c6d507bfabccfdaa56dd8fdaa1c0f39ed2240e9
     linux:
       amd64: dec5040d90478fdf0af3c1548d46f9ded642f156245bba83fe99171c8461e4f7
+      arm64: e2047147c4bc00d8fcf9a5f01bdefc9fdf0aa5de9582357f03e54d4cd9fc0d75
   dev: true
   name: kapp
   repo: vmware-tanzu/carvel-kapp
@@ -33,6 +36,7 @@
       amd64: b8462c9b2b9c060e1a81c206ea9810937bf7357ae2ee150f94ec8635dff1d612
     linux:
       amd64: c7190adcb8445480e4e457c899aecdf7ca98606c625493b904c0eb2ab721ce19
+      arm64: 415120c4ede6ff1456feae6b980319998c0a3fc2128bbac1946ba98ce4d07aaf
   dev: true
   name: imgpkg
   repo: vmware-tanzu/carvel-imgpkg
@@ -43,6 +47,7 @@
       amd64: c26547097d67f21e129a25557d9d36c7c0e109afe130adff63d3c83ce9459ecc
     linux:
       amd64: 1aa12d070f2e91fcb0f4d138704c5061075b0821e6f943f5a39676d7a4709142
+      arm64: 015977ae54d85bf2366d7affb0d582fecf79737f0eb80fa8a66de9f66e877b61
   dev: true
   name: vendir
   repo: vmware-tanzu/carvel-vendir
@@ -51,6 +56,7 @@
 - checksums:
     linux:
       amd64: 1484ffb0c7a608d8069470f48b88d729e88c41a1b6602f145231e8ea7b43b50a
+      arm64: 5c0aa709c5aaeedd190907d70f9012052c1eea7dff94bffe941b879a33873947
   dev: false
   name: helm
   repo: helm/helm
@@ -60,6 +66,7 @@
 - checksums:
     linux:
       amd64: 53aec65e45f62a769ff24b7e5384f0c82d62668dd96ed56685f649da114b4dbb
+      arm64: 4945313ed0dfddba52a12ab460d750c91ead725d734039493da0285ad6c5f032
   dev: false
   name: sops
   repo: mozilla/sops
@@ -68,6 +75,7 @@
 - checksums:
     linux:
       amd64: 6414f71ce947fbbea1314f6e9786c5d48436ebc76c3fd6167bf018e432b3b669
+      arm64: 6c82aa1d406e5a401ec3bb344cd406626478be74d5ae628f192d907cd78af981
   dev: false
   name: age
   repo: FiloSottile/age
@@ -77,6 +85,7 @@
 - checksums:
     linux:
       amd64: 5e7ecb614b5926acfc36eb1258800391ab7c6e6e026fa7cacbfe92006bac895c
+      arm64: a8c3f4140d18c324cc69f5de4df0566e529e1636cff340095a42475799bf3fed
   dev: false
   name: cue
   repo: cue-lang/cue
--- a/.github/workflows/release-process.yml
+++ b/.github/workflows/release-process.yml
@@ -32,6 +32,8 @@ jobs:
       - name: Run release script
         run: |
           set -e -x
+          minikube start --driver=docker --wait=all
+          docker buildx create minikube --use --driver=kubernetes --bootstrap
           ./hack/build-release.sh
           cd cli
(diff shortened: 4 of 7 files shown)


benmoss:
Need to sync Dockerfile / dependencies.yml version drift

hack/dependencies.yml
```diff
@@ -1,16 +1,134 @@
...
-kapp_checksum_linux: 2b1bd5e3a35ed63045be80a7cc8cbc054fe4e98c89c55f9006c8fec2243658a4
-imgpkg_version: v0.28.0
-imgpkg_checksum_darwin: e43142fdb197a62844acb29cb619d513346aac3c23732a4d180c0ad974d9562e
-imgpkg_checksum_linux: 8d22423dd6d13efc0e580443d8f88d2183c52c6f851ba51e3e54f25bf140be58
-vendir_version: v0.27.0
-vendir_checksum_darwin: c26547097d67f21e129a25557d9d36c7c0e109afe130adff63d3c83ce9459ecc
-vendir_checksum_linux: 1aa12d070f2e91fcb0f4d138704c5061075b0821e6f943f5a39676d7a4709142
+- autoupdate:
+    checksums:
+      file: checksums.txt
```
neil-hickey: What's the difference between this line and the checksums section a few lines down?

neil-hickey, line 3: What's the difference between this line and the checksums section a few lines down?

.github/workflows/dependency-updater.yml
```diff
@@ -4,55 +4,31 @@ on:
...
-          old_linux_amd64_sha=$(yq eval '.${{ matrix.tool }}_checksum_linux' ./hack/dependencies.yml)
-          
-          release_notes=$(curl https://api.github.com/repos/vmware-tanzu/carvel-${{ matrix.tool }}/releases/latest | jq .body -r)
-          new_darwin_amd64_sha=$( echo "$release_notes" | grep ${{ matrix.tool }}-darwin-amd64 | awk '{print $1;}')
-          new_linux_amd64_sha=$( echo "$release_notes" | grep ${{ matrix.tool }}-linux-amd64 | awk '{print $1;}')
-                    
-          yq eval '.${{ matrix.tool }}_version = "${{ steps.latest-version.outputs.version }}"' -i ./hack/dependencies.yml
-          yq eval ".${{ matrix.tool }}_checksum_darwin = \"${new_darwin_amd64_sha}\"" -i ./hack/dependencies.yml
-          yq eval ".${{ matrix.tool }}_checksum_linux = \"${new_linux_amd64_sha}\"" -i ./hack/dependencies.yml          
+        run: go run ./hack/dependencies.go update
```
neil-hickey: I the one downside (potentially) between what exists now and this approach is that all the updates are bundled together.

One could argue it's a good thing, but if we find a bug in something and need to revert, it's a little less obvious. Still very doable. Just a call out, I'm fine with this though

config/build.yml
```diff
@@ -5,14 +5,17 @@ kind: Sources
 sources:
 - image: kapp-controller
   path: .
-  #@ if/end not data.values.image_cache:
-  docker:
+  kubectlBuildkit:
```
neil-hickey: this was a really neat find @benmoss . I like that we can keep kbld but also use more functionality that we need. 💯

hack/dependencies.yml
```diff
@@ -1,16 +1,134 @@
...
+    github: vmware-tanzu/carvel-ytt
+  checksums:
+    darwin:
+      amd64: d46dba5e729e2fe36c369e96eaa2eb5354fb1bf7cf9184f9bfa829b8e5558b94
+      arm64: a478eb4a6a88a8e3de9dd2e97c1fb96ca7a65c74525f2e132e6a5b333803f7c8
+    linux:
+      amd64: 11222665c627b8f0a1443534a3dde3c9b3aac08b322d28e91f0e011e3aeb7df5
+      arm64: c5d2f033b375ee87414b51d16c324d7a441de2f256865b5c774c4d5aea30ef60
+  dev: true
```
neil-hickey, line 7: It feels like there should be a doc that describes (or comments) these fields without looking at the kinda sorta ugly go code. Like what does `dev: true` vs `dev: false` mean? And `prod`?

neil-hickey: I also don't THINK we have tests, so just wondering what happens if we mess up some of these field positions in yaml, does it fail silently?

hack/dependencies.yml
benmoss: This is part of `autoUpdate`, it's for reimplementing the github action that checks for new versions. It's specifying that this release has checksums in a file named `checksums.txt`

hack/dependencies.yml
benmoss: > without looking at the kinda sorta ugly go code.

I'd rather we make the code reasonable enough to understand rather than maintain documentation for what's essentially a fancy bash script

> I also don't THINK we have tests, so just wondering what happens if we mess up some of these field positions in yaml, does it fail silently?
Yeah there aren't tests, though there weren't tests for our Dockerfile/dependencies.yml before either. I guess whether it causes something to fail or not depends on what you change in the YAML file.

- If you made a `dev` dependency no longer `dev`, the tests/other scripts that depend on those tools being installed would fail.
- If you removed a checksum it would no longer install on that os/arch

neil-hickey: > I'd rather we make the code reasonable enough to understand rather than maintain documentation for what's essentially a fancy bash script

Fair, I was referring to the `dependencies.yml` structure though, not the code. The code can parse, ignore, whatever the fields in this yaml file. But how do I know what to put in here, without reading gocode? Why should I read gocode for an api? That's how I am reading this `dependencies.yml` file. It's our interface for dependencies.

.github/workflows/benchmark-action.yaml
```diff
@@ -31,13 +31,9 @@ jobs:
       - name: Setup k8s and Run benchmark
         run: |
           set -e -x
-          mkdir /tmp/bin
-          export PATH=/tmp/bin:$PATH
-          curl -sLo /tmp/bin/minikube https://github.com/kubernetes/minikube/releases/latest/download/minikube-linux-amd64
```
joe-kimmel-vmw: it just magically has minikube now?

Dockerfile
```diff
@@ -1,36 +1,19 @@
-FROM golang:1.17.9 AS deps
```
joe-kimmel-vmw: looks like this divergence already existed but we're at 17.9 down here, and 17.6 somewhere above

hack/dependencies.go
```diff
@@ -0,0 +1,443 @@
...
+			if hdr.Name == tarballSubpath.String() {
+				out, err := os.Create(path.Join(c.destDir, dep.Name))
+				if err != nil {
+					return err
+				}
+				_, err = io.Copy(out, tgzReader)
+				if err != nil {
+					return err
+				}
+				if err := out.Chmod(0777); err != nil {
```
joe-kimmel-vmw: nit:  i hate to say it but this is stylistically inconsistent with the nil checks above and also with our general thing of preferring to spend max lines possible on nil checks.

joe-kimmel-vmw (commented):
i left you some annoying comments but this seems fine and I appreciate you taking it on

hack/dependencies.go
benmoss: yeah, i think single-line `if err := foo(); err != nil`s are great, I will never change

Dockerfile
benmoss: You mean like [in the GitHub actions](https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/.github/workflows/benchmark-action.yaml#L27)? Yeah, I think we should probably float those versions. This one dependabot will keep up to date.

.github/workflows/benchmark-action.yaml
benmoss: [always has been](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#tools)

hack/dependencies.yml
benmoss: I did another pass at the code and removed a lot of the more confusing elements of the `dependencies.yml` now. I still would be ok with documenting the structure somehow if you think that'd be helpful.

.github/workflows/dependency-updater.yml
benmoss: Is the concern that we update multiple dependencies in a single PR and then later discover some kind of bug, and it's harder to discern which PR caused the bug? I think even if they were separate PRs, unless the PR tests fail it's going to be hard later down the line to connect the bug to a dependency bump.

.github/workflows/benchmark-action.yaml
neil-hickey: I guess maybe it's just not latest? Depends on if we want to be explicit with the minikube version or not

.github/workflows/dependency-updater.yml
neil-hickey: Yeh, I just like separate PRs because of what you described. And reverting is easier.

hack/dependencies.yml
neil-hickey: I think it's fine to leave as is. The fields are self-explanatory (I was just confused by the `dev` vs `prod` flags and how they interacted with each other). Looks like `prod` was removed, so 👍

.github/workflows/benchmark-action.yaml
benmoss: IMHO seems better just to let them manage this. https://github.com/actions/virtual-environments/commit/12fa229b4a3eebb871ed90ef2c45c897a4fd728d implies they lag like a few days but are pretty quick to bump, 1.25.2 was released 2/24/22, they bumped 3/7.

benmoss:
- Test release from my fork built here: https://github.com/benmoss/carvel-kapp-controller/runs/6817310991
- Multi-arch image from that build: https://github.com/benmoss/carvel-kapp-controller/pkgs/container/carvel-kapp-controller/24748131?tag=rand-1654793952642415300-6418976100130-kapp-controller
- release: https://github.com/benmoss/carvel-kapp-controller/releases/tag/v0.37.14

Dockerfile
```diff
@@ -1,20 +1,20 @@
...
 RUN mkdir out
-RUN go run ./hack/dependencies.go install -d out
+RUN go run ./hack/dependencies.go install -d out --arch ${TARGETARCH} --os ${TARGETOS}
 
 # kapp-controller
 COPY . .
 # helpful ldflags reference: https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications
-RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags="-X 'main.Version=$KCTRL_VER'" -trimpath -o out/kapp-controller ./cmd/controller/...
-RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags="-X 'main.Version=$KCTRL_VER'" -trimpath -o out/kapp-controller-sidecarexec ./cmd/sidecarexec/main.go
+RUN CGO_ENABLED=0 GOOS=linux GOOS=${TARGETOS} GOSARCH=${TARGETARCH} go build -mod=vendor -ldflags="-X 'main.Version=$KCTRL_VER'" -trimpath -o out/kapp-controller ./cmd/controller
```
neil-hickey: `GOOS` is duplicated, is that needed?

benmoss: whoops

Dockerfile
```diff
@@ -1,20 +1,20 @@
...
 COPY go.* ./
 COPY vendor vendor
 RUN mkdir out
-RUN go run ./hack/dependencies.go install -d out
+RUN go run ./hack/dependencies.go install -d out --arch ${TARGETARCH} --os ${TARGETOS}
 
 # kapp-controller
 COPY . .
 # helpful ldflags reference: https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications
-RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags="-X 'main.Version=$KCTRL_VER'" -trimpath -o out/kapp-controller ./cmd/controller/...
```
neil-hickey, line 16: I imagine the `./cmd/controller/...` and `./cmd/sidecarexec/main.go` were shortened to `./cmd/controller` and `./cmd/sidecarexec` because go build just looks those directories anyways right?

neil-hickey (approved):
LGTM,  you have my stamp of approval ⭐

Dockerfile
benmoss: Yeah, that was just a tiny nit thing I threw in there. We're not building `controller/...`, just `controller` and its unnecessary to specify `main.go` (there could be other files that are part of the main package)

DennisDenuto:
Really cool stuff here!

> * Test release from my fork built here: https://github.com/benmoss/carvel-kapp-controller/runs/6817310991

Nice! I was wondering how you tested / iterated on this.

I am assuming that the `upgrade-testing` gh-action job is to test folk upgrading to the freshly cut release? If that is the case, can you update (on your fork) the upgrade-test to use the release produced on your fork? (Just to get added reassurance that the release.yaml produced in your fork results in a successful deploy)

https://github.com/vmware-tanzu/carvel-kapp-controller/blob/0726f760ad4f5685d7d9aab123d4732726e8b2cd/.github/workflows/upgrade-testing.yml#L40


Also, along this train of thought... and this suggestion probably makes way more sense to implement in a different PR, should we be adding upgrade-tests for [`arm`](https://github.com/marketplace/actions/arm-runner) deployments in addition to `amd64`, now that we produce images for both?

benmoss:
> Nice! I was wondering how you tested / iterated on this.

The main thing I needed to do was change `image_repo` from [`config/release`](https://github.com/vmware-tanzu/carvel-kapp-controller/blob/d49800d7d5024058911e096cfd7487f49ccf0c5f/config-release/values.yml#L5) to my github repo's GHCR, which I did by allowing `--data-values-env=KCTRL` and then [overriding it in my fork](https://github.com/benmoss/carvel-kapp-controller/blob/develop/.github/workflows/release-process.yml#L36) to just be relative to the current repo. It would be nice if this wasn't hard-coded in this repo.

Other than that just enabling GitHub actions on my fork and pushing tags is all you need to do. Just be extra paranoid that you're not pushing to this repo :). I'll usually just be explicit and `git push fork develop` instead of `git push`.

> I am assuming that the upgrade-testing gh-action job is to test folk upgrading to the freshly cut release? If that is the case, can you update (on your fork) the upgrade-test to use the release produced on your fork? (Just to get added reassurance that the release.yaml produced in your fork results in a successful deploy)

Yeah, I could definitely set this up.

>Also, along this train of thought... and this suggestion probably makes way more sense to implement in a different PR, should we be adding upgrade-tests for [arm](https://github.com/marketplace/actions/arm-runner) deployments in addition to amd64, now that we produce images for both?

Yeah, I was thinking we maybe want to run tests against an arm-deployed cluster. Will have to play around with that arm runner and see how well that works with minikube or what our other options would be.

benmoss:
Actually I don't think I understand your suggestion of running the upgrade tests from my release.. What would we be upgrading *to*? The `test-gh` and `Kind Cluster E2E tests` jobs test that this PR deploys successfully.

I think what you're describing is something like us generating a release.yml for every PR and just deploy/e2e that. I like the idea, I've been trying to think of how we could automate exercising our release process especially for work like this.

EIther way [this build](https://github.com/benmoss/carvel-kapp-controller/runs/6819794392?check_suite_focus=true) should test against my release.yml.

DennisDenuto:
> EIther way [this build](https://github.com/benmoss/carvel-kapp-controller/runs/6819794392?check_suite_focus=true) should test against my release.yml.

Yeah, seeing a successful deploy of the generated `release.yaml` based on your fork is what I was after. Alternatively a manual test using that release.yaml and seeing it succeed would also have given me that same level of confidence.

 > I think what you're describing is something like us generating a release.yml for every PR and just deploy/e2e that. I like the idea, I've been trying to think of how we could automate exercising our release process especially for work like this. 

+1

this PR LGTM! 🚀

DennisDenuto (approved):

Issue #661, opened 2022-05-04
Kicking an installed package should show fetch and deploy logs


voor:
**Describe the problem/challenge you have**
I am rapidly falling in love with kick in its many forms.  When I have an installed package and I kick it, I usually do this:

```
kctrl package installed kick -n common-install -i external-dns -y
kctrl app kick -n common-install -a external-dns -y
```

I would really like to be able to just do this:

```
kctrl package installed kick -n common-install -i external-dns -y --SOMETHING
```

And combine these into one command.

**Describe the solution you'd like**
Add a new `--follow` or `--recurse` or you come up with a better name, 

**Anything else you would like to add:**
Please name the parameter something that makes sense and don't feel like my suggestions are part of the feature request.

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

praveenrewar:
@voor Any specific reason you would want to kick the underlying app CR as well? 
(As kicking the package install would technically kick the underlying app CR)

voor:
I am interested in seeing the fetch and deploy outputs, that's not shown when I kick the pkg.

praveenrewar:
Ideally you should be able to see deploy outputs while kicking the package install, but I think that there are still some edge cases that needs to be covered. (cc @100mik)
Would you be able to share the output that you are seeing when kicking package install?

voor:
```
$ kctrl package installed kick -n common-install -i external-dns -y
Target cluster 'https://rv2-tap-apiserver-xxxx.us-gov-east-1.elb.amazonaws.com:6443' (nodes: ip-10-0-19-154.us-gov-east-1.compute.internal, 7+)

Triggering reconciliation for package install 'external-dns' in namespace 'common-install'

12:21:30PM: Waiting for PackageInstall reconciliation for 'external-dns'
12:21:31PM: Waiting for generation 4 to be observed 
12:20:23PM: App reconciled (1m ago)

Succeeded
```

voor:
I would expect to see this:
```
kctrl app kick -a external-dns -n common-install -y                                                                                                                         
[...]

PR #683, opened 2022-05-11
Add `--values` to `kctrl package installed update`

### LINKED ISSUE #675: `kctrl`: Allow users to drop a values secret being consumed by a package install
Allow users to dereference and garbage collect a secret suopplied to a package install craeted by kctrl

A flag `--no-values-file` which deletes and dereferences value secrets created by kctrl

**Acceptance criteria**
- [ ] Add flag to drop values if the user does not supply a values file
- [ ] Dereference 
- [ ] Assert ownership before deletion
- [ ] Tests 

---
Vote on this request

100mik:
This flag allows users to drop values secrets being consumed by a package and remove it's references in a package installation while updating it. 
PR also fixes value secret updates clobbering ownership annotations.

Fixes #675 

```release-note
`--values` can be used to drop values secret being consumed by package install, when no value is provided via `--values-file`
```

##### Review Checklist:

  a link to that PR
  change

### DIFF
--- a/cli/pkg/kctrl/cmd/package/installed/create_or_update.go
+++ b/cli/pkg/kctrl/cmd/package/installed/create_or_update.go
@@ -43,7 +43,7 @@ type CreateOrUpdateOptions struct {
 	packageName        string
 	version            string
 	valuesFile         string
-	dropValuesFile     bool
+	withValues         bool
 	serviceAccountName string
 
 	install bool
@@ -149,6 +149,8 @@ func NewUpdateCmd(o *CreateOrUpdateOptions, flagsFactory cmdcore.FlagsFactory) *
 			},
 			cmdcore.Example{"Update package install with new values file",
 				[]string{"package", "installed", "update", "-i", "cert-man", "--values-file", "values.yml"}},
+			cmdcore.Example{"Update package install to stop consuming supplied values",
+				[]string{"package", "installed", "update", "-i", "cert-man", "--with-values", "false"}},
 		}.Description("-i", o.pkgCmdTreeOpts),
 	}
 	o.NamespaceFlags.SetWithPackageCommandTreeOpts(cmd, flagsFactory, o.pkgCmdTreeOpts)
@@ -163,7 +165,7 @@ func NewUpdateCmd(o *CreateOrUpdateOptions, flagsFactory cmdcore.FlagsFactory) *
 	cmd.Flags().StringVarP(&o.packageName, "package", "p", "", "Name of package install to be updated")
 	cmd.Flags().StringVar(&o.version, "version", "", "Set package version")
 	cmd.Flags().StringVar(&o.valuesFile, "values-file", "", "The path to the configuration values file, optional")
-	cmd.Flags().BoolVar(&o.dropValuesFile, "drop-values-file", false, "Drop values file and remove reference from installation if values are not supplied, optional")
+	cmd.Flags().BoolVar(&o.withValues, "with-values", true, "Add or keep values supplied to package install, optional")
 
 	o.WaitFlags.Set(cmd, flagsFactory, &cmdcore.WaitFlagsOpts{
 		AllowDisableWait: true,
@@ -278,8 +280,6 @@ func (o *CreateOrUpdateOptions) RunUpdate(args []string) error {
 	}
 
 	o.createdAnnotations = NewCreatedResourceAnnotations(o.Name, o.NamespaceFlags.Name)
-
-	o.statusUI.PrintMessagef("Getting package install for '%s'", o.Name)
 	pkgInstall, err := kcClient.PackagingV1alpha1().PackageInstalls(o.NamespaceFlags.Name).Get(
 		context.Background(), o.Name, metav1.GetOptions{},
 	)
@@ -301,8 +301,8 @@ func (o CreateOrUpdateOptions) update(client kubernetes.Interface, kcClient kccl
 		return err
 	}
 
-	if o.valuesFile == "" && !changed && !o.dropValuesFile {
-		return err
+	if o.valuesFile == "" && !changed && o.withValues {
+		return nil
 	}
 
 	isSecretCreated, err := o.createOrUpdateValuesSecret(updatedPkgInstall, client)
@@ -311,7 +311,7 @@ func (o CreateOrUpdateOptions) update(client kubernetes.Interface, kcClient kccl
 	}
 
 	isSecretDeleted := false
-	if o.dropValuesFile && o.valuesFile == "" {
+	if !o.withValues && len(updatedPkgInstall.Spec.Values) > 0 {
 		isSecretDeleted, err = o.dropValuesSecret(client)
 		if err != nil {
 			return err
@@ -322,6 +322,8 @@ func (o CreateOrUpdateOptions) update(client kubernetes.Interface, kcClient kccl
 			o.removeValuesSecretReference(updatedPkgInstall)
 			changed = true
 		}
+	} else if !o.withValues && len(updatedPkgInstall.Spec.Values) == 0 {
+		o.statusUI.PrintMessagef("No values have been supplied to installation '%s' in namespace '%s'", o.Name, o.NamespaceFlags.Name)
 	}
 
 	o.statusUI.PrintMessagef("Updating package install for '%s' in namespace '%s'", o.Name, o.NamespaceFlags.Name)
--- a/cli/pkg/kctrl/cmd/package/installed/create_or_update.go
+++ b/cli/pkg/kctrl/cmd/package/installed/create_or_update.go
@@ -722,11 +722,10 @@ func (o *CreateOrUpdateOptions) updateDataValuesSecret(client kubernetes.Interfa
 	if err != nil {
 		return fmt.Errorf("Reading data values file '%s': %s", o.valuesFile, err.Error())
 	}
-	secret := &corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{Name: secretName, Namespace: o.NamespaceFlags.Name}, Data: dataValues,
-	}
 
-	_, err = client.CoreV1().Secrets(o.NamespaceFlags.Name).Update(context.Background(), secret, metav1.UpdateOptions{})
+	updatedSecret := createdSecret.DeepCopy()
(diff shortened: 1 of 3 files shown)


100mik:
This also seems to be affected by #661 
i.e. the tail is sometimes not picked up after the changes are applied even though the desired effect is observed on the cluster. This bnehaviour should be corrected with its fix as well.

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -333,6 +349,48 @@ func (o CreateOrUpdateOptions) update(client kubernetes.Interface, kcClient kccl
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			o.statusUI.PrintMessagef("Values secret '%s' not found", secretName)
+			return true, nil
+		}
+		return false, fmt.Errorf("Getting values secret: %s", err.Error())
+	}
+
+	// Do not delete secrets not created by kctrl
+	// TODO: Should we not error out and continue with the rest of the changes?
```
100mik, line 368: Food for thought ^
Rn I lean towards erroring out
(Do not feel like it makes sense to just bump version, if the user wanted to drop config and bump)
Might remove this TODO if everyone feels the same way

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -333,6 +349,48 @@ func (o CreateOrUpdateOptions) update(client kubernetes.Interface, kcClient kccl
...
+	if err != nil {
+		return false, fmt.Errorf("Deleting values secret: %s", err.Error())
+	}
+
+	return true, nil
+}
+
+func (o *CreateOrUpdateOptions) removeValuesSecretReference(pkgi *kcpkgv1alpha1.PackageInstall) {
+	for i, valueRef := range pkgi.Spec.Values {
+		if valueRef.SecretRef.Name == o.createdAnnotations.SecretAnnValue() {
```
100mik, line 390: Should we just clobber all secrets here?
I was trying to ensure that we have minimum chances of doing that to config not added by kctrl.

We could simply do something like
```go
pkgi.Spec.Values = []kcpkgv1alpha1.PackageInstallValues{}
```

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -162,6 +163,7 @@ func NewUpdateCmd(o *CreateOrUpdateOptions, flagsFactory cmdcore.FlagsFactory) *
 	cmd.Flags().StringVarP(&o.packageName, "package", "p", "", "Name of package install to be updated")
 	cmd.Flags().StringVar(&o.version, "version", "", "Set package version")
 	cmd.Flags().StringVar(&o.valuesFile, "values-file", "", "The path to the configuration values file, optional")
+	cmd.Flags().BoolVar(&o.dropValuesFile, "drop-values-file", false, "Drop values file and remove reference from installation if values are not supplied, optional")
```
cppforlife: it seems that drop-values-file is too specific of a name for the feature since it's removing all values. i wonder if we should really do `--values=bool` (with default=true) or something along those lines that represents that no values will be set on the packageinstall.

--with-values?

100mik: Just `--values` makes sense to me as well.
In that case the behaviour would be that the provided values are ignored, if the flag is set to false imo.

`--with-values` might be more instinctive

cppforlife: > provided values are ignored, if the flag is set to false imo.

may be even raise an error?

100mik: Sounds good!

100mik: Went ahead with `--values` as it might make more sense since it is in create, update and install commands

100mik:
Doc PR here : https://github.com/vmware-tanzu/carvel/pull/439

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -127,6 +128,7 @@ func NewInstallCmd(o *CreateOrUpdateOptions, flagsFactory cmdcore.FlagsFactory)
 	cmd.Flags().StringVar(&o.version, "version", "", "Set package version (required)")
 	cmd.Flags().StringVar(&o.serviceAccountName, "service-account-name", "", "Name of an existing service account used to install underlying package contents, optional")
 	cmd.Flags().StringVar(&o.valuesFile, "values-file", "", "The path to the configuration values file, optional")
+	cmd.Flags().StringVar(&o.valuesFile, "values", "", "The path to the configuration values file, optional")
```
cppforlife: string?

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -148,6 +151,8 @@ func NewUpdateCmd(o *CreateOrUpdateOptions, flagsFactory cmdcore.FlagsFactory) *
 			},
 			cmdcore.Example{"Update package install with new values file",
 				[]string{"package", "installed", "update", "-i", "cert-man", "--values-file", "values.yml"}},
+			cmdcore.Example{"Update package install to stop consuming supplied values",
+				[]string{"package", "installed", "update", "-i", "cert-man", "--with-values", "false"}},
```
praveenrewar:
```suggestion
				[]string{"package", "installed", "update", "-i", "cert-man", "--values", "false"}},
```

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -299,17 +303,38 @@ func (o CreateOrUpdateOptions) update(client kubernetes.Interface, kcClient kccl
...
 		return err
 	}
 
-	o.statusUI.PrintMessagef("Updating package install for '%s'", o.Name)
-	o.addCreatedResourceAnnotations(&updatedPkgInstall.ObjectMeta, false, isSecretCreated)
+	isSecretDeleted := false
+	if !o.values && len(updatedPkgInstall.Spec.Values) > 0 {
+		isSecretDeleted, err = o.dropValuesSecret(client)
+		if err != nil {
+			return err
```
praveenrewar: Should we have something like this instead of wrapping the error at all the return statements in `dropValuesSecret`?
```suggestion
			fmt.Errorf("Deleting values secret: %s", err.Error())
```

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -299,17 +303,38 @@ func (o CreateOrUpdateOptions) update(client kubernetes.Interface, kcClient kccl
...
-	if o.valuesFile == "" && !changed {
-		return err
+	if o.valuesFile == "" && !changed && o.values {
+		o.statusUI.PrintMessagef("No changes to package install '%s' in namespace '%s'", o.Name, o.NamespaceFlags.Name)
+		return nil
 	}
 
 	isSecretCreated, err := o.createOrUpdateValuesSecret(updatedPkgInstall, client)
 	if err != nil {
+		o.statusUI.PrintMessagef("", o.Name)
```
praveenrewar: This looks weird. Was it intentional?

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
praveenrewar: > // Do not delete secrets not created by kctrl

I am just trying to remind myself why we decided not to delete secrets that were not created by kctrl 🤔

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -327,6 +352,48 @@ func (o CreateOrUpdateOptions) update(client kubernetes.Interface, kcClient kccl
...
+		return false, fmt.Errorf("Deleting values secret: %s", err.Error())
+	}
+
+	return true, nil
+}
+
+func (o *CreateOrUpdateOptions) removeValuesSecretReference(pkgi *kcpkgv1alpha1.PackageInstall) {
+	for i, valueRef := range pkgi.Spec.Values {
+		if valueRef.SecretRef.Name == o.createdAnnotations.SecretAnnValue() {
+			pkgi.Spec.Values = append(pkgi.Spec.Values[:i], pkgi.Spec.Values[i+1:]...)
```
praveenrewar, line 391: Maybe it won't happen in this specific scenario, but I think this could panic with `slice bounds out of range` in some situations as the variable `i` is not updated after removing a specific item from the slice, so value of `i` could be greater than the length of the slice.

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
100mik: It is to safeguard around scenarios where there is an installation up and running and an user who might not be aware of the configuration.
Similar to how we error out if there are multiple value references present.

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
100mik: You are correct, I was assuming that there won't be more than one value which matches our naming convention.
To be safer I am gonna return after removing one reference now.
Then we do not run into the funky case.

I am wondering if we should just clobber all the values supplied and make it an empty array / nil value. I thought this would be safer though.

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
100mik: Yep makes sense! did that

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
100mik: Definitely not 
Cleaning up 🧹

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
100mik: Good catch! changed it

praveenrewar (approved):
LGTM

PR #693, opened 2022-05-17
Inject cmd runner

cppforlife:
extracted from dev deploy PR. purely non user facing changes.


pkg/app/app_factory.go
```diff
@@ -9,24 +9,33 @@ import (
...
 type CRDAppFactory struct {
-	CoreClient kubernetes.Interface
-	AppClient  kcclient.Interface
-	KcConfig   *config.Config
-	AppMetrics *metrics.AppMetrics
+	CoreClient     kubernetes.Interface
+	AppClient      kcclient.Interface
+	KcConfig       *config.Config
+	AppMetrics     *metrics.AppMetrics
+	VendirHookFunc func(vendirconf.Config) vendirconf.Config
```
joe-kimmel-vmw: just a note that at this point in the PR i'm still trying to guess what a "vendir hook func" does -- is it like a callback that gets executed at some point while vendir is executing?     To the extent you think that future readers of this file might also not be able to guess what a "vendir hook" is,  comments are a wonderful technological innovation for communicating additional context & understanding,

pkg/fetch/vendir.go
```diff
@@ -23,21 +25,36 @@ const (
...
+	SkipTLSConfig SkipTLSConfig
 }
 
-func NewVendir(nsName string, coreClient kubernetes.Interface, skipTLSConfig SkipTLSConfig) *Vendir {
+// NewVendir returns vendir.
+func NewVendir(nsName string, coreClient kubernetes.Interface,
+	opts VendirOpts, cmdRunner exec.CmdRunner) *Vendir {
+
+	if opts.HookFunc == nil {
+		opts.HookFunc = func(conf vendirconf.Config) vendirconf.Config { return conf }
```
joe-kimmel-vmw: so it's a config-hook?

joe-kimmel-vmw (commented):
i think im' pretty confused overall , maybe the context of the other PR would help?

it seems like the main change is that now all the command-line invocations go through what used to be the "cancelleable runner" but is now the PlainVanillaCommandExecutor or something right?   so that much seems unobjectionable.

 though maybe the `New` function feels a little overengineering for a struct that currently takes no params and sets no default settings?   and i still don't really get the vendir hook function?  so far i think it just takes the config and returns the config.  why is it called hook, and what would you use it for?

pkg/fetch/vendir.go
cppforlife: renamed to ConfigHook

pkg/app/app_factory.go
cppforlife: added suggested comment

PR #694, opened 2022-05-17
Add global kapp raw options and set default --app-changes-max-to-keep=5

cppforlife:

see release notes section below.

Fixes #

```release-note
- configures kapp to `--app-changes-max-to-keep=5` which is lower than a default offered by kapp itself (200). this is done to avoid having large number of ConfigMaps accumulate. ideally in future, kapp can be more intelligent about configmap deduplication.
- introduces `kappDeployRawOptions` controller level configuration in `kapp-controller-config` Secret. it allows to place additional kapp deploy flags that apply to all App CRs. App CR level configuration takes precedence over global ones.
```

##### Review Checklist:

- [x] N/A Relevant docs in this repo added or updated
- [ ] TODO Relevant carvel.dev docs added or updated in a separate PR and there's
  a link to that PR
  change


pkg/config/config.go
```diff
@@ -223,23 +244,33 @@ func (gc *Config) addKubernetesServiceHostInNoProxy() {
...
 	gc.noProxy = data[noProxyKey]
 	gc.skipTLSVerify = data[skipTLSVerifyKey]
+
+	if val := data[kappDeployRawOptionsKey]; len(val) > 0 {
+		var opts []string
+		err := json.Unmarshal([]byte(val), &opts)
+		if err != nil {
+			return fmt.Errorf("Unmarshaling kappDeployRawOptions as JSON: %s", err)
+		}
+		gc.kappDeployRawOptions = opts
```
joe-kimmel-vmw, line 274: i just dug around to remind myself where the allow-list is for this feature,  since at a glance it looks like we're taking unverified user input and passing on the command line.   how would you feel about adding a comment that says something like,  `// don't worry, input is verified before kapp is invoked. see pkg/deploy/kapp_restrict.go` ?

pkg/deploy/factory.go
```diff
@@ -16,16 +16,25 @@ import (
...
 	coreClient kubernetes.Interface
+	kappConfig KappConfiguration
 
 	kubeconfigSecrets *KubeconfigSecrets
 	serviceAccounts   *ServiceAccounts
 
 	cmdRunner exec.CmdRunner
 }
 
+// KappConfiguration allows to inject shared kapp settings.
```
joe-kimmel-vmw: nit but maybe "allows us to..." ?

pkg/deploy/kapp.go
```diff
@@ -162,6 +163,12 @@ func (a *Kapp) addDeployArgs(args []string) ([]string, error) {
 		args = append(args, []string{"--map-ns", val}...)
 	}
 
+	// Global raw options are applied first to be able to override them within an App
+	args, err := a.addRawOpts(args, a.globalDeployRawOpts, kappAllowedDeployFlagSet)
```
joe-kimmel-vmw, line 167: it feels funny to call the same fn twice in a row,  especially because the first time we do an error check but the second time we can just return whatever it returned.  Is it worth it to add a handler like, :

```
a.multiAddRawOpts(args, []RawOptions, FlagSet) args, error {
  var args;
   for opts in RawOptions {
       args, err := a.addRawOpts(args, opts, FlagSet)
       return err if err != nil
     }
     return args, nil
}
```

just to hide the sort of double-call? 

(forgive my pseudo code above. hopefully gets the gist across)

go.sum
```diff
@@ -233,13 +233,16 @@ github.com/coreos/go-systemd/v22 v22.3.2/go.mod h1:Y58oyj3AT4RCenI/lSvhwexgC+NSV
 github.com/coreos/pkg v0.0.0-20160727233714-3ac0863d7acf/go.mod h1:E3G3o1h8I7cfcXa63jLwjI0eiQQMgzzUDFVpN/nH/eA=
 github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f/go.mod h1:E3G3o1h8I7cfcXa63jLwjI0eiQQMgzzUDFVpN/nH/eA=
 github.com/cppforlife/cobrautil v0.0.0-20200514214827-bb86e6965d72/go.mod h1:2w+qxVu2KSGW78Ex/XaIqfh/OvBgjEsmN53S4T8vEyA=
+github.com/cppforlife/color v1.9.1-0.20200716202919-6706ac40b835 h1:mYQweUIBD+TBRjIeQnJmXr0GSVMpI6O0takyb/aaOgo=
```
joe-kimmel-vmw, line 236: I have to admit i'm a little surprised at how many libraries this change adds, since ostensibly it just adds one new command line argument to kapp-controller?    am i missing something? is there a story here?

joe-kimmel-vmw (commented):
i left you a couple small questions. no major objections

go.sum
cppforlife: it's because of `uitest "github.com/cppforlife/go-cli-ui/ui/test"` addition to the test to parse kapp json table.

pkg/deploy/kapp.go
cppforlife: ill disagree with this one. less indirection reads clearer.

pkg/deploy/factory.go
cppforlife: fixed.

pkg/config/config.go
cppforlife: fixed.

joe-kimmel-vmw (approved):
lgtm

Issue #687, opened 2022-05-13
Kapp-controller will not work properly in Kubernetes 1.24.


GrahamDumpleton:
I raised this in a private Slack channel discussion but since not seen any issue created here by others since that discussion, creating one.

**What steps did you take:**

Install `kapp-controller` in Kubernetes 1.24.

Try installing a package. Installation fails with the error:

```
kctrl: Error: packageinstall/xxx (packaging.carvel.dev/v1alpha1) namespace: xxx:
  Reconciling:
    Preparing kapp:
      Expected to find one service account token secret, but found none. Reconcile failed:
        Error (see .status.usefulErrorMessage for details)
```

**What happened:**

The problem relates to:

* https://www.padok.fr/en/blog/new-kubernetes-changes#No_Secret_by_default_for_service_account_tokens
* https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#no-really-you-must-read-this-before-you-upgrade-1
* https://github.com/kubernetes/kubernetes/pull/108309

That is, in Kubernetes 1.24, an API token secret is no longer automatically created and listed as a secret against a service account. For in-cluster API tokens for pods a new mechanism is now used to allow token refreshes every hours.

A service account therefore now looks like:

```
apiVersion: v1
kind: ServiceAccount
metadata:
  creationTimestamp: "2022-05-13T03:34:02Z"
  name: default
  namespace: lab-k8s-fundamentals-w01-s001
  resourceVersion: "1744"
  uid: ff853727-1989-4f49-b5ab-a0e1ec51944c
```

Note how there is no `secrets` property like before.

It seems `kapp-controller` would look up the API token secret name in `secrets` of the service account and use the token in the linked secret to do stuff. Since this no longer exists, `kapp-controller` fails with the error shown above.

**What did you expect:**

It works.

**Anything else you would like to add:**

This will likely have implications for where a distinct service account is provided by a user when installing a package, rather than the default one used. For the default, `kapp-controller` can do something else whereby it has created an API token secret of its own to use and linked it up to the default service account used.

For a user providing a service account that will no longer be sufficient, unless you tell users to also create a separate API token secret linked to the service account. What then needs to happen is that `kapp-controller` needs to search for any secret in the same namespace with the `kubernetes.io/service-account.name` annotation giving the name of the service account it is for.

[...]

PR #695, opened 2022-05-19
[k8 1.24] Use TokenRequest API to get SA token

joe-kimmel-vmw:

As part of kubernetes 1.24, a `Secret` is no longer created alongside a serviceaccount. This means that we no longer have access to the `token` generated by the Secret. This PR updates our codepath to use the `TokenRequest` API instead. This has been available (and turned on by default) since kubernetes 1.12 so we should not have any incompatibility from this change. There is no user facing change.

Fixes # https://github.com/vmware-tanzu/carvel-kapp-controller/issues/687

```release-note
NONE
```

##### Review Checklist:

  a link to that PR
  change

```docs

```


pkg/deploy/service_accounts.go
```diff
@@ -49,43 +56,26 @@ func (s *ServiceAccounts) fetchServiceAccount(nsName string, saName string) (str
 		return "", fmt.Errorf("Internal inconsistency: Expected service account name to not be empty")
 	}
 
-	sa, err := s.coreClient.CoreV1().ServiceAccounts(nsName).Get(context.Background(), saName, metav1.GetOptions{})
+	tokenMgr := token.NewManager(s.coreClient, s.log)
```
cppforlife: since NewManager is made every time we call Find no caching happens afaik.

pkg/deploy/service_accounts.go
```diff
@@ -49,43 +56,26 @@ func (s *ServiceAccounts) fetchServiceAccount(nsName string, saName string) (str
...
+
+	t, err := tokenMgr.GetServiceAccountToken(nsName, saName, &authenticationv1.TokenRequest{
+		Spec: authenticationv1.TokenRequestSpec{
+			ExpirationSeconds: getInt64Point(5000), // 5 minutes
+		},
+	})
 	if err != nil {
-		return "", fmt.Errorf("Getting service account: %s", err)
+		return "", fmt.Errorf("Failed to get service account token: %s", err)
```
cppforlife: nit: dont commonly include "Failed" in the error itself (it's an error, so it did indeed fail)

pkg/deploy/service_accounts.go
```diff
@@ -49,43 +56,26 @@ func (s *ServiceAccounts) fetchServiceAccount(nsName string, saName string) (str
 		return "", fmt.Errorf("Internal inconsistency: Expected service account name to not be empty")
 	}
 
-	sa, err := s.coreClient.CoreV1().ServiceAccounts(nsName).Get(context.Background(), saName, metav1.GetOptions{})
+	tokenMgr := token.NewManager(s.coreClient, s.log)
+
+	t, err := tokenMgr.GetServiceAccountToken(nsName, saName, &authenticationv1.TokenRequest{
+		Spec: authenticationv1.TokenRequestSpec{
+			ExpirationSeconds: getInt64Point(5000), // 5 minutes
```
cppforlife: a deploy might taken 20min to run as an example, i think our tokens needs to be longer e.g. 1 hour.

pkg/deploy/service_accounts.go
```diff
@@ -49,43 +56,26 @@ func (s *ServiceAccounts) fetchServiceAccount(nsName string, saName string) (str
...
-		if err != nil {
-			return "", fmt.Errorf("Getting service account secret: %s", err)
-		}
-
-		if secret.Type != corev1.SecretTypeServiceAccountToken {
-			continue
-		}
-
-		return s.buildKubeconfig(secret)
+	cert, err := os.ReadFile(caCertPath)
```
cppforlife: i dont believe cert is changing so we can in theory cache it after reading it once.

pkg/deploy/service_accounts.go
cppforlife: nit: instead of having a method getInt64Point you can just set variable to 5000 and take a pointer of it.

pkg/token/token_manager.go
```diff
@@ -0,0 +1,151 @@
+// Copyright 2022 VMware, Inc.
```
cppforlife: can we rename this package to `satoken`, token just seems way too generic.

pkg/deploy/service_accounts.go
neil-hickey: yup, I am wondering where it makes sense to have this, perhaps all the way back up the call stack.

Both AppCR and PKGR called `deploy.NewFactory()` so they would both get a different cache, I think having a shared cache is fine.

pkg/deploy/service_accounts.go
neil-hickey: hmm, I suppose if one ever had to rotate the kube-api cert that all deployments would get restarted anyways? Thus we would re-read from disk in that case.

pkg/satoken/token_manager.go
```diff
@@ -0,0 +1,151 @@
...
+	defer m.cacheMutex.Unlock()
+	m.cache[key] = tr
+}
+
+func (m *Manager) expired(t *authenticationv1.TokenRequest) bool {
+	return m.clock.Now().After(t.Status.ExpirationTimestamp.Time)
+}
+
+// requiresRefresh returns true if the token is older than 80% of its total
+// ttl, or if the token is older than 24 hours.
```
joe-kimmel-vmw: minor but should update the '24' here to be whatever our token lifetime is.

pkg/satoken/token_manager.go
```diff
@@ -0,0 +1,138 @@
...
+func NewManager(c clientset.Interface, log logr.Logger) *Manager {
+	m := &Manager{
+		getToken: func(name, namespace string, tr *authenticationv1.TokenRequest) (*authenticationv1.TokenRequest, error) {
+			return c.CoreV1().ServiceAccounts(namespace).CreateToken(context.TODO(), name, tr, metav1.CreateOptions{})
+		},
+		cache: make(map[string]*authenticationv1.TokenRequest),
+		clock: clock.RealClock{},
+		log:   log,
+	}
+	go wait.Forever(m.cleanup, gcPeriod)
```
neil-hickey, line 40: So, I left this in here - but increased the periodic time to a day for garbage collection

I suspect the only time this will be needed is when a service account is deleted / renamed by the user and we still maybe have it in our cache. There is the potential for this to grow without bounds, so I think having this periodic cleanup makes sense in that case.

@cppforlife

pkg/deploy/service_accounts.go
```diff
@@ -49,43 +61,27 @@ func (s *ServiceAccounts) fetchServiceAccount(nsName string, saName string) (str
...
+		},
+	})
 	if err != nil {
-		return "", fmt.Errorf("Getting service account: %s", err)
+		return "", fmt.Errorf("Get service account token: %s", err)
 	}
 
-	for _, secretRef := range sa.Secrets {
-		secret, err := s.coreClient.CoreV1().Secrets(nsName).Get(context.Background(), secretRef.Name, metav1.GetOptions{})
+	if len(s.caCert) == 0 {
```
cppforlife, line 78: needs to be behind mutex since parallel calls to Find may/will happen. alternatively can move it into constructor and record err (to be returned from Find) to avoid rereading cert.

pkg/deploy/service_accounts.go
```diff
@@ -4,21 +4,33 @@
...
+	"github.com/go-logr/logr"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/satoken"
+	authenticationv1 "k8s.io/api/authentication/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
+const (
+	caCertPath      = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+	tokenExpiration = time.Hour * 2
+)
```
cppforlife: can move these consts inside the function. (since it's only needed by that one func)

pkg/satoken/token_manager_test.go
```diff
@@ -0,0 +1,229 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// This file is a modified version of
+// https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/token/token_manager_test.go
+
+package satoken
```
cppforlife, line 7: not sure if we can move this into satoken_test

pkg/deploy/service_accounts.go
neil-hickey: `tokenExpiration` feels like a magic number, so I like having it as a constant. The cert location is used once, so I'm fine with that being inlined.

pkg/satoken/token_manager_test.go
neil-hickey: Lemme give it a shot

neil-hickey: We mock out `getToken` which is a private function of `Manager`. Mocking out the call to k8s is more complicated if we go the `_test` package route.

We'd also have to make the `clock` public which should not be public.

pkg/deploy/service_accounts.go
cppforlife: im not suggesting it to be inlined, just be const inside the function e.g.

```
func (Foo) something() {
   const (
     x = 123
   )
   ...
} 
```

PR #697, opened 2022-05-21
move some App CR reconciliation parts into sidecar

cppforlife:

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```


Dockerfile
```diff
@@ -31,6 +31,7 @@ RUN curl -sLo cue.tgz https://github.com/cue-lang/cue/releases/download/v0.4.2/c
 COPY . .
 # helpful ldflags reference: https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications
 RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags="-X 'main.Version=$KCTRL_VER'" -trimpath -o controller ./cmd/main.go
+RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags="-X 'main.Version=$KCTRL_VER'" -trimpath -o controller-sidecarexec ./pkg/sidecarexec/cmd/main.go
```
benmoss: can we clean up toplevel `cmd` to have `cmd/controller/main.go` and `cmd/sidecar/main.go` or whatever? having mains under pkg seems odd

cmd/controller/run.go
```diff
@@ -147,15 +151,24 @@ func Run(opts Options, runLog logr.Logger) error {
 		if err != nil {
 			return fmt.Errorf("Setting up Config reconciler watches: %s", err)
 		}
+
+		// Reconcile once synchronously to ensure configuration is loaded before any execution happens
```
benmoss: I don't understand this comment, why do we have to do this now?

cmd/controller/run.go
```diff
@@ -237,17 +250,3 @@ func Run(opts Options, runLog logr.Logger) error {
 
 	return nil
 }
-
-func logProxies(runLog logr.Logger) {
```
benmoss, line 241: Unrelated change?

cmd/main.go
```diff
@@ -40,16 +36,13 @@ func main() {
...
+	if err != nil {
+		mainLog.Error(err, "Exited run with error")
+		os.Exit(1)
+		panic("unreachable: controller exited 1")
 	}
 
-	controllerinit.Run(os.Args[0], os.Args[1:], log.WithName("init"))
-	panic("unreachable: init proc returned")
+	os.Exit(0)
+	panic("unreachable: controller exited 0")
```
benmoss: what is with this unreachable code stuff and the Exit(0) call?

config/deployment.yml
```diff
@@ -49,6 +51,31 @@ spec:
...
+            memory: 100Mi
+        env:
+        - name: KAPPCTRL_SIDECAREXEC_SOCK
+          value: /etc/kappctrl-mem-tmp/sidecarexec.sock
+        volumeMounts:
+        - name: template-fs
+          mountPath: /etc/kappctrl-mem-tmp
+        - name: home
+          mountPath: /home/kapp-controller
+        #! mount over service account token to make it not available
```
benmoss, line 76: interesting hack

pkg/config/config.go
```diff
@@ -7,56 +7,30 @@ import (
 	"context"
```
benmoss, line 7: Why did all this need to change rather than just being called by the sidecar instead of the controller?

benmoss (commented):
It's an interesting idea, and a step in the right direction at least in terms of isolating the control plane from the semi-user-invoked CLIs.

Dockerfile
cppforlife: yeah, absolutely. didnt want to push it in this PR, but happy to. or as a followup.

cmd/controller/run.go
cppforlife: since sidecar can be configured with ca certs / proxies we want to ensure that that happens before actual execution. any suggestions on rephrasing this comment?

cmd/controller/run.go
cppforlife: that got moved into sidecar itself since only sidecar needs to be configured with proxies.

cmd/main.go
cppforlife: yeah... thats just weirdness... im ok removing it.

config/deployment.yml
cppforlife: the only way i could find how to "remove" sa token from one of the containers. this same trick is used by agones for game servers.

pkg/config/config.go
cppforlife: sidecar doesnt have access to k8s client. it only exposes api how to configure ca cert / proxies. Config object got simplified to just be a thing that has info, vs thing that knows how to apply it.

config/deployment.yml
praveenrewar: 🤯

cmd/main.go
cppforlife: removed.

cmd/controller/run.go
cppforlife: updated comment with a little bit more info.

Dockerfile
cppforlife: did it in this PR.

benmoss (approved):

PR #708, opened 2022-05-26
Use UID of ServiceAccount for token cache

### LINKED ISSUE #701: Kapp controller doesnt generate new token for a SA with the same name as a deleted SA
**What steps did you take:**
1. Deploy kapp controller
2. Apply a Package, PackageInstall, ServiceAccount (used in PackageInstall as .spec.serviceAccountName)
3. Delete the ServiceAccount
4. Recreate the ServiceAccount with the same name

**What happened:**
PackageInstall  fails as it get the token for the old ServiceAccount which is now invalid.

**What did you expect:**
A new token to be generated for this new ServiceAccount.

Looks like it was due to the key of the TokenManager cache being non-unique across all ServiceAccounts across the lifetime of the cluster.  Two ways around this are using something that is unique (UID of ServiceAccount) or use the TokenReview API which does track whether the ServiceAccount is new or not.

**Environment:**

- Any branch including https://github.com/vmware-tanzu/carvel-kapp-controller/pull/695.  
- Will hit on Kubernetes 1.24.0 and above where ServiceAccounts no longer autogenerate tokens and the TokenRequest API is used

---
Vote on this request

neil-hickey:
- Handles the case where a SA is created, deleted and recreated

Fixes #701

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/satoken/token_manager.go
+++ b/pkg/satoken/token_manager.go
@@ -34,9 +34,6 @@ func NewManager(c clientset.Interface, log logr.Logger) *Manager {
 		getToken: func(name, namespace string, tr *authenticationv1.TokenRequest) (*authenticationv1.TokenRequest, error) {
 			return c.CoreV1().ServiceAccounts(namespace).CreateToken(context.TODO(), name, tr, metav1.CreateOptions{})
 		},
-		reviewToken: func(tr *authenticationv1.TokenReview) (*authenticationv1.TokenReview, error) {
-			return c.AuthenticationV1().TokenReviews().Create(context.TODO(), tr, metav1.CreateOptions{})
-		},
 		cache: make(map[string]*authenticationv1.TokenRequest),
 		clock: clock.RealClock{},
 		log:   log,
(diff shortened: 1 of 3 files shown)


neil-hickey:
I couldn't see a nicer way than another k8s call to check the token status.

Even if we changed our cache to use the unique key of the SA, it would require a call to the TokenReview endpoint, we don't get it for free as is. So this is just as many calls, and slightly simpler.

Any other options are highly welcomed! For now, this is the best solution I can think of for the use case of `create SA, delete SA, re-create SA with same name`

pkg/satoken/token_manager.go
```diff
@@ -118,6 +122,16 @@ func (m *Manager) expired(t *authenticationv1.TokenRequest) bool {
 
 // requiresRefresh returns true if the token is older half of it's maxTTL
 func (m *Manager) requiresRefresh(tr *authenticationv1.TokenRequest) bool {
+	review, err := m.reviewToken(&authenticationv1.TokenReview{
+		Spec: authenticationv1.TokenReviewSpec{
+			Token: tr.Status.Token,
+		},
+	})
```
cppforlife: x-post from slack:

> i think additional call is a pretty expensive solution. any reason why we cant just use metadata.uid of service account in the cache key?

PR #715, opened 2022-05-31
Bunch of UX fixes tweaking output

### LINKED ISSUE #676: `kctrl`: UX enhancements
A bunch of snacky tweaks that will enhance UX. These might be resolved across multiple PRs.

Listing AIs in the acceptance criteria

- [x] When a reconciliation completes we should be logging `Deploy complete` instead of `App reconciled to stay consistent with the other stages
- [x] Colour app list status column
- [x] Colour package installed list status output column
- [x] ~~Conditions in get shouldnt include reason/message when empty (check for missing omitempty tags)~~ (Relevant PR cppforlife/go-cli-ui#11)
- [x] Handle invalid package names better (example: `kctrl pkg i -i contour -p foo`) 
- [ ] Explore if we need to leverage wordwrapping for certain columns in list and get commands. This can diverge into a separate issue if it warrants it.

Adding a few more to the list:
- [ ] Add shorthand `-v` for `--version` in package installed command.
- [x] Add namespace name to not found errors.
- [x] Update `package installed list` example to use `-A` instead of `A`.
- [x] Update error message to be more user friendly when `values` and `values-file-output` flag are used for a package install with no values.yml
- [x] Update error message for `package available get -p <pkg-name> --values-schema to include information about how to specify a version`

---
Vote on this request

100mik:
Fixes some kctrl UX issues

Fixes (parts of) #676 

#### Release notes for user facing features:
```
A bunch of UX improvements for kctrl
```

Adds:
- Consistent status coloring across commands
- Corrects misspelt example

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/cli/pkg/kctrl/cmd/app/app.go
+++ b/cli/pkg/kctrl/cmd/app/app.go
@@ -4,10 +4,10 @@
 package app
 
 import (
-	"github.com/cppforlife/color"
 	"github.com/spf13/cobra"
 	kcv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
 	kcpkgv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
 )
 
 func NewCmd() *cobra.Command {
@@ -33,19 +33,19 @@ func appStatusString(app *kcv1alpha1.App) string {
 		return ""
 	}
 	if app.Spec.Canceled {
-		return color.RedString("Canceled")
+		return "Canceled"
 	}
 	if app.Spec.Paused {
-		return color.YellowString("Paused")
+		return "Paused"
 	}
 	for _, condition := range app.Status.Conditions {
 		switch condition.Type {
 		case kcv1alpha1.ReconcileFailed:
-			return color.RedString("Reconcile failed")
+			return "Reconcile failed"
 		case kcv1alpha1.ReconcileSucceeded:
-			return color.GreenString("Reconcile succeeded")
+			return "Reconcile succeeded"
 		case kcv1alpha1.DeleteFailed:
-			return color.RedString("Deletion failed")
+			return "Deletion failed"
 		case kcv1alpha1.Reconciling:
 			return "Reconciling"
 		case kcv1alpha1.Deleting:
@@ -53,4 +53,13 @@ func appStatusString(app *kcv1alpha1.App) string {
 		}
 	}
 	return app.Status.FriendlyDescription
+}
+
+func isFailing(conditions []kcv1alpha1.AppCondition) bool {
+	for _, condition := range conditions {
+		if condition.Type == kcv1alpha1.ReconcileFailed && condition.Status == corev1.ConditionTrue {
+			return true
+		}
+	}
+	return false
 }
--- a/cli/pkg/kctrl/cmd/app/app.go
+++ b/cli/pkg/kctrl/cmd/app/app.go
@@ -33,7 +33,7 @@ func appStatusString(app *kcv1alpha1.App) string {
 		return ""
 	}
 	if app.Spec.Canceled {
-		return color.RedString("Camceled")
+		return color.RedString("Canceled")
 	}
 	if app.Spec.Paused {
 		return color.YellowString("Paused")
--- a/cli/pkg/kctrl/cmd/app/app.go
+++ b/cli/pkg/kctrl/cmd/app/app.go
@@ -7,7 +7,6 @@ import (
 	"github.com/spf13/cobra"
 	kcv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
 	kcpkgv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
-	corev1 "k8s.io/api/core/v1"
 )
 
 func NewCmd() *cobra.Command {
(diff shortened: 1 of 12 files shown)


cli/pkg/kctrl/cmd/app/app.go
```diff
@@ -29,7 +28,29 @@ func isOwnedByPackageInstall(app *kcv1alpha1.App) bool {
...
+	if app.Spec.Canceled {
+		return color.RedString("Camceled")
+	}
+	if app.Spec.Paused {
+		return color.YellowString("Paused")
+	}
+	for _, condition := range app.Status.Conditions {
+		switch condition.Type {
+		case kcv1alpha1.ReconcileFailed:
+			return color.RedString("Reconcile failed")
```
cppforlife: lets use ValueFmt (https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/cmd/app/list.go#L139)

100mik: This would only allow us to color the output red if the Error boolean is true right?
As of now we are also coloring paused status yellow, that won't be possible with ValueFmt I think.

I see there is the possibility of having a "Func" formatting the value but it looks like the contructor does not allow for it

100mik: Will give it a shot after creating the struct directly, however the information in the status does not differentiate between Cancelled and paused statuses, might be a hacky use of ValueFmt

cli/pkg/kctrl/cmd/app/app.go
```diff
@@ -29,7 +28,29 @@ func isOwnedByPackageInstall(app *kcv1alpha1.App) bool {
...
 
-func getAppDescription(name string, namespace string) string {
-	description := fmt.Sprintf("app/%s (kappctrl.k14s.io/v1alpha1)  namespace: %s", name, namespace)
-	return description
+func appStatusString(app *kcv1alpha1.App) string {
+	if len(app.Status.Conditions) < 1 {
+		return ""
+	}
+	if app.Spec.Canceled {
+		return color.RedString("Camceled")
```
cppforlife: typo.

cli/pkg/kctrl/cmd/app/app.go
100mik: I was giving this a shot, but I think it does not fit the case well here since we are not picking up a value directly from the status (which we would assign to `ValueFmt.V`)
If the app is paused or cancelled, we show "Paused" or "Cancelled" instead of formatting the friendly error message that is "Canceled/Paused"

cppforlife: > I was giving this a shot, but I think it does not fit the case well here since we are not picking up a value directly from the status (which we would assign to ValueFmt.V)

im not following.

ValueFmt takes arbitrary value. example: https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/cmd/app/list.go#L139-L142

V of ValueFmt could be anything including string value, any way you decide to construct. there is no relationship to where the value comes from. whatever logic you have today, you can construct a string value and give it to ValueFmt.

cli/pkg/kctrl/cmd/app/get.go
```diff
@@ -85,7 +85,9 @@ func (o *GetOptions) Run() error {
 			uitable.NewValueString(app.Namespace),
 			uitable.NewValueString(app.Name),
 			uitable.NewValueString(app.Spec.ServiceAccountName),
-			uitable.NewValueString(app.Status.FriendlyDescription),
+			uitable.ValueFmt{
+				V: uitable.NewValueString(appStatusString(app)),
```
cppforlife: should have Error key set to true when it's a "failure".

cli/pkg/kctrl/cmd/app/list.go
```diff
@@ -91,7 +91,11 @@ func (o *ListOptions) Run() error {
 		table.Rows = append(table.Rows, []uitable.Value{
 			cmdcore.NewValueNamespace(app.Namespace),
 			uitable.NewValueString(app.Name),
-			uitable.NewValueString(app.Status.FriendlyDescription),
+			uitable.ValueFmt{
+				V:     uitable.NewValueString(appStatusString(&app)),
+				Error: isFailing(app.Status.Conditions) || app.Spec.Canceled,
+			},
+			// uitable.NewValueString(appStatusString(&app)),
```
praveenrewar: nit: remove comment

cli/pkg/kctrl/cmd/app/app.go
```diff
@@ -29,7 +28,38 @@ func isOwnedByPackageInstall(app *kcv1alpha1.App) bool {
 	return false
 }
 
-func getAppDescription(name string, namespace string) string {
-	description := fmt.Sprintf("app/%s (kappctrl.k14s.io/v1alpha1)  namespace: %s", name, namespace)
-	return description
+func appStatusString(app *kcv1alpha1.App) string {
+	if len(app.Status.Conditions) < 1 {
```
cppforlife: == 0?

cli/pkg/kctrl/cmd/app/app.go
```diff
@@ -29,7 +28,38 @@ func isOwnedByPackageInstall(app *kcv1alpha1.App) bool {
 	return false
 }
 
-func getAppDescription(name string, namespace string) string {
-	description := fmt.Sprintf("app/%s (kappctrl.k14s.io/v1alpha1)  namespace: %s", name, namespace)
-	return description
+func appStatusString(app *kcv1alpha1.App) string {
```
cppforlife: one thing to consider if this function should return ValueFmt directly. that way you dont have to calculate Error bool externally. alternatively you can also return (string, bool) and use ValueFmt.

cli/pkg/kctrl/cmd/app/get.go
```diff
@@ -85,7 +84,10 @@ func (o *GetOptions) Run() error {
 			uitable.NewValueString(app.Namespace),
 			uitable.NewValueString(app.Name),
 			uitable.NewValueString(app.Spec.ServiceAccountName),
-			uitable.NewValueString(app.Status.FriendlyDescription),
+			uitable.ValueFmt{
+				V:     uitable.NewValueString(appStatusString(app)),
+				Error: isFailing || app.Spec.Canceled,
```
cppforlife: it's interesting that we are using spec section. should we also check on paused?

100mik: Mainly because an app being paused or cancelled does not affect the conditions and even the friendly message is ambiguous ("Canceled/paused"). For pkgi there are no changes to status at all.
I was initially coloring "Paused" as yellow but the table formatting was not happy about that.

A bit unsure as to if pause should be colored red as well 🤔 but we can definitely make that change.

cli/pkg/kctrl/cmd/app/app.go
100mik: Yep, that sounds better putting it in!

cli/pkg/kctrl/cmd/app/app.go
100mik: I think returning (string, bool) makes more sense (in case this ends up being used someplace else.

PS: Me and praveen have been meaning to create an issue to move stuff common to Pkgi and app like waiting logic, and status checks to be scoped to respective objects, with the primary intention of deducing. I guess keeping it generic and not returning a ValueFmt object makes more sense in that scenario 🤔

100mik: Changed up the function to return (string, bool)
And the bool is used to determine failure status

PR #725, opened 2022-06-07
`kctrl`:  ytt overlay flags

### LINKED ISSUE #674: `kctrl`: Allow users to use ytt overlay while creating package installs easily
Allow users to supply an overlay file or inline (simiolar to how we handle values file) that can be fed into the `ext.packaging.carvel.dev/ytt-paths-from-secret-name` annotation.

A flag like `--overlay` which can be used multiple times (similar to how `kapp` accepts multiple `-f` flag) which can be used to create multiple overlaying annotations.

**Acceptance criteria:**
- [x] Creates secrets with ownership annotations using overlays provided by the user
- [x] These secrets should be added as overlays via annotation
- [x] They should be garbage collected while uninstalling
- [x] Tests

**Open questions**
- Should we allow multiple overlays? Does this help users organise their overlays better?
- Should we create one secret with multiple file keys? 
- How do we handle updates? Do we merge if only one file is provided? Or do we clobber and overwrite the list of overlays each time? (not merging would be consistent with how we handle values secrets today)

---
Vote on this request

100mik:
Add flags which allow overlaying of package content while creating package installs.

Fixes #674 

```release-note
Introduce ytt overlay flags
- `--ytt-overlays` (boolean) decides whether or not ytt overlays are added/kept.
- `--ytt-overlay-file` files can be used to supply ytt overlay file paths or paths to directories wioth ytt overlays.
```

##### Review Checklist:

  a link to that PR vmware-tanzu/carvel#558
  change

```docs

```

### DIFF
--- a/cli/pkg/kctrl/cmd/package/installed/create_or_update.go
+++ b/cli/pkg/kctrl/cmd/package/installed/create_or_update.go
@@ -304,8 +304,13 @@ func (o *CreateOrUpdateOptions) RunUpdate(args []string) error {
 		return fmt.Errorf("Expected package install to be non-empty")
 	}
 
-	if len(o.version) == 0 && len(o.valuesFile) == 0 && o.values &&
-		o.YttOverlayFlags.yttOverlays && len(o.YttOverlayFlags.yttOverlayFiles) == 0 {
+	switch {
+	case len(o.version) != 0: // Continue if version is changed
+	case len(o.valuesFile) != 0: // Continue if supplied values might have changed
+	case !o.values: // Continue if supplied values might be getting removed
+	case !o.YttOverlayFlags.yttOverlays: // Continue if supplied overlays might be getting removed
+	case len(o.YttOverlayFlags.yttOverlayFiles) != 0: // Continue if supplied overlays might be getting changed
+	default:
 		return fmt.Errorf("Expected either package version, values file or overlays to update the package")
 	}
 
@@ -346,8 +351,13 @@ func (o CreateOrUpdateOptions) update(client kubernetes.Interface, kcClient kccl
 		return err
 	}
 
-	if o.valuesFile == "" && !changed && o.values &&
-		o.YttOverlayFlags.yttOverlays && len(o.YttOverlayFlags.yttOverlayFiles) == 0 {
+	switch {
+	case changed: // Continue if package install resource is changed
+	case len(o.valuesFile) != 0: // Continue if supplied values might have changed
+	case !o.values: // Continue if supplied values might be getting removed
+	case !o.YttOverlayFlags.yttOverlays: // Continue if supplied overlays might be getting removed
+	case len(o.YttOverlayFlags.yttOverlayFiles) != 0: // Continue if supplied overlays might be getting changed
+	default:
 		o.statusUI.PrintMessagef("No changes to package install '%s' in namespace '%s'", o.Name, o.NamespaceFlags.Name)
 		return nil
 	}
--- a/cli/pkg/kctrl/cmd/package/installed/create_or_update.go
+++ b/cli/pkg/kctrl/cmd/package/installed/create_or_update.go
@@ -34,7 +34,7 @@ import (
 const (
 	valuesFileKey        = "values.yaml"
 	yttOverlayPrefix     = "ext.packaging.carvel.dev/ytt-paths-from-secret-name"
-	yttOverlayAnnotation = yttOverlayPrefix + ".kctrl-overlay-secret"
+	yttOverlayAnnotation = yttOverlayPrefix + ".kctrl-ytt-overlays"
 )
 
 type CreateOrUpdateOptions struct {
--- a/cli/pkg/kctrl/cmd/package/installed/create_or_update.go
+++ b/cli/pkg/kctrl/cmd/package/installed/create_or_update.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/cppforlife/go-cli-ui/ui"
@@ -31,7 +32,8 @@ import (
 )
 
 const (
-	valuesFileKey = "values.yaml"
+	valuesFileKey        = "values.yaml"
+	yttOverlayAnnotation = "ext.packaging.carvel.dev/ytt-paths-from-secret-name.kctrl"
 )
 
 type CreateOrUpdateOptions struct {
@@ -40,7 +42,8 @@ type CreateOrUpdateOptions struct {
 	depsFactory cmdcore.DepsFactory
 	logger      logger.Logger
 
-	WaitFlags cmdcore.WaitFlags
+	WaitFlags       cmdcore.WaitFlags
+	YttOverlayFlags YttOverlayFlags
 
 	packageName        string
 	version            string
@@ -101,6 +104,7 @@ func NewCreateCmd(o *CreateOrUpdateOptions, flagsFactory cmdcore.FlagsFactory) *
 		DefaultInterval:  1 * time.Second,
 		DefaultTimeout:   30 * time.Minute,
 	})
+	o.YttOverlayFlags.Set(cmd)
 
 	return cmd
 }
@@ -144,6 +148,7 @@ func NewInstallCmd(o *CreateOrUpdateOptions, flagsFactory cmdcore.FlagsFactory) 
 		DefaultInterval:  1 * time.Second,
 		DefaultTimeout:   30 * time.Minute,
 	})
+	o.YttOverlayFlags.Set(cmd)
 
 	return cmd
 }
@@ -185,6 +190,7 @@ func NewUpdateCmd(o *CreateOrUpdateOptions, flagsFactory cmdcore.FlagsFactory) *
 		DefaultInterval:  1 * time.Second,
 		DefaultTimeout:   30 * time.Minute,
 	})
+	o.YttOverlayFlags.Set(cmd)
 
 	return cmd
 }
@@ -263,8 +269,17 @@ func (o *CreateOrUpdateOptions) create(client kubernetes.Interface, kcClient kcc
 		return err
 	}
 
+	overlaySecretName := ""
+	if o.YttOverlayFlags.yttOverlays {
+		overlaysSecret, err := o.createOrUpdateYttOverlaySecrets(nil, client)
+		if err != nil {
+			return err
+		}
+		overlaySecretName = overlaysSecret.Name
+	}
+
 	o.statusUI.PrintMessagef("Creating package install resource")
-	if err = o.createPackageInstall(isServiceAccountCreated, isSecretCreated, kcClient); err != nil {
+	if err = o.createPackageInstall(isServiceAccountCreated, isSecretCreated, overlaySecretName, kcClient); err != nil {
 		return err
 	}
 
@@ -327,14 +342,17 @@ func (o CreateOrUpdateOptions) update(client kubernetes.Interface, kcClient kccl
 		return err
 	}
 
-	if o.valuesFile == "" && !changed && o.values {
+	if o.valuesFile == "" && !changed && o.values &&
+		o.YttOverlayFlags.yttOverlays && len(o.YttOverlayFlags.yttOverlayFiles) == 0 {
 		o.statusUI.PrintMessagef("No changes to package install '%s' in namespace '%s'", o.Name, o.NamespaceFlags.Name)
 		return nil
 	}
 
 	// Pause reconciliation so that kctrl can tail status if an existing values secret is updated
+	// Or when existing ytt overlay has to be updated
 	reconciliationPaused := false
-	if o.valuesFile != "" && len(pkgInstall.Spec.Values) > 0 {
+	if (o.valuesFile != "" && len(pkgInstall.Spec.Values) > 0) ||
+		(len(o.YttOverlayFlags.yttOverlayFiles) > 0 && o.hasYttOverlays(pkgInstall)) {
 		updatedPkgInstall, err = o.pauseReconciliation(kcClient)
 		if err != nil {
 			return err
@@ -368,6 +386,25 @@ func (o CreateOrUpdateOptions) update(client kubernetes.Interface, kcClient kccl
 		if !changed {
 			return nil
 		}
+	}
+
+	if o.YttOverlayFlags.yttOverlays {
+		overlaysSecret, err := o.createOrUpdateYttOverlaySecrets(updatedPkgInstall, client)
+		if err != nil {
+			return err
+		}
+		if overlaysSecret != nil {
+			updatedPkgInstall.Annotations[yttOverlayAnnotation] = overlaysSecret.Name
+			changed = true
+		}
+	}
+
+	if !o.YttOverlayFlags.yttOverlays {
+		err = o.dropYttOverlaySecrets(updatedPkgInstall, client)
+		if err != nil {
+			return err
+		}
+		o.removeYttOverlaysAnnotation(updatedPkgInstall)
 	}
 
 	if isSecretCreated || changed {
@@ -602,7 +639,7 @@ func (o *CreateOrUpdateOptions) createOrUpdateDataValuesSecret(client kubernetes
 	return true, nil
 }
 
-func (o *CreateOrUpdateOptions) createPackageInstall(serviceAccountCreated, secretCreated bool, kcClient kcclient.Interface) error {
+func (o *CreateOrUpdateOptions) createPackageInstall(serviceAccountCreated, secretCreated bool, overlaysSecretName string, kcClient kcclient.Interface) error {
 	svcAccount := o.serviceAccountName
 	if svcAccount == "" {
 		svcAccount = o.createdAnnotations.ServiceAccountAnnValue()
@@ -632,6 +669,11 @@ func (o *CreateOrUpdateOptions) createPackageInstall(serviceAccountCreated, secr
 				},
 			},
 		}
+	}
+
+	// Add reference to ytt overlaty annotation if overlay secret has been created
+	if overlaysSecretName != "" {
+		packageInstall.Annotations[yttOverlayAnnotation] = overlaysSecretName
 	}
 
 	o.addCreatedResourceAnnotations(&packageInstall.ObjectMeta, serviceAccountCreated, secretCreated, false)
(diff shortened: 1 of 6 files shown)


cli/pkg/kctrl/cmd/package/installed/ytt_overlay_flags.go
```diff
@@ -0,0 +1,16 @@
...
+
+import "github.com/spf13/cobra"
+
+type YttOverlayFlags struct {
+	yttOverlays     bool
+	yttOverlayFiles []string
+}
+
+func (s *YttOverlayFlags) Set(cmd *cobra.Command) {
+	cmd.Flags().StringSliceVar(&s.yttOverlayFiles, "ytt-overlay-file", nil, "A file or folder with ytt overlays to be applied to package install")
```
praveenrewar:
```suggestion
	cmd.Flags().StringSliceVar(&s.yttOverlayFiles, "ytt-overlay-file", nil, "Path to ytt overlay file")
```

cli/test/e2e/ytt_overlays_test.go
```diff
@@ -0,0 +1,179 @@
...
+			StdinReader: strings.NewReader(overlay1),
+		})
+		require.NoError(t, err)
+		//Ensure that progress is tailed
+		require.Contains(t, out, "Fetch succeeded")
+		require.Contains(t, out, "Template succeeded")
+		require.Contains(t, out, "Deploy succeeded")
+
+		out = kubectl.Run([]string{"get", "deployment", deploymentName, "-o", "yaml"})
+		require.Contains(t, out, "foo")
```
praveenrewar:
```suggestion
		require.Contains(t, out, "test: foo")
```

cli/pkg/kctrl/cmd/package/installed/delete.go
```diff
@@ -255,6 +267,17 @@ func (o *DeleteOptions) cleanUpIfInstallNotFound(dynamicClient dynamic.Interface
 		return err
 	}
 
+	//Delete ytt overlays secret created by kctrl if any
+	err = o.deleteIfExistsAndOwned(
```
praveenrewar, line 272: Should we also include a test for this in the e2e test?

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -353,6 +374,27 @@ func (o CreateOrUpdateOptions) update(client kubernetes.Interface, kcClient kccl
...
+			return err
+		}
+		_, found := updatedPkgInstall.Annotations[yttOverlayAnnotation]
+		if !found && overlaysSecret != nil {
+			updatedPkgInstall.Annotations[yttOverlayAnnotation] = overlaysSecret.Name
+			changed = true
+		}
+	}
+
+	if !o.YttOverlayFlags.yttOverlays {
```
praveenrewar: Seems like an `else` to previous if condition

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -31,7 +32,8 @@ import (
 )
 
 const (
-	valuesFileKey = "values.yaml"
+	valuesFileKey        = "values.yaml"
+	yttOverlayAnnotation = "ext.packaging.carvel.dev/ytt-paths-from-secret-name.kctrl.carvel.dev"
```
praveenrewar: yttOverlayPrefix + ".kctrl.carvel.dev"?

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -40,7 +42,8 @@ type CreateOrUpdateOptions struct {
 	depsFactory cmdcore.DepsFactory
 	logger      logger.Logger
 
-	WaitFlags cmdcore.WaitFlags
+	WaitFlags       cmdcore.WaitFlags
+	YttOverlayFlags YttOverlayFlags
```
praveenrewar, line 47: Maybe group it with valuesFile instead?

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -274,8 +291,9 @@ func (o *CreateOrUpdateOptions) RunUpdate(args []string) error {
 		return fmt.Errorf("Expected package install to be non-empty")
 	}
 
-	if len(o.version) == 0 && len(o.valuesFile) == 0 && o.values {
-		return fmt.Errorf("Expected either package version or values file to update the package")
+	if len(o.version) == 0 && len(o.valuesFile) == 0 && o.values &&
+		o.YttOverlayFlags.yttOverlays && len(o.YttOverlayFlags.yttOverlayFiles) == 0 {
```
praveenrewar: It feels like this is getting more and more complex. Should we break this into bits? Something like:
```go
hasValuesFile := o.values && len(o.valuesFile) > 0
hasYttOverlayFiles := o.YttOverlayFlags.yttOverlays && len(o.YttOverlayFlags.yttOverlayFiles) > 0

if len(o.version) == 0 && !hasValuesFile && !hasYttOverlayFiles {}
```

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -274,8 +291,9 @@ func (o *CreateOrUpdateOptions) RunUpdate(args []string) error {
...
+		return fmt.Errorf("Expected either package version ,values file or overlays to update the package")
```
praveenrewar:
```suggestion
		return fmt.Errorf("Expected either package version, values file or overlays to update the package")
```

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -957,6 +1006,78 @@ func (o *CreateOrUpdateOptions) showVersions(client pkgclient.Interface) error {
...
+	if err != nil {
+		return fmt.Errorf("Deleting overlays secret: %s", err)
+	}
+
+	return nil
+}
+
+func (o *CreateOrUpdateOptions) removeYttOverlaysAnnotation(pkgi *kcpkgv1alpha1.PackageInstall) {
+	delete(pkgi.Annotations, yttOverlayAnnotation)
+}
```
praveenrewar, line 1103: Maybe we can do this inline instead of separate function?

cli/test/e2e/ytt_overlays_test.go
100mik: Yep this makes sense I will make this change

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
100mik: I did that initially but I thought this helps add context

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
100mik: Will make this change

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
100mik: I was planning on leavinga comment instead

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
100mik: Yep!
I believe I was gonna add a hasYttOverlays check here as well

cli/pkg/kctrl/cmd/package/installed/delete.go
100mik: Yep that can be done
I will add one

cli/pkg/kctrl/cmd/package/installed/ytt_overlay_flags.go
100mik: I think we need to specify that it can be a folder too, but we can make this better, I agree

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
praveenrewar: I would prefer a line of code instead of a comment, but no strong opinions :)

cli/pkg/kctrl/cmd/package/installed/ytt_overlay_flags.go
praveenrewar: > I think we need to specify that it can be a folder too

We usually don't do that, but if we want to, then how about `Path to ytt overlay file (can be a directory)`?

cli/pkg/kctrl/cmd/package/installed/ytt_overlays.go
```diff
@@ -0,0 +1,102 @@
...
+
+		if fileInfo.IsDir() {
+			err := filepath.Walk(file, func(path string, fi os.FileInfo, err error) error {
+				if err != nil || fi.IsDir() {
+					return err
+				}
+				ext := filepath.Ext(path)
+				for _, allowedExt := range fileResourcesAllowedExts {
+					if allowedExt == ext {
+						bytes, err := cmdcore.NewInputFile(path).Bytes()
```
100mik, line 83: This should be the relative path instead of the entire path to ensure that it is as short as it can be

100mik:
Right now a secret created by the command:
```
kctrl package installed update --ytt-overlay-file overlay-file.yaml --ytt-overlay-file /overlays-folder
```
Will look something like:
```
apiVersion: v1
kind: Secret
metadata:
  annotations:
    packaging.carvel.dev/package: test-default
  creationTimestamp: "2022-06-21T11:23:11Z"
  name: test-default-overlays
  namespace: default
  resourceVersion: "398587"
  uid: 4a29b5a5-65c4-407e-8cc8-fe1f2adaf7fd
type: Opaque
data:
  0000-overlay-file.yaml: I0AgbG9hZCgiQHl0dDpvd #truncated......
  0001-_Users_soumikm_Documents_playground_kctrloverlays_overlays-folder_a-overlay.yml: I0AgbG9hZCgiQHl0dDpvdm #truncated...
  0001-_Users_soumikm_Documents_playground_kctrloverlays_overlays-folder_b-overlay.yaml: I0AgbG9hZCgiQHl0dDpvdmVy #truncated...
```
One caveat could be folders with names which have characters that are not permitted in secret data keys.
Working on making sure the filepaths are as short as they need to be (relative file path)

cli/pkg/kctrl/cmd/package/installed/ytt_overlay_flags.go
100mik: Makes sense ! Adopted this

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
100mik: Added the additional check
It prevents us from making unnecesary update calls

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
100mik: Happy to switch this up if you think the other way is better, gonna keep it like this for now to prevent calling low level golang stuff from high level functions

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
100mik: Grouped it with WaitFlags as it is also a set of flags used across multiple commands
Happy to move it tho, no strong opinions

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
100mik: Made this change and moved the declaration for the prefix

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
praveenrewar: It just felt weird to have a one liner function. I don't have any strong opinions though.

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -31,7 +32,9 @@ import (
 )
 
 const (
-	valuesFileKey = "values.yaml"
+	valuesFileKey        = "values.yaml"
+	yttOverlayPrefix     = "ext.packaging.carvel.dev/ytt-paths-from-secret-name"
+	yttOverlayAnnotation = yttOverlayPrefix + ".kctrl.carvel.dev"
```
cppforlife: seems a bit weird imho to use `.` and generic `carvel.dev` suffix for this. may be something like `.kctrl-ytt-overlay`

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -274,8 +290,9 @@ func (o *CreateOrUpdateOptions) RunUpdate(args []string) error {
 		return fmt.Errorf("Expected package install to be non-empty")
 	}
 
-	if len(o.version) == 0 && len(o.valuesFile) == 0 && o.values {
-		return fmt.Errorf("Expected either package version or values file to update the package")
+	if len(o.version) == 0 && len(o.valuesFile) == 0 && o.values &&
```
cppforlife: i would use a switch statement:

```
switch {
case len(o.version): // do nothing
case len(o.valuesFile): // do nothing
...
default:
  return fmt.Errorf...
}
```

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -310,14 +327,17 @@ func (o CreateOrUpdateOptions) update(client kubernetes.Interface, kcClient kccl
...
 		o.statusUI.PrintMessagef("No changes to package install '%s' in namespace '%s'", o.Name, o.NamespaceFlags.Name)
 		return nil
 	}
 
 	// Pause reconciliation so that kctrl can tail status if an existing values secret is updated
+	// Or when existing ytt overlay has to be updated
 	reconciliationPaused := false
-	if o.valuesFile != "" && len(pkgInstall.Spec.Values) > 0 {
+	if (o.valuesFile != "" && len(pkgInstall.Spec.Values) > 0) ||
+		(len(o.YttOverlayFlags.yttOverlayFiles) > 0 && hasYttOverlays(pkgInstall)) {
```
cppforlife, line 369: dont we already have logic above (that switch statement)...

cli/test/e2e/ytt_overlays_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+	logger.Section("install with ytt overlays", func() {
+		out, err := kappCtrl.RunWithOpts([]string{"package", "install", "-i", pkgiName, "--package", packageName, "--version", packageVersion, "--ytt-overlay-file", "-"}, RunOpts{
+			StdinReader: strings.NewReader(overlay1),
+		})
+		require.NoError(t, err)
+		//Ensure that progress is tailed
+		require.Contains(t, out, "Fetch succeeded")
+		require.Contains(t, out, "Template succeeded")
+		require.Contains(t, out, "Deploy succeeded")
```
cppforlife: i dont see a need to assert on this output. package install should be exiting after install is complete.

cli/test/e2e/ytt_overlays_test.go
```diff
@@ -0,0 +1,179 @@
...
+		out, err := kappCtrl.RunWithOpts([]string{"package", "installed", "get", "--package-install", pkgiName, "--json"}, RunOpts{})
+		require.NoError(t, err)
+
+		output := uitest.JSONUIFromBytes(t, []byte(out))
+
+		expectedOutputRows := []map[string]string{{
+			"conditions":      "- type: ReconcileSucceeded\n  status: \"True\"\n  reason: \"\"\n  message: \"\"",
+			"namespace":       env.Namespace,
+			"name":            pkgiName,
+			"overlay_secrets": fmt.Sprintf("- %s-%s-overlays", pkgiName, env.Namespace),
```
cppforlife: if the goal is to assert on overlay_secrets column, i would recommend only asserting on that field. dont want this test to keep changing to unrelated reasons.

cli/test/e2e/ytt_overlays_test.go
```diff
@@ -0,0 +1,179 @@
...
+		require.NotContains(t, out, "test: bar")
+	})
+
+	logger.Section("get package install", func() {
+		out, err := kappCtrl.RunWithOpts([]string{"package", "installed", "get", "--package-install", pkgiName, "--json"}, RunOpts{})
+		require.NoError(t, err)
+
+		output := uitest.JSONUIFromBytes(t, []byte(out))
+
+		expectedOutputRows := []map[string]string{{
```
cppforlife: where is the overlay_secrets column gone?

cli/test/e2e/ytt_overlays_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+	packageName := "test-pkg.carvel.dev"
+	packageVersion := "1.0.0"
+	appName := "test-package"
+	pkgiName := "overlay-test"
+	deploymentName := "simple-app"
+
+	yaml := `---
+apiVersion: data.packaging.carvel.dev/v1alpha1
+kind: PackageMetadata
```
cppforlife: no need to include PackageMetadata (not a required API resource)

cli/test/e2e/ytt_overlays_test.go
```diff
@@ -0,0 +1,179 @@
...
+  version: 1.0.0
+  valuesSchema:
+    openAPIv3:
+      properties:
+        app_port:
+          default: 80
+          description: App port
+          type: integer
+        app_name:
+          description: App Name
```
cppforlife: no need to include valuesSchema

cli/test/e2e/ytt_overlays_test.go
```diff
@@ -0,0 +1,179 @@
...
+  template:
+    spec:
+      fetch:
+      - imgpkgBundle:
+          image: k8slt/kctrl-example-pkg:v1.0.0
```
cppforlife: to eliminate as much external dependencies as possible, lets use inline definition and add a simple ConfigMap.

cli/test/e2e/ytt_overlays_test.go
```diff
@@ -0,0 +1,179 @@
...
+`
+
+	cleanUp := func() {
+		kapp.Run([]string{"delete", "-a", appName})
+	}
+	cleanUp()
+	defer cleanUp()
+
+	logger.Section("add test package", func() {
+		_, err := kapp.RunWithOpts([]string{"deploy", "-a", appName, "-f", "-"}, RunOpts{
```
cppforlife: kapp by default checks on errors, no need to assert

100mik:
To illustrate further, lets say a folder `kctrloverlays` has the following files,
```
├── overlays
│   ├── first
│   │   └── overlay.yml
│   └── second
│       ├── a-overlay.yml
│       └── b-overlay.yml
└── rep-overlay.yaml
```
On running the command: 
```bash
$ kctrl package install -i hello -p hello-app.corp.com --version 1.0.0 --ytt-overlay-file kctrloverlays/rep-overlay.yaml --ytt-overlay-file kctrloverlays/overlays
```
Would result in a secret like,
```bash
apiVersion: v1
data:
  0000-rep-overlay.yaml: I0AgbG9hZCgiQHl0dD...
  0001-first_overlay.yml: I0AgbG9hZCgiQHl0dDpv....
  0001-second_a-overlay.yml: I0AgbG9hZCgiQ....
  0001-second_b-overlay.yml: I0AgbG9hZCgiQ....
kind: Secret
metadata:
  annotations:
    packaging.carvel.dev/package: hello-default
  creationTimestamp: "2022-06-28T07:54:22Z"
  name: hello-default-overlays
  namespace: default
  resourceVersion: "4193758"
  uid: e9288fe4-fe51-4ac5-bfeb-1447be45cbac
type: Opaque
```
The first flag is given priority by the suffix `0000`
All files in the `overlays` folder get the suffix `0001` and are ordered by ytt alphabetically by their file paths

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
100mik: I think the switch case method Dmitriy suggested would help out?

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
100mik: Switched it to `kctrl-ytt-overlays`

cli/test/e2e/ytt_overlays_test.go
100mik: The column is hidden in case there is not overlay supplied

cli/test/e2e/ytt_overlays_test.go
100mik: Adopted!

cli/test/e2e/ytt_overlays_test.go
100mik: Adopted!

cli/test/e2e/ytt_overlays_test.go
100mik: Removed 🗑️

cli/test/e2e/ytt_overlays_test.go
100mik: Removed 🗑️

cli/test/e2e/ytt_overlays_test.go
100mik: Made that change!

cli/test/e2e/ytt_overlays_test.go
100mik: Changed it up to only check if it exits successfully

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
100mik: The condition above checks whether any of the flags that are needed to change the installation are provided.
The latter checks if the installation needs to be paused, asks "Do I need to update existing secrets"

The one above asks "Are there any flags that are needed to update the package present"

Additionally, both would be run only if `kctrl package installed update` is used. Not if an install falls back to an update.

However I believe we could format it in a similar fashion 🤔

praveenrewar:
Looks good to me other than the switch conditions which are almost similar, but I couldn't think of a better way to have those conditions.

100mik:
Spent some time on it as well, but most ways around it seem to just be moving the same conditionals elsewhere :(

praveenrewar (approved):

PR #738, opened 2022-06-10
E2e controller config scoped

### LINKED ISSUE #732: Get rid of config-test
Right now our e2e tests depend on a cluster that's been deployed with a special configuration (the files in https://github.com/vmware-tanzu/carvel-kapp-controller/tree/d49800d7d5024058911e096cfd7487f49ccf0c5f/config-test). That means they won't pass against a kapp-controller deployed by the release manifest, or `hack/deploy.sh`. 

It would be better if we had the tests that require special configuration reconfigure the server in the way they need to test, and then revert it back when they're done. This might also mean these tests can't run in parallel with any other tests, since they're changing global state and potentially restarting the pod.

cppforlife:

fixes https://github.com/vmware-tanzu/carvel-kapp-controller/issues/732

as well as making all fields in the Config reloadable.

not really

recommend to review by each commit.

##### Review Checklist:

  a link to that PR
  change

```docs

```


test/e2e/secretgencontroller/private_registry_auth_test.go
```diff
@@ -182,42 +141,98 @@ spec:
...
+  name: kapp-controller-config
+  namespace: kapp-controller
+stringData:
+  dangerousSkipTLSVerify: registry-svc.registry.svc.cluster.local
+`
+		kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", configName},
+			e2e.RunOpts{StdinReader: strings.NewReader(config)})
+
+		// Since config propagation is async, just wait a little bit
+		time.Sleep(2 * time.Second)
```
benmoss, line 165: flaky test detected. can we figure out a way to not use sleeps?

cppforlife: im open to ideas, but i dont have any (that would be worth investment over time.sleep)

benmoss: i can't think of anything easy, we'd need some way of turning the config update into a synchronous operation, like an endpoint we could poll or an event emitted, even a line logged. i think i'm fine with ignoring this.

PR #743, opened 2022-06-15
Enhance tty experience

### LINKED ISSUE #717: The "source <(kctrl completion bash)" command outputs an error message
**What steps did you take:**
[A clear and concise description steps and any files that can be used to reproduce the problem.]

Added `source <(kctrl completion bash)` to my `/etc/profile` so it would automatically load.

To reproduce the issue, you can simply run `source <(kctrl completion bash)`.

**What happened:**
[A small description of the result]

Got the following error message:

```text
Succeeded: command not found
```

This is because the `kctrl completion bash` commands outputs a `Succeeded` string at the end, which is not a valid command, and when used with `source`, bash tries to execute it.

Bash completion does work, though.

**What did you expect:**

As a workaround, I'm using the following:

```bash
source <(kctrl completion bash | grep -v Succeeded)
```

This removes the `Succeeded` string from the output, so the error doesn't occur.

Same issue has also occurred with kapp before, as I described here:
https://github.com/vmware-tanzu/carvel-kapp/issues/447
In kapp, this issue was fixed using the `--tty=false` option.

Same issue also occurs in imgpkg, as I described here:
https://github.com/vmware-tanzu/carvel-imgpkg/issues/342

This issue doesn't occur with ytt, because `ytt completion bash` doesn't output the `Succeeded` string.

**Environment:**

- kctrl version (use `kctrl --version`): `0.36.1`
- OS (e.g. from `/etc/os-release`): Using `Linux Ubuntu 20.04.4 LTS (Focal Fossa)` and `macOS Monterey 12.2.1`, but this issue occurs on any bash.

---
Vote on this request

praveenrewar:
- Set tty flag as an alternate/other flag instead of global flag.
- Set default value of tty to true for add/update, delete, pause, kick and status commands. For rest of the commands, default value should be false.

Fixes #717 

```release-note
NONE
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/cli/pkg/kctrl/cmd/kctrl.go
+++ b/cli/pkg/kctrl/cmd/kctrl.go
@@ -98,6 +98,7 @@ func NewKctrlCmd(o *KctrlOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Comm
 
 	cmd.AddCommand(NewCmdCompletion())
 
+	ConfigureTTY(o, cmd)
 	return cmd
 }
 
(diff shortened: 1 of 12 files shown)


cli/pkg/kctrl/cmd/kctrl.go
```diff
@@ -143,6 +144,20 @@ func ConfigureGlobalFlags(o *KctrlOptions, cmd *cobra.Command, flagsFactory cmdc
 	}
 }
 
+func ConfigureTTY(o *KctrlOptions, cmd *cobra.Command) {
```
cppforlife: do we need to have this public?

praveenrewar: Hmm, right, we can actually make the other function `ConfigureGlobal` private as well.

praveenrewar: Done

PR #744, opened 2022-06-15
add downward api

### LINKED ISSUE #659: app cr downward api to pass info as values into templating

there have been a few cases where we want to pass info about App CR as values into templates:
- App CR name and namespace so that resources can be customized
  - if i am creating multiple App CRs in the same namespace, i would like to avoid users having to manually specify unique name (App CR name would be unique within a namespace, or in cluster if it's namespace+name)
- App CR annotations/labels
  - may want to propagate certain annotations to created resources
- App CR uid
- App CR fetch information (vendir.lock.yml)
  - may want to include git info about where content came from as an annotation on a resource

something along these lines... (existing valuesFrom expanded)
```yaml
valuesFrom:
- secretRef:
    name: secret-name
- configMapRef:
    name: cfgmap-name
- path: values/shared.yml
# new type downwardAPI
- downwardAPI:
    items:
    - name: namespace
      fieldPath: metadata.namespace
    - name: name
      fieldPath: metadata.name
    - name: somewhere.source
      vendirLock: {} # since it's not a field on App CR
```

prior art for downward API in Pods -- https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/

---
Vote on this request

DennisDenuto:
Draft PR WIP. Still need to:

- [x] clean up commits
- [x]  fix pb gen code
- [x]  add tests around annotations / labels
- [x]  add test coverage for other non-ytt templating
- [x]  add comments
- [x]  do the TODO's
- [x]  think about allowing configuring a path as a valid option
- [x]  fix up commits / messages
- [ ]  split out doc change (dev.md) into a separate PR
- [x]  Add documentation to carvel.dev as a separate PR

Fixes #659 

```release-note
Allow specifying the App name, namespace, uid, annotations and labels as values during templating
```

##### Review Checklist:

  a link to that PR
  change


pkg/template/values.go
```diff
@@ -102,6 +110,105 @@ func (t Values) writeFromSecret(dstPath string,
...
+		}
+		result = append(result, path)
+	}
+
+	sort.Strings(result)
+
+	return result, nil
+}
+
+func (t Values) extractFieldPathAsKeyValue(name string, fieldPath string) ([]byte, error) {
```
cppforlife: wdyt about using json path library that kubectl uses for extracting values?

DennisDenuto: yeah thx for the pointer! quick search shows up https://github.com/kubernetes/kubectl/blob/ac26f503e81287d9903761a1a8ded25fdebec6a7/pkg/cmd/get/customcolumn.go#L47

I'll def have a closer look as I can see this being a more generalized solution.

Should I just go ahead and load up the entire metadata and allow folk to inject any jsonpath within '.metadata'? or do we still want to be restrictive to only allow name, namespace, labels, annotations, uid as valid json paths?

cppforlife: looks like it's using https://github.com/kubernetes/client-go/tree/master/util/jsonpath.

i think it's still probably a good idea to limit, but instead of limiting the json path, we can limit input structure. that way you dont have make assertions against a query.

pkg/apis/kappctrl/v1alpha1/types_template.go
```diff
@@ -57,16 +57,29 @@ type AppTemplateHelmTemplate struct {
...
 // +k8s:openapi-gen=true
 type AppTemplateValuesSourceRef struct {
 	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
 }
 
+// +k8s:openapi-gen=true
+type DownwardAPIAppTemplateValues struct {
+	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
+	// Required: Selects a field of the app: only annotations, labels, uid, name and namespace are supported.
+	FieldPath string `json:"fieldPath" protobuf:"bytes,2,opt,name=fieldPath"`
```
cppforlife: probably should add omitempty here as well?

pkg/template/helm_template.go
```diff
@@ -109,8 +109,3 @@ func (t *HelmTemplate) template(dirPath string, input io.Reader) exec.CmdRunResu
 
 	return result
 }
-
-// auxiliary struct used for Chart.yaml unmarshalling
```
cppforlife, line 113: nice!

pkg/template/interfaces.go
```diff
@@ -23,7 +25,26 @@ type Template interface {
 	TemplateStream(stream io.Reader, dirPath string) exec.CmdRunResult
 }
 
-type GenericOpts struct {
+// AppContext carries App information used across API boundaries.
+// Primarily used in a context when templating with values
+type AppContext struct {
 	Name      string
 	Namespace string
+	Metadata  *PartialObjectMetadata
```
cppforlife: do we need this to be a pointer?

pkg/template/values.go
```diff
@@ -102,10 +112,104 @@ func (t Values) writeFromSecret(dstPath string,
 	return result, nil
 }
 
+func (t Values) writeFromDownwardAPI(dstPath string, downwardAPIRef v1alpha1.AppTemplateValuesDownwardAPI) ([]string, error) {
```
cppforlife: feels like we should extract this method and associated methods below into ValuesDownwardAPI class. seems like Values is growing a bit too much. (unit tests can stay the same though)

pkg/template/values.go
```diff
@@ -102,10 +112,104 @@ func (t Values) writeFromSecret(dstPath string,
...
+		}
+		result = append(result, path)
+	}
+
+	return result, nil
+}
+
+func (Values) validateName(name string) error {
+	if strings.HasSuffix(name, ".") || strings.HasPrefix(name, ".") || strings.Contains(name, "..") {
+		return errors.New("Invalid name was provided (hint: separate paths should only use a single '.' character)")
```
cppforlife: probably best to include name in the error.

pkg/template/values.go
```diff
@@ -102,10 +112,104 @@ func (t Values) writeFromSecret(dstPath string,
...
+		return nil, err
+	}
+
+	results, err := path.FindResults(t.appContext.Metadata)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(results) != 1 || len(results[0]) != 1 {
+		return nil, errors.New("invalid field spec provided to DownwardAPI. Only single supported fields are allowed")
```
cppforlife: nit: capital `I`

pkg/template/values.go
```diff
@@ -102,10 +112,104 @@ func (t Values) writeFromSecret(dstPath string,
...
+	}
+
+	return t.keyValue(name, fmt.Sprintf("%v", result.Interface())), nil
+}
+
+// operator may wish to assign a downward API value into a nested key structure to use within their template
+func (Values) nestedKey(name string) string {
+	var nested string
+	for idx, n := range splitWithEscaping(name, '.', '\\') {
+		nested += fmt.Sprintf("%s%s:\n", strings.Repeat("  ", idx), n)
```
cppforlife: instead of building up yaml content via strings, lets use maps and yaml.Marshal the result.

pkg/template/values.go
```diff
@@ -102,10 +112,104 @@ func (t Values) writeFromSecret(dstPath string,
...
+		return t.keyValues(name, result.Interface().(map[string]string))
+	}
+
+	return t.keyValue(name, fmt.Sprintf("%v", result.Interface())), nil
+}
+
+// operator may wish to assign a downward API value into a nested key structure to use within their template
+func (Values) nestedKey(name string) string {
+	var nested string
+	for idx, n := range splitWithEscaping(name, '.', '\\') {
```
cppforlife: im trying to think, we have any prior art in escaping value keys... in carvel/kapp-controller cc @pivotaljohn

pkg/template/values.go
DennisDenuto: Yes! +1 

I went back and forth on adding a new `ValuesDownwardAPI` to encapsulate this logic. I landed on "let's see if someone else thinks this during the review"

:-)

pkg/template/values.go
DennisDenuto: attention 👏 to 👏 detail 👏

pkg/template/values.go
DennisDenuto: ok done

pkg/template/interfaces.go
DennisDenuto: don't think so. Removed pointer

pkg/apis/kappctrl/v1alpha1/types_template.go
DennisDenuto: done

pkg/template/values.go
DennisDenuto: ok done

pkg/apis/kappctrl/v1alpha1/types_template.go
```diff
@@ -57,16 +57,29 @@ type AppTemplateHelmTemplate struct {
...
+	DownwardAPI  *AppTemplateValuesDownwardAPI `json:"downwardAPI,omitempty" protobuf:"bytes,4,opt,name=downwardAPI"`
 }
 
 // +k8s:openapi-gen=true
 type AppTemplateValuesSourceRef struct {
 	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
 }
 
+// +k8s:openapi-gen=true
+type DownwardAPIAppTemplateValues struct {
```
cppforlife: nit: i would recommend calling this `AppTemplateValuesDownwardAPIItem` so that it matches parent `AppTemplateValuesDownwardAPI`

pkg/template/metadata_values_extractor.go
```diff
@@ -0,0 +1,176 @@
...
+	result := results[0][0]
+	kind := result.Kind()
+	if kind == reflect.Interface {
+		kind = result.Elem().Kind()
+	}
+	if kind == reflect.Map {
+		return a.keyValues(name, result.Interface().(map[string]string))
+	}
+
+	yamlContent, _, err := a.nestedKeyValue(name, fmt.Sprintf("%v", result.Interface()))
```
cppforlife: to get rid of all string building use `a.nestedKeyValue(name, result.Interface())`.

also no need to check on `if kind == reflect.Map {` above.

pkg/template/values.go
```diff
@@ -102,27 +109,28 @@ func (t Values) writeFromSecret(dstPath string,
...
-	configMapRef v1alpha1.AppTemplateValuesSourceRef) ([]string, error) {
+// DownwardAPIValuesExtractor produces multiple key-values based on the DownwardAPI config
+// provided by the operator
+type DownwardAPIValuesExtractor interface {
+	Values() ([][]byte, error)
+}
 
-	configMap, err := t.coreClient.CoreV1().ConfigMaps(t.genericOpts.Namespace).Get(
-		context.Background(), configMapRef.Name, metav1.GetOptions{})
+func (t Values) writeFromDownwardAPI(dstPath string, valuesExtractor DownwardAPIValuesExtractor) ([]string, error) {
```
cppforlife: can just take a concrete type especially since method is called writeFromDownwardAPI.

pkg/template/metadata_values_extractor.go
```diff
@@ -0,0 +1,176 @@
...
+
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
+	"gopkg.in/yaml.v2"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/util/jsonpath"
+)
+
+// ObjectMetadataValuesExtractor produces multiple key-values based on the DownwardAPI config
+// queried against the object metadata
+type ObjectMetadataValuesExtractor struct {
```
cppforlife: wdyt about just calling this: DownwardAPIValues

pkg/template/metadata_values_extractor.go
```diff
@@ -0,0 +1,176 @@
...
+	items    []v1alpha1.DownwardAPIAppTemplateValues
+	metadata PartialObjectMetadata
+}
+
+// Values returns many key-values queried (using jsonpath) against an object metadata provided.
+func (a ObjectMetadataValuesExtractor) Values() ([][]byte, error) {
```
cppforlife: i would recommend calling this `AsYAMLs()`

test/e2e/kappcontroller/template_test.go
```diff
@@ -113,6 +117,18 @@ spec:
...
+        - downwardAPI:
+            items:
+            - name: namespace
+              fieldPath: metadata.namespace
+            - name: name
+              fieldPath: metadata.name
+            - name: uid
+              fieldPath: metadata.uid
+            - name: annotation
+              fieldPath: metadata.annotations['expectedAnnotation']
```
cppforlife, line 130: can we also throw item that returns a map

```
- name: annotations
  fieldPath: metadata.annotations
```

pkg/template/metadata_values_extractor.go
DennisDenuto: if map is provided to `a.nestedKeyValue(name, map)` then the yaml produced will not be ordered / consistent across multiple templating.

Is that ok?

pkg/apis/kappctrl/v1alpha1/types_template.go
DennisDenuto: ok done

pkg/template/values.go
DennisDenuto: ok done

pkg/template/metadata_values_extractor.go
DennisDenuto: ok changed

pkg/template/metadata_values_extractor.go
DennisDenuto: ok changed

test/e2e/kappcontroller/template_test.go
DennisDenuto: ok will do.

pkg/template/metadata_values_extractor.go
DennisDenuto: Actually, yaml.Marshall seems to also provide stable out. so I just went ahead and did that! :-)

pkg/template/values.go
cppforlife: related conversations in ytt: https://github.com/vmware-tanzu/carvel-ytt/issues/671

pivotaljohn: LGTM... IMO, this is actually the ideal UX for ytt users.  So, when we implement 671, we'll also use the slash for escaping dots in paths. 👍🏻

PR #746, opened 2022-06-16
Fix dev-deploy

benmoss:
Fixes the hack/dev-deploy script to work with the latest changes to the build process.


joe-kimmel-vmw:
~hmm. this failed~  - 

~it installs successfully but the first e2e test that ran failed~

just want to update that we figured out that this was working, and my cluster was in a failing state for other reasons.

joe-kimmel-vmw (approved):
LGTM

PR #757, opened 2022-06-21
Disallow use of shared namespaces for package installs

### LINKED ISSUE #672: `kctrl`: disallow use of shared namespaces while installing packages
`kctrl` should not allow installation of packages into shared namespaces

This could be disabled by default and we could add a dangerous flag overriding this behaviour.

- [ ] disallow use of shared namespaces for installation
- [ ] Add a flag which accepts comma seperated values where packages can be installed to override this behaviour, similar to `--dangerous-allow-use-of-shared-namespace=default,kube-public`

---
Vote on this request

100mik:
Disallows use of shared namespaces for package installations by default, to avoid storing sensitive configuration and RBAC resources in them. Adds a dangerous flag to override this behaviour.

Fixes #672 

```release-note
- Disallow use of shared namespaces (`default`, `kube-public`) by default
    - `--dangerous-allow-use-of-shared-namespace` can be used to override this default behaviour
```

##### Review Checklist:

  a link to that PR
  change

To be added
```docs

```

### DIFF
--- a/cli/pkg/kctrl/cmd/core/namespace_flags.go
+++ b/cli/pkg/kctrl/cmd/core/namespace_flags.go
@@ -13,8 +13,14 @@ import (
 	"github.com/spf13/pflag"
 )
 
+var sharedNamespaces = []string{
+	"default",
+	"kube-public",
+}
+
 type NamespaceFlags struct {
-	Name string
+	Name                    string
+	AllowedSharedNamespaces []string
 }
 
 func (s *NamespaceFlags) Set(cmd *cobra.Command, flagsFactory FlagsFactory) {
@@ -26,6 +32,7 @@ func (s *NamespaceFlags) SetWithPackageCommandTreeOpts(cmd *cobra.Command, flags
 	namespaceEnvVariableKey := fmt.Sprintf("%s_NAMESPACE", strings.ToUpper(opts.BinaryName))
 	name := flagsFactory.NewNamespaceNameFlag(&s.Name, namespaceEnvVariableKey)
 	cmd.Flags().VarP(name, "namespace", "n", fmt.Sprintf("Specified namespace ($%s or default from kubeconfig)", namespaceEnvVariableKey))
+	cmd.Flags().StringArrayVar(&s.AllowedSharedNamespaces, "dangerous-allow-use-of-shared-namespace", []string{}, "Allow use of shared namespaces (optional, comma separated strings)")
 }
 
 type NamespaceNameFlag struct {
@@ -57,6 +64,22 @@ func (s *NamespaceNameFlag) Resolve() error {
 
 	*s.value = value
 
+	return nil
+}
+
+func (s *NamespaceFlags) CheckForDisallowedSharedNamespaces() error {
+	for _, ns := range sharedNamespaces {
+		if s.Name == ns {
+			for _, allowedNs := range s.AllowedSharedNamespaces {
+				if ns == allowedNs {
+					return nil
+				}
+			}
+			return fmt.Errorf(`Creating sensitive resources in a shared namespace (%s)
+			(hint: Specify a namespace using the '-n' flag or use kubeconfig to change default namespace 'kubectl config set-context --current --namespace=private-namespace'.
+			Or use '--dangerous-allow-allow-use-of-shared-namespace=%s' to allow use of shared namespace)`, s.Name, s.Name)
+		}
+	}
 	return nil
 }
 
--- a/cli/pkg/kctrl/cmd/core/namespace_flags.go
+++ b/cli/pkg/kctrl/cmd/core/namespace_flags.go
@@ -13,14 +13,8 @@ import (
 	"github.com/spf13/pflag"
 )
 
-var sharedNamespaces = []string{
-	"default",
-	"kube-public",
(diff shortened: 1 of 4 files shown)


100mik:
I think we should disallow addition of repositories as well, if we do not allow installation of packages.

Have not added tests yet as the tests default to `kctrl-test` as of now. Not super feasible to test while changing up the ns flag.

cli/pkg/kctrl/cmd/core/namespace_flags.go
```diff
@@ -26,6 +32,7 @@ func (s *NamespaceFlags) SetWithPackageCommandTreeOpts(cmd *cobra.Command, flags
 	namespaceEnvVariableKey := fmt.Sprintf("%s_NAMESPACE", strings.ToUpper(opts.BinaryName))
 	name := flagsFactory.NewNamespaceNameFlag(&s.Name, namespaceEnvVariableKey)
 	cmd.Flags().VarP(name, "namespace", "n", fmt.Sprintf("Specified namespace ($%s or default from kubeconfig)", namespaceEnvVariableKey))
+	cmd.Flags().StringArrayVar(&s.AllowedSharedNamespaces, "dangerous-allow-use-of-shared-namespace", []string{}, "Allow use of shared namespaces (optional, comma separated strings)")
```
praveenrewar: I am thinking if this should be a boolean 🤔

100mik: The idea is to allow the user to be specific about which ns they want to allow installation in 🤔 
I can see it being a boolean because we are just protecting two namespaces anyway

cppforlife: @100mik im +1 on making this a boolean.

100mik: Adopted!

100mik:
Added docs for the same here VMware-tanzu/carvel#475

cli/pkg/kctrl/cmd/core/namespace_flags.go
```diff
@@ -26,6 +32,7 @@ func (s *NamespaceFlags) SetWithPackageCommandTreeOpts(cmd *cobra.Command, flags
 	namespaceEnvVariableKey := fmt.Sprintf("%s_NAMESPACE", strings.ToUpper(opts.BinaryName))
 	name := flagsFactory.NewNamespaceNameFlag(&s.Name, namespaceEnvVariableKey)
 	cmd.Flags().VarP(name, "namespace", "n", fmt.Sprintf("Specified namespace ($%s or default from kubeconfig)", namespaceEnvVariableKey))
+	cmd.Flags().BoolVar(&s.AllowedSharedNamespaces, "dangerous-allow-use-of-shared-namespace", false, "Allow use of shared namespaces (default: false)")
```
cppforlife: nit: remove `(default: false)` this is something that cobra should do for you

cli/pkg/kctrl/cmd/core/namespace_flags.go
```diff
@@ -60,6 +67,20 @@ func (s *NamespaceNameFlag) Resolve() error {
...
 
+func (s *NamespaceFlags) CheckForDisallowedSharedNamespaces() error {
+	if s.AllowedSharedNamespaces {
+		return nil
+	}
+	for _, ns := range sharedNamespaces {
+		if s.Name == ns {
+			return fmt.Errorf(`Creating sensitive resources in a shared namespace (%s)
+			(hint: Specify a namespace using the '-n' flag or use kubeconfig to change default namespace 'kubectl config set-context --current --namespace=private-namespace'.
+			Or use '--dangerous-allow-allow-use-of-shared-namespace=%s' to allow use of shared namespace)`, s.Name, s.Name)
```
cppforlife: is this error going to appear as multi-line indented to users? if, so that would be a little weird. if you want to split it for readability of code i would recommend using `"..." + "..." + "..."`

cli/pkg/kctrl/cmd/core/namespace_flags.go
```diff
@@ -13,8 +13,14 @@ import (
...
 
+var sharedNamespaces = []string{
+	"default",
+	"kube-public",
+}
+
 type NamespaceFlags struct {
-	Name string
+	Name                    string
+	AllowedSharedNamespaces bool
```
cppforlife: just realized that this is going to affect all commands. we should only have this logic for package/repo add/update i think. you can either create "SecureNamespaceFlags" struct that embeds NamespaceFlags, or have some conditionals.

cli/pkg/kctrl/cmd/core/namespace_flags.go
100mik: Done!

cli/pkg/kctrl/cmd/core/namespace_flags.go
100mik: Made the change

cli/pkg/kctrl/cmd/core/namespace_flags.go
100mik: Moved it to a separate struct!

PR #766, opened 2022-06-27
wip kctrl pkg init release

rohitagg2020:
wip do not merge

to build:

cd cli
./hack/build.sh
./kctrl pkg init
./kctrl app init
./kctrl pkg release
./kctrl pkg repo release


cli/pkg/kctrl/cmd/app/init/build/default.go
```diff
@@ -0,0 +1,52 @@
...
+apiVersion: kctrl.carvel.dev/v1alpha1
+kind: AppBuild
+spec: 
+  export:
+  - imgpkgBundle:
+      image:
+      useKbldImagesLock: true
+      includePaths:
+      - kubernetes-manifests/**/*
+      - config-dev/build.yml
```
cppforlife: dont think we should have these in default.

cli/pkg/kctrl/cmd/app/init/configure/fetch/vendir.go
```diff
@@ -0,0 +1,272 @@
...
+		vendirDirectories = v.initializeVendirDirectorySection()
+	} else {
+		directory := vendirDirectories[0]
+		if len(directory.Contents) > 1 {
+			//TODO what needs to be done in this case
+			//As multiple content sections are configured, we dont want to touch them.
+			return nil
+		}
+	}
+	currentFetchOptionSelected := v.appBuild.ObjectMeta.Annotations[common.FetchContentAnnotationKey]
```
cppforlife: we should detect what's current based on whats inside the vendir configuration, not based on annotation, since annotation is not the source of truth.

cli/pkg/kctrl/cmd/core/authoring_ui.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/cppforlife/color"
+	"github.com/cppforlife/go-cli-ui/ui"
+	"github.com/mitchellh/go-wordwrap"
+)
+
+type IAuthoringUI interface {
```
cppforlife: in go, `I` is not commonly used for interfaces.

cli/pkg/kctrl/cmd/core/authoring_ui.go
```diff
@@ -0,0 +1,73 @@
...
+	AskForChoice(opts ui.ChoiceOpts) (int, error)
+	PrintCmdExecutionOutput(text string)
+	PrintHeaderText(text string)
+}
+
+type AuthoringUIImpl struct {
+	ui ui.UI
+}
+
+func NewAuthoringUI(ui ui.UI) IAuthoringUI {
```
cppforlife: general rule of thumb: accept interfaces, return concrete types.

NewAuthoringUI => NewAuthoringUIImpl since we are not "creating" interfaces.

cli/pkg/kctrl/cmd/package/init/build/pkg-build.go
```diff
@@ -0,0 +1,203 @@
...
+	Template appBuild.AppBuild  `json:"template, omitempty"`
+	Release  []appBuild.Release `json:"release, omitempty"`
+}
+
+type Template struct {
+	Spec TemplateSpec `json:"spec, omitempty"`
+}
+
+type TemplateSpec struct {
+	Export []appBuild.Export         `json:"export, omitempty"`
```
cppforlife: global nit: no space in front of omitempty.

cli/pkg/kctrl/cmd/package/init/build/pkg-build.go
```diff
@@ -0,0 +1,203 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package build
+
+import (
+	v1alpha12 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
+	"os"
+
+	kapp "github.com/k14s/kapp/pkg/kapp/resources"
```
cppforlife: use cmdlocal.NewConfigFromFiles instead of kapp.

cli/pkg/kctrl/cmd/package/init/init.go
```diff
@@ -0,0 +1,343 @@
...
+
+	cmd.PersistentFlags().StringVarP(&o.pkgVersion, "version", "v", "", "Version of a package (in semver format)")
+	return cmd
+}
+
+func (o *InitOptions) Run() error {
+	o.ui.PrintHeaderText("\nPre-requisite")
+	o.ui.PrintInformationalText("Welcome! Before we start on the package creation journey, please ensure the following pre-requites are met:\n* The Carvel suite of tools are installed. Do get familiar with the following Carvel tools: ytt, imgpkg, vendir, and kbld.\n* You have access to an OCI registry, and authenticated locally so that images can be pushed. e.g. docker login <REGISTRY URL>\n")
+
+	pkgBuildFilePath := filepath.Join(PkgBuildFileName)
```
cppforlife: what are we joining here?

cli/pkg/kctrl/cmd/package/init/init.go
```diff
@@ -0,0 +1,343 @@
...
+	if err != nil {
+		return err
+	}
+	pkgInstall, err := build.GetPackageInstall(PkgResourcesFileName)
+	if err != nil {
+		return err
+	}
+	createStep := NewCreateStep(o.ui, pkgBuild, pkg, pkgMetadata, pkgInstall, o.logger, o.pkgVersion, o.depsFactory)
+	createStep.pkg = pkg
+	createStep.pkgMetadata = pkgMetadata
```
cppforlife: shouldn't be setting private accessors (especially since we pass info down to constructor).

cli/pkg/kctrl/cmd/package/init/init.go
```diff
@@ -0,0 +1,343 @@
...
+		existingPkgInstall.ObjectMeta.Annotations = make(map[string]string)
+		existingPkgInstall.ObjectMeta.Annotations[common.LocalFetchAnnotationKey] = "."
+	}
+
+	if len(existingPkgInstall.Name) == 0 {
+		existingPkgInstall.Name = createStep.pkg.Name
+	}
+
+	if len(existingPkgInstall.Spec.ServiceAccountName) == 0 {
+		existingPkgInstall.Spec.ServiceAccountName = "sa-" + createStep.pkg.Name
```
cppforlife: most commonly "sa" is a suffix and not a prefix.

cli/pkg/kctrl/cmd/package/init/init.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+func NewInitCmd(o *InitOptions) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "init",
+		Short: "Initialize Package",
+		RunE:  func(_ *cobra.Command, _ []string) error { return o.Run() },
+	}
+
+	cmd.PersistentFlags().StringVarP(&o.pkgVersion, "version", "v", "", "Version of a package (in semver format)")
```
cppforlife: for init we should always use version `0.0.0` since we shouldnt be overriding configuration files every time version changes. at the release/build time however, correct version will be inserted based on what user has specified via --version flag.

cli/pkg/kctrl/cmd/package/init/init.go
```diff
@@ -0,0 +1,343 @@
...
+func (createStep CreateStep) printInformation() {
+	createStep.ui.PrintInformationalText("\n**Information**\npackage-build.yml is generated as part of this flow. This file can be used for further updating and adding complex scenarios while using the `kctrl pkg build create` command. Please read the link'ed documentation for more explanation.")
+}
+
+func (createStep CreateStep) generateAppBuildFromPackageBuild() *appBuild.AppBuild {
+	appBuild := createStep.pkgBuild.Spec.Template
+	return &appBuild
+}
+
+func (createStep CreateStep) populatePkgBuildFromAppBuild(appBuild *appBuild.AppBuild) error {
```
cppforlife: nit: i would recommend inlining this operation so it's immediately clear we are modifying createStep.

cli/pkg/kctrl/cmd/app/init/common/constants.go
```diff
@@ -0,0 +1,18 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package common
+
+const (
+	FetchContentAnnotationKey = "fetch-content-from"
+	LocalFetchAnnotationKey   = "kctrl.carvel.dev/local-fetch-0"
+	PackageBuildFileName      = "package-build.yml"
+)
```
cppforlife: lets move these constants to places that "own" them. e.g. PackageBuildFileName should not be within an app, but rather should be within package related code.

cli/pkg/kctrl/cmd/app/init/configure/fetch/vendir.go
```diff
@@ -0,0 +1,272 @@
...
+	IncludeAllFiles     = "*"
+)
+
+type VendirStep struct {
+	ui       cmdcore.IAuthoringUI
+	appBuild *build.AppBuild
+	config   vendirconf.Config
+}
+
+func NewVendirStep(ui cmdcore.IAuthoringUI, appBuild *build.AppBuild, config vendirconf.Config) *VendirStep {
```
cppforlife: i dont think vendir step should know anything appBuild. (relates also to source of truth comment below)

cli/pkg/kctrl/cmd/app/init/configure/fetch/fetch.go
```diff
@@ -0,0 +1,103 @@
...
+	vendirconf "github.com/vmware-tanzu/carvel-vendir/pkg/vendir/config"
+)
+
+type FetchStep struct {
+	ui                        cmdcore.IAuthoringUI
+	appBuild                  *build.AppBuild
+	isAppCommandRunExplicitly bool
+}
+
+func NewFetchStep(ui cmdcore.IAuthoringUI, appBuild *build.AppBuild, isAppCommandRunExplicitly bool) *FetchStep {
```
cppforlife: dont think fetch step should know anything about AppBuild. fetching is done separately from building.

cli/pkg/kctrl/cmd/app/init/common/interface.go
```diff
@@ -0,0 +1,26 @@
...
+func Run(step Step) error {
+	err := step.PreInteract()
+	if err != nil {
+		return err
+	}
+	err = step.Interact()
+	if err != nil {
+		return err
+	}
+	err = step.PostInteract()
```
neil-hickey: suggestion: 

```
return step.PostInteract()
```

cli/pkg/kctrl/cmd/app/init/configure/fetch/fetch.go
```diff
@@ -0,0 +1,103 @@
...
+		vendirConfig = NewDefaultVendirConfig()
+	} else {
+		vendirConfig, err = ReadVendirConfig()
+		if err != nil {
+			return err
+		}
+	}
+
+	vendirStep := NewVendirStep(fetchStep.ui, fetchStep.appBuild, vendirConfig)
+	err = common.Run(vendirStep)
```
neil-hickey: similar to above, I think anywhere that we are just returning an error / nil at the end, we can just inline it to be `return func()` instead of `err = func(), if err != nil { return err } else { return nil }`

cli/pkg/kctrl/cmd/app/init/build/default.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/core/authoring_ui.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/core/authoring_ui.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/init/build/pkg-build.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/init/init.go
rohitagg2020: Removed. Adopted

cli/pkg/kctrl/cmd/package/init/init.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/release/release.go
```diff
@@ -0,0 +1,157 @@
...
+
+	cmd.Flags().StringVarP(&o.pkgVersion, "version", "v", "", "Version to be released")
+	cmd.Flags().BoolVar(&o.debug, "debug", false, "Version to be released")
+
+	return cmd
+}
+
+func (o *ReleaseOptions) Run() error {
+	if o.pkgVersion == "" {
+		o.pkgVersion = fmt.Sprintf("build-%d", time.Now().Unix())
```
cppforlife, line 63: should be valid semver `0.0.0+build.%d`

cli/pkg/kctrl/cmd/package/release/release.go
```diff
@@ -0,0 +1,157 @@
...
+
+func NewReleaseCmd(o *ReleaseOptions) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "release",
+		Short: "Release package",
+		RunE:  func(cmd *cobra.Command, args []string) error { return o.Run() },
+	}
+
+	cmd.Flags().StringVarP(&o.pkgVersion, "version", "v", "", "Version to be released")
+	cmd.Flags().BoolVar(&o.debug, "debug", false, "Version to be released")
```
cppforlife: copy pasta for description

cli/pkg/kctrl/cmd/package/release/release.go
```diff
@@ -0,0 +1,157 @@
...
+	err = os.Mkdir(".imgpkg", fs.FileMode(0777))
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(".imgpkg")
+	wd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+	imgpkgLockPath := filepath.Join(wd, ".imgpkg", "images.yml")
```
cppforlife: since at this point we do not know whether it's an imgpkg export or not, we shouldnt be creating it here (may be in the switch statement?)

also seems a bit awkward to create real .imgpkg in main directory where user is operating. i think we should create a temp directory where we copy all the necessary bits and there create .imgpkg directory. (partly why we have includePaths on imgpkgBundle export)

cli/pkg/kctrl/cmd/package/release/imgpkg_runner.go
```diff
@@ -0,0 +1,64 @@
...
+	wd, err := os.Getwd()
+	for _, path := range r.Paths {
+		var stderrBuf bytes.Buffer
+		cmd := goexec.Command("cp", "-r", filepath.Join(wd, path), dir)
+		cmd.Stderr = &stderrBuf
+		err := cmd.Run()
+		if err != nil {
+			return "", fmt.Errorf("%s", stderrBuf.String())
+		}
+	}
```
cppforlife: copying should not have to use an external command e.g. https://github.com/vmware-tanzu/carvel-vendir/blob/develop/pkg/vendir/directory/directory.go#L168-L171

cli/pkg/kctrl/cmd/package/init/build/pkg-build.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/init/init.go
rohitagg2020: Hypothetical question: What should we do if user has explicitly mentioned the version as something else than 0.0.0 in the package.yml. Still should we override that with 0.0.0 or we honor what user has provided in package.yml.

cli/pkg/kctrl/cmd/app/init/common/interface.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/configure/fetch/fetch.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/configure/fetch/helm.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return err
+	}
+	return helmStep.configureHelmChartRepositoryURL()
+}
+
+func (helmStep *HelmStep) initializeHelmRelease() error {
+	helmChartContent := vendirconf.DirectoryContentsHelmChart{
+		// TODO if the binary name is helm2, then it wont work.
+		HelmVersion: "3",
```
rohitagg2020: Should we add helm also as a pre-requisite.

cli/pkg/kctrl/cmd/app/init/configure/fetch/vendir.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func (v *VendirStep) PreInteract() error {
+	return nil
+}
+
+func (v *VendirStep) Interact() error {
+	vendirDirectories := v.config.Directories
+	if len(vendirDirectories) > 1 {
+		//TODO what if we have >1 Directories section in the vendir conf
```
rohitagg2020: If the vendir.yml has more than 1 directories, then should we return from here as this is a advance use case for the tool? This mean we wont ask for question, but we will still run `vendir sync`

cli/pkg/kctrl/cmd/app/init/configure/fetch/vendir.go
```diff
@@ -0,0 +1,218 @@
...
+
+	}
+	if len(vendirDirectories) == 0 {
+		vendirDirectories = v.initializeVendirDirectorySection()
+	} else {
+		directory := vendirDirectories[0]
+		if len(directory.Contents) > 1 {
```
rohitagg2020: If the vendir.yml has more than 1 content, then should we return from here as this is a complex use case for the tool? This mean we wont ask for question, but we will still run `vendir sync`

cli/pkg/kctrl/cmd/app/init/configure/template/ytt.go
```diff
@@ -0,0 +1,89 @@
...
+	yttTemplateStep.appBuild.Spec.App.Spec.Template = append(yttTemplateStep.appBuild.Spec.App.Spec.Template,
+		v1alpha1.AppTemplate{Ytt: &v1alpha1.AppTemplateYtt{}})
+	yttTemplateStep.appBuild.Save()
+}
+
+func (yttTemplateStep *YttTemplateStep) PostInteract() error {
+	existingTemplates := yttTemplateStep.appBuild.Spec.App.Spec.Template
+	fetchSource := yttTemplateStep.appBuild.ObjectMeta.Annotations[fetch.FetchContentAnnotationKey]
+	//TODO how to handle if from local directory?
+	if fetchSource == fetch.FetchFromLocalDirectory {
```
rohitagg2020: For local directory, I am just initialising yttTemplate if it doesn't exist. I think thats's enough. Do we see something more needs to be done ?

cli/pkg/kctrl/cmd/app/init/configure/template/ytt.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil
+	}
+
+	yttTemplateStep.addUpstreamAsPathIfNotExist()
+	return nil
+}
+
+func (yttTemplateStep *YttTemplateStep) addUpstreamAsPathIfNotExist() {
+	//TODO I am consciously adding upstream folder. Should we do that even in scenario where user doesn't want explicitly. Cant think of the scenario though.
+	appTemplates := yttTemplateStep.appBuild.Spec.App.Spec.Template
```
rohitagg2020: I am consciously adding upstream folder. Should we do that even in scenarios where a user doesn't want explicitly? Can't think of the scenario though.

cli/pkg/kctrl/cmd/app/init/init.go
```diff
@@ -0,0 +1,336 @@
...
+func (createStep CreateStep) createAppFromAppBuild() kcv1alpha1.App {
+	//TODO Should we ask for the app Name ?
+	appName := "microservices-demo"
+	serviceAccountName := fmt.Sprintf("sa-%s", appName)
+	appAnnotation := map[string]string{
+		fetch.LocalFetchAnnotationKey: ".",
+	}
+	appTemplateSection := createStep.appBuild.Spec.App.Spec.Template
+	//TODO should we remove the fetch section as it is not beind used and add it dynamically during dev deploy.
+	appFetchSection := []kcv1alpha1.AppFetch{kcv1alpha1.AppFetch{
```
rohitagg2020: I am not adding the fetchSection to generated AppBuild as this is not being used. Should we add it?

cli/pkg/kctrl/cmd/app/init/init.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+func (createStep CreateStep) configureExportSection() {
+	fetchSource := createStep.appBuild.ObjectMeta.Annotations[fetch.FetchContentAnnotationKey]
+	if fetchSource == fetch.FetchFromLocalDirectory {
+		return
+	}
+
+	// TODO current implementation is if export section is already defined, we will not touch it. Confirm the same.
+	if createStep.appBuild.Spec.Export == nil || len(createStep.appBuild.Spec.Export) == 0 {
```
rohitagg2020: If user has sync'ed vendir configuration from somewhere and then in the next run, they remove the upstream folder from includePaths, should we add it or leave it removed? Similarly, If they move to LocalDirectory in the next run, we should remove it, right?

cli/pkg/kctrl/cmd/app/init/init.go
```diff
@@ -0,0 +1,336 @@
...
+	appTemplates = append(appTemplates, oldAppTemplates[index:]...)
+
+	return
+}
+
+func addUpstreamAsPathToHelmIfNotExist(appTemplates []kcv1alpha1.AppTemplate) {
+	for _, appTemplate := range appTemplates {
+		if appTemplate.HelmTemplate != nil {
+			path := appTemplate.HelmTemplate.Path
+			//If a helmTemplate exist, it will always be the first one in the template section. Theoretically, it can exist anywhere but every real use case needs it to be first.
```
rohitagg2020: If a helmTemplate exist, it will always be the first one in the template section. Theoretically, it can exist anywhere but every real use case needs it to be first. Not handled the scenario if helmTemplate exists as not the first element of AppTemplate slice.  Also I think we cant have two helmTemplate section in the AppTemplate as there is no StdIn support for helmTemplate

cli/pkg/kctrl/cmd/package/init/init.go
```diff
@@ -0,0 +1,431 @@
...
+	if len(existingPkgInstall.Spec.PackageRef.RefName) == 0 {
+		existingPkgInstall.Spec.PackageRef.RefName = createStep.pkg.Spec.RefName
+	}
+	return nil
+}
+
+func (createStep CreateStep) updatePackage() error {
+	existingPkg := createStep.pkg
+
+	// TODO What if user has explicitly given the version anything else from 0.0.0?
```
rohitagg2020: What if user has explicitly given the version anything else from 0.0.0 in case of `pkg init`?

cli/pkg/kctrl/cmd/package/init/init.go
```diff
@@ -0,0 +1,431 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func getPackage(configs cmdlocal.Configs) (*v1alpha1.Package, error) {
+	var pkg v1alpha1.Package
+	// TODO should we handle this condition?
+	if len(configs.Pkgs) > 1 {
```
rohitagg2020: While reading, if I am getting > 1 packages, I am returning an error

cli/pkg/kctrl/cmd/app/init/configure/fetch/vendir.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/configure/fetch/fetch.go
rohitagg2020: Reason why we are having AppBuild in the Fetch step: Fetch step populates the AppBuild annotation(CurrentFetchOptionSelected). This annotation is being used later in the Template step as well to decide whether to add `helmTemplate` or not. We are passing AppBuild even in case of Template Step.

The alternative is to expose a method which returns the current fetch option. This method can be used by init.go to pass this information while instantiating `Template Step` in the init.go.

cli/pkg/kctrl/cmd/app/init/configure/fetch/vendir.go
cppforlife: lets just return an error. i dont think we should run vendir sync (at least at this point).

cli/pkg/kctrl/cmd/app/init/configure/fetch/vendir.go
cppforlife: (same as above comment)

cli/pkg/kctrl/cmd/package/init/init.go
cppforlife: sounds good.

cli/pkg/kctrl/cmd/package/init/init.go
cppforlife: i think we should just use that version then.

cli/pkg/kctrl/cmd/app/init/init.go
cppforlife: well... sometimes there might be sops, for example as a first step.

cli/pkg/kctrl/cmd/package/init/init.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/init/init.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/configure/fetch/vendir.go
rohitagg2020: Adopted.

cli/pkg/kctrl/cmd/app/init/configure/fetch/vendir.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/configure/fetch/vendir.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/init/init.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/init/init.go
rohitagg2020: Adopted

rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/configure/template/ytt.go
rohitagg2020: If appTemplate section is defined already, we will not touch it and return immediately

cli/pkg/kctrl/cmd/app/init/init.go
rohitagg2020: We are not gonna allow transition from one fetch source to another.

cli/pkg/kctrl/cmd/app/init/init.go
rohitagg2020: If appTemplate section is defined already, we will not touch it and return immediately.

cli/pkg/kctrl/cmd/app/init/configure/template/ytt.go
rohitagg2020: Thats enough

cli/pkg/kctrl/cmd/app/init/common/constants.go
rohitagg2020: Adopted

cli/examples/dynatrace/package-resources.yml
```diff
@@ -0,0 +1,44 @@
+apiVersion: data.packaging.carvel.dev/v1alpha1
+kind: Package
+metadata:
+  creationTimestamp: "2022-07-12T15:55:33Z"
```
cppforlife, line 4: can we avoid having creationTimestamp entirely (date or null)?

cli/examples/dynatrace/package-resources.yml
```diff
@@ -0,0 +1,44 @@
...
+  name: dynatrace.corp.com.0.0.0
+spec:
+  refName: dynatrace.corp.com
+  releasedAt: "2022-07-12T15:55:33Z"
```
cppforlife, line 8: lets not fill the releasedAt since package-resources.yml is "dev copy".

cli/examples/dynatrace/package-resources.yml
```diff
@@ -0,0 +1,44 @@
...
+  template:
+    spec:
+      template:
+      - ytt:
+          paths:
+          - upstream
+      - kbld: {}
+  valuesSchema:
+    openAPIv3: null
```
cppforlife, line 17: can we also remove this (in the generation code)?

cli/pkg/kctrl/cmd/app/init/app_build.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+type Export struct {
+	ImgpkgBundle *ImgpkgBundle `json:"imgpkgBundle,omitempty"`
+	IncludePaths []string      `json:"includePaths,omitempty"`
+}
+
+type Release struct {
+	Resource *ReleaseResource `json:"resource,omitempty"`
+	Yaml     interface{}      `json:"yaml,omitempty"`
```
cppforlife: nit: by go convention, this should be Yaml => `YAML`

cppforlife: type should be *ReleaseYAML similar to *ReleaseResource.

cli/pkg/kctrl/cmd/app/init/file_util.go
```diff
@@ -0,0 +1,37 @@
...
+
+// Check if file exists
+func IsFileExists(filePath string) (bool, error) {
+	_, err := os.Stat(filePath)
+	if err == nil {
+		return true, nil
+	} else if os.IsNotExist(err) {
+		return false, nil
+	} else {
+		return false, fmt.Errorf("failed to check for the existence of file. Error is: %s", err.Error())
```
cppforlife: fmt.Errorf("failed to check for the existence of file. Error is: %s", err.Error())
=> 
fmt.Errorf("Checking file: %s", err)

cli/pkg/kctrl/cmd/app/init/git.go
```diff
@@ -0,0 +1,101 @@
...
+	gitContent.URL = strings.TrimSpace(name)
+	return SaveVendir(gitStep.vendirConfig)
+}
+
+func (gitStep *GitStep) configureGitRef() error {
+	gitStep.ui.PrintInformationalText("A git reference can be any branch, tag, commit; origin is the name of the remote (required)")
+	gitContent := gitStep.vendirConfig.Directories[0].Contents[0].Git
+	defaultRef := gitContent.Ref
+	if defaultRef == "" {
+		defaultRef = "origin/develop"
```
cppforlife: develop => main (since that's what github uses by default)

cli/pkg/kctrl/cmd/app/init/helm.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return err
+	}
+	return helmStep.configureHelmChartVersion()
+}
+
+func (helmStep *HelmStep) initializeHelmRelease() error {
+	helmChartContent := vendirconf.DirectoryContentsHelmChart{
+		// TODO if the binary name is helm2, then it wont work.
+		HelmVersion: "3",
```
cppforlife: no need for this

cli/pkg/kctrl/cmd/app/init/init.go
```diff
@@ -0,0 +1,230 @@
...
+		appContent, err := yaml.Marshal(appConfig)
+		if err != nil {
+			return err
+		}
+		err = WriteFile(AppFileName, appContent)
+		if err != nil {
+			return err
+		}
+
+		createStep.ui.PrintInformationalText(fmt.Sprintf("Both the files can be accessed from the following location:"))
```
cppforlife: what locations?

cli/pkg/kctrl/cmd/app/init/init.go
```diff
@@ -0,0 +1,230 @@
...
+	ui.PrintInformationalText("\nCreated app can be consumed in following ways:\n")
+}
+
+func printInformation(ui cmdcore.AuthoringUI) {
+	ui.PrintInformationalText("\n**Information**\napp-build.yml is generated as part of this flow. This file can be used for further updating and adding complex scenarios while using the `kctrl dev deploy` command. Please read the link'ed documentation for more explanation.")
+}
+
+func (createStep CreateStep) createAppFromAppBuild() kcv1alpha1.App {
+	//TODO Should we ask for the app Name ?
+	appName := "microservices-demo"
```
cppforlife: we should ask for the name, default: directory name.

cli/pkg/kctrl/cmd/app/init/vendir.go
```diff
@@ -0,0 +1,228 @@
...
+		Stderr: stderrBs.String(),
+	}
+	if result.Error != nil {
+		return fmt.Errorf("Printing file: %s", result.Stderr)
+	}
+
+	v.ui.PrintCmdExecutionOutput(result.Stdout)
+
+	return nil
+}
```
cppforlife: lets not run commands to print vendir.yml. if we want to show it, then lets just use regular read file functions. cat may not be available in all environments (e.g. windows)

cli/pkg/kctrl/cmd/app/init/vendir.go
```diff
@@ -0,0 +1,228 @@
...
+
+	result := exec.CmdRunResult{
+		Stdout: stdoutBs.String(),
+		Stderr: stderrBs.String(),
+	}
+	if result.Error != nil {
+		return fmt.Errorf("Listing files.\n %s", result.Stderr)
+	}
+	v.ui.PrintCmdExecutionOutput(result.Stdout)
+	return nil
```
cppforlife, line 210: i dont think we need to list files especially if we take a dependence on ls binary.

cli/pkg/kctrl/cmd/app/init/vendir_util.go
```diff
@@ -0,0 +1,82 @@
...
+	err = yaml.Unmarshal(content, &vendirConfig)
+	if err != nil {
+		return vendirconf.Config{}, err
+	}
+	return vendirConfig, nil
+}
+
+func NewDefaultVendirConfig() vendirconf.Config {
+	config := vendirconf.Config{
+		APIVersion: "vendir.k14s.io/v1alpha1", // TODO: use constant from vendir package
```
cppforlife: > // TODO: use constant from vendir package

the less constants the better. it's an anti-pattern to pull out strings used only in one place into constants since it introduces unnecessary indirection.

cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
```diff
@@ -0,0 +1,135 @@
...
+				},
+			},
+		},
+		Template: b.opts.BuildTemplate,
+		Deploy:   b.opts.BuildDeploy,
+	}
+	if useKbldImagesLock {
+		for _, templateStage := range appSpec.Template {
+			if templateStage.Kbld != nil {
+				templateStage.Kbld.Paths = append(templateStage.Kbld.Paths, "-")
```
cppforlife: append can take multiple parameters

cppforlife: also i dont think we should be appending, but rather setting directly.

cli/pkg/kctrl/cmd/app/release/imgpkg_runner.go
```diff
@@ -0,0 +1,99 @@
...
+
+type ImgpkgRunner struct {
+	BundlePath        string
+	Paths             []string
+	UseKbldImagesLock bool
+	ImgLockFilepath   string
+	UI                cmdcore.AuthoringUI
+}
+
+func (r ImgpkgRunner) Run() (string, error) {
```
cppforlife, line 26: already mentioned elsewhere regarding copy implementation

cli/pkg/kctrl/cmd/app/release/imgpkg_runner.go
```diff
@@ -0,0 +1,99 @@
...
+	cmd.Stdout = io.MultiWriter(inMemoryStdoutWriter)
+	cmd.Stderr = &stderrBuf
+	err = cmd.Run()
+	if err != nil {
+		return stdoutBuf.String(), fmt.Errorf("%s", stderrBuf.String())
+	}
+	err = inMemoryStdoutWriter.Flush()
+	if err != nil {
+		return "", err
+	}
```
cppforlife: why are we flushing stdout?

cli/pkg/kctrl/cmd/app/release/release.go
```diff
@@ -0,0 +1,90 @@
...
+	logger      logger.Logger
+
+	pkgVersion     string
+	chdir          string
+	outputLocation string
+	debug          bool
+}
+
+const (
+	defaultArtifactDir = "carvel-arteficts"
```
cppforlife: arteficts => artifacts

cli/pkg/kctrl/cmd/app/release/release.go
```diff
@@ -0,0 +1,90 @@
...
+	defaultVersion     = "0.0.0-%d"
```
cppforlife: 0.0.0-%d => 0.0.0+build.%d

cli/pkg/kctrl/cmd/app/release/release.go
```diff
@@ -0,0 +1,90 @@
...
+	cmd := &cobra.Command{
+		Use:   "release",
+		Short: "Release package",
+		RunE:  func(cmd *cobra.Command, args []string) error { return o.Run() },
+	}
+
+	cmd.Flags().StringVarP(&o.pkgVersion, "version", "v", "", "Version to be released")
+	cmd.Flags().StringVar(&o.chdir, "chdir", "", "Working directory with package-build and other config")
+	cmd.Flags().StringVar(&o.outputLocation, "copy-to", defaultArtifactDir, "Output location for artifacts")
+	cmd.Flags().BoolVar(&o.debug, "debug", false, "Version to be released")
```
cppforlife: wrong description message

cli/pkg/kctrl/cmd/app/release/release_cmd_runner.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+var _ exec.CmdRunner = &ReleaseCmdRunner{}
+
+func NewReleaseCmdRunner(log io.Writer, fullOutput bool, tempImgLockFilepath string, ui cmdcore.AuthoringUI) *ReleaseCmdRunner {
+	return &ReleaseCmdRunner{log: log, fullOutput: fullOutput, tempImgLockFilepath: tempImgLockFilepath, ui: ui}
+}
+
+func (r ReleaseCmdRunner) Run(cmd *goexec.Cmd) error {
+	if strings.Contains(cmd.Path, "/vendir") {
```
cppforlife: filepath.Base(cmd.Path) == "vendir"

cli/pkg/kctrl/cmd/package/init/pkg-build.go
```diff
@@ -0,0 +1,121 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package init
+
+import (
+	"os"
+
+	appBuild "github.com/vmware-tanzu/carvel-kapp-controller/cli/pkg/kctrl/cmd/app/init"
```
cppforlife: per go conventions import directives should not be camelcase

cppforlife (commented):
havent gotten through everything, but you should be able to extrapolate based on these comments.

cli/pkg/kctrl/cmd/app/release/release.go
```diff
@@ -0,0 +1,90 @@
...
+
+	return cmd
+}
+
+func (o *ReleaseOptions) Run() error {
+	if o.chdir != "" {
+		err := os.Chdir(o.chdir)
```
rohitagg2020, line 53: Reconciller fetch has an issue if package-build.yml is not in the root dir.

cli/pkg/kctrl/cmd/app/init/helm.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/git.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/file_util.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/app_build.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/init/pkg-build.go
rohitagg2020: @sethiyash @100mik

cli/pkg/kctrl/cmd/app/init/vendir_util.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
100mik: Makes sense, I do a case for needing files added at the kbld stage exclusively if kbld is already picking up values from the ytt stage

cli/pkg/kctrl/cmd/app/release/release.go
100mik: Putting in here from discussion we had.

Chdir is not to be used in most cases, and when used it is reasonable to expect config files to reside in the folder itself.
Since, our primary intention here was to move the files created by us to a folder that is not the root, we should be doing that by default if that seems to be preferred.

cli/examples/dynatrace/package-resources.yml
100mik: It seems to be an `omitempty` field as it should be not sure why that is not being respected.
Will take  a closer look 🤔

cli/examples/dynatrace/package-resources.yml
100mik: Now that we are generating the schema, if there is no schema provided. The value is:
```
openAPIv3:
  default: null
  nullable: true
```
Which essentially means it is of type "any".
You think this should be removed as well in the default case?

cli/pkg/kctrl/cmd/app/release/imgpkg_runner.go
100mik: Seems to be dead code from when I was trying to put together something that streams imgpkg output, removing this

cli/pkg/kctrl/cmd/app/init/app_build.go
```diff
@@ -0,0 +1,155 @@
+// Copyright 2020 VMware, Inc.
```
praveenrewar: nit: Should we change the year to 2022 in all the files?

cli/pkg/kctrl/cmd/app/init/app_build.go
```diff
@@ -0,0 +1,155 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package init
+
+import (
+	"os"
+	"path/filepath"
+
+	v1alpha12 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
```
praveenrewar: v1alpha12 -> kcv1alpha1

cli/pkg/kctrl/cmd/app/init/app_build.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/apis/datapackaging/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	AppBuildFileName = "app-build.yml"
+
+	AppBuildAPIVersion = "kctrl.carvel.dev/v1alpha1"
+	AppBuildKind       = "AppBuild"
```
praveenrewar: Should we _not_ export them?
Also, file name is app_build.go and since we are not using them in any other package, should we just call them `fileName`, `apiVersion` and `kind` only?

cli/pkg/kctrl/cmd/app/init/app_build.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+type AppBuild struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              Spec `json:"spec,omitempty"`
```
praveenrewar: nit: add newline

cli/pkg/kctrl/cmd/app/init/app_build.go
```diff
@@ -0,0 +1,155 @@
...
+
+type ImgpkgBundle struct {
+	Image             string `json:"image,omitempty"`
+	UseKbldImagesLock bool   `json:"useKbldImagesLock,omitempty"`
+}
+
+type ReleaseResource struct {
+}
+
+// Save will persist the appBuild onto the fileSystem. Before saving, it will remove the Annotations from the AppBuild.
```
praveenrewar, line 53: `Before saving, it will remove the Annotations from the AppBuild.` seems redundant with the comment on the first line of the method.

cli/pkg/kctrl/cmd/app/init/app_build.go
```diff
@@ -0,0 +1,155 @@
...
+func (appBuild AppBuild) Save() error {
```
praveenrewar: appBuild -> b
or
appBuild -> ab

cli/pkg/kctrl/cmd/app/init/app_build.go
```diff
@@ -0,0 +1,155 @@
...
+		return err
+	}
+
+	return WriteFile(AppBuildFileName, content)
+}
+
+func NewAppBuild() (AppBuild, error) {
+	var appBuild AppBuild
+	appBuildFilePath := filepath.Join(AppBuildFileName)
+	exists, err := IsFileExists(appBuildFilePath)
```
praveenrewar: nit: inline
Also, if only one string is passed to filepath.Join(), it would return the string as it is. Are we planning to include directories as well in the future?

cli/pkg/kctrl/cmd/app/init/app_build.go
```diff
@@ -0,0 +1,155 @@
...
+		return AppBuild{}, err
+	}
+
+	if exists {
+		appBuild, err = NewAppBuildFromFile(appBuildFilePath)
+		if err != nil {
+			return AppBuild{}, err
+		}
+
+		//In case user has manually removed the app section from the app-build
```
praveenrewar: nit: space after `//`

cli/pkg/kctrl/cmd/app/init/app_build.go
```diff
@@ -0,0 +1,155 @@
...
+		if appBuild.Spec.App == nil {
+			defaultApp := NewDefaultAppTemplateSpec()
+			appBuild.Spec.App = defaultApp
```
praveenrewar: inline

cli/pkg/kctrl/cmd/app/init/app_build.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+	} else {
```
praveenrewar, line 82: Maybe we can do something like this?
```go
if !exists {
       return NewDefaultAppBuild(), nil
}
// do something if file exists
```

We wouldn't also need the initial declaration then `var appBuild AppBuild`

cli/pkg/kctrl/cmd/app/init/app_build.go
```diff
@@ -0,0 +1,155 @@
...
+	content, err := os.ReadFile(filePath)
+	if err != nil {
+		return AppBuild{}, err
+	}
+	var appBuild AppBuild
+	err = yaml.Unmarshal(content, &appBuild)
+	if err != nil {
+		return AppBuild{}, err
+	}
+	return appBuild, nil
```
praveenrewar:
```suggestion
	return appBuild, err
```

cli/pkg/kctrl/cmd/app/init/app_build.go
```diff
@@ -0,0 +1,155 @@
...
+
+func NewDefaultAppTemplateSpec() *v1alpha1.AppTemplateSpec {
+	appSpec := v1alpha12.AppSpec{
+		Fetch:    []v1alpha12.AppFetch{},
+		Template: []v1alpha12.AppTemplate{},
+		Deploy: []v1alpha12.AppDeploy{
+			{Kapp: &v1alpha12.AppDeployKapp{}},
+		},
+	}
+	return &v1alpha1.AppTemplateSpec{&appSpec}
```
praveenrewar: nit: inline

cli/pkg/kctrl/cmd/app/init/app_build.go
```diff
@@ -0,0 +1,155 @@
...
+	appBuild := AppBuild{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       AppBuildKind,
+			APIVersion: AppBuildAPIVersion,
+		},
+		Spec: Spec{
+			App: NewDefaultAppTemplateSpec(),
+		},
+	}
+	return appBuild
```
praveenrewar: nit: inline

cli/pkg/kctrl/cmd/app/init/app_build.go
```diff
@@ -0,0 +1,155 @@
...
+
+func (appBuild *AppBuild) SetAppSpec(appSpec *v1alpha12.AppSpec) {
+	if appBuild.Spec.App == nil {
+		appBuild.Spec.App = &v1alpha1.AppTemplateSpec{}
+	}
+	appBuild.Spec.App.Spec = appSpec
+}
+
+func (appBuild AppBuild) GetObjectMeta() *metav1.ObjectMeta {
+	return nil
```
praveenrewar: inline

cli/pkg/kctrl/cmd/app/init/constants.go
```diff
@@ -0,0 +1,17 @@
...
+	LocalFetchAnnotationKey   = "kctrl.carvel.dev/local-fetch-0"
+)
+
+const (
+	FetchFromGithubRelease  string = "Github Release"
+	FetchManifestFromGit    string = "Git Repository"
+	FetchChartFromHelmRepo  string = "Helm Chart from Helm Repository"
+	FetchChartFromGit       string = "Helm Chart from Git Repository"
+	FetchFromLocalDirectory string = "Local Directory"
+)
```
praveenrewar: Seems like these would be better suited in `fetch.go`.
Also, maybe we can have a type for the different fetch options?

cli/pkg/kctrl/cmd/app/init/fetch.go
```diff
@@ -0,0 +1,108 @@
...
+	isAppCommandRunExplicitly bool
+}
+
+func NewFetchStep(ui cmdcore.AuthoringUI, build Build, isAppCommandRunExplicitly bool) *FetchStep {
+	fetchStep := FetchStep{
+		ui:                        ui,
+		build:                     build,
+		isAppCommandRunExplicitly: isAppCommandRunExplicitly,
+	}
+	return &fetchStep
```
praveenrewar: nit: inline

cli/pkg/kctrl/cmd/app/init/fetch.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (fetchStep *FetchStep) PreInteract() error {
+	return nil
```
praveenrewar: nit: inline

cli/pkg/kctrl/cmd/app/init/fetch.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (fetchStep *FetchStep) Interact() error {
+	fetchStep.ui.PrintHeaderText("Content (Step 2/3)")
+	fetchStep.ui.PrintInformationalText("We need to fetch the manifest which defines how the package would be deployed in a K8s cluster. This manifest can be in the form of a yaml file used with `kubectl apply ...` or it could be a helm chart used with `helm install ...`. They can be available in any of the following locations. Please select one among them")
+
+	var vendirConfig vendirconf.Config
```
praveenrewar: not needed

cli/pkg/kctrl/cmd/package/release/release.go
100mik: We do have the temp directly at a later stage, when we are bundling. The lock output is actually generated while building the images using the reconciler.
So this folder merely keeps the lock file till it is copied to the temp bundle folder if it is an imgpkg export.

We do create a temp directory if it is an imgpkg bundle export. We could do this at an earlier stage as well. I do agree that we should probably not create files in the root directory.

cli/pkg/kctrl/cmd/package/release/imgpkg_runner.go
100mik: Using the lib now!

cli/pkg/kctrl/cmd/app/init/app_build.go
100mik: Agreed we should do this

cli/pkg/kctrl/cmd/app/init/app_build.go
100mik: Agreed, should be uniform across the project

cli/pkg/kctrl/cmd/app/init/app_build.go
100mik: I would go with `b`

cli/examples/dynatrace/package-resources.yml
rohitagg2020: I am adding it explicitly. Will remove it.

cli/pkg/kctrl/cmd/app/init/fetch.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/fetch.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/fetch.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/app_build.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/app_build.go
rohitagg2020: moved apiversion and kind inline. Renamed fileName

cli/pkg/kctrl/cmd/app/init/app_build.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/app_build.go
rohitagg2020: I want to highlight that `it will remove the Annotations from the AppBuild.`

cli/pkg/kctrl/cmd/app/init/app_build.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/app_build.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/app_build.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/app_build.go
rohitagg2020: If we do as above, then I am creating the file and then reading it. I think it is ok to have a separate if/else block based on file existence.

cli/pkg/kctrl/cmd/app/init/app_build.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/app_build.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/app_build.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/app_build.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/app_build.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/vendir.go
rohitagg2020: By doing this, I think that we are slowly moving towards abstraction rather than being transparent about what is happening under the hood.

cli/pkg/kctrl/cmd/app/init/app_build.go
praveenrewar: > If we do as above, then I am creating the file and then reading it.

I didn't understand this. (I honestly don't seen any difference in both the implementations, apart from readability)

100mik: >If we do as above, then I am creating the file and then reading it. I think it is ok to have a separate if/else block based on file existence.

Praveen means that we can use a return statement to return early if `!exists`. Makes sense to me!

cli/pkg/kctrl/cmd/app/init/init.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/vendir.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/init/init.go
rohitagg2020: I am not able to understand it, can you please elaborate it.

cli/pkg/kctrl/cmd/app/init/constants.go
rohitagg2020: As they are being used in vendir.go as well, hence not moving to fetch.go.

cli/pkg/kctrl/cmd/app/init/app_build.go
```diff
@@ -0,0 +1,149 @@
...
+import (
+	kcv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/apis/datapackaging/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"os"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	FileName = "app-build.yml"
```
praveenrewar, line 16: FileName is being used in the same package (mostly). There is only use other place where it is being passed as a parameter which doesn't make much sense because it's a constant. Maybe we should not export this as well?

cli/pkg/kctrl/cmd/app/init/file_util.go
```diff
@@ -0,0 +1,37 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package init
```
praveenrewar, line 4: Maybe this should be in cmdcore? And scoped to a struct?

cli/pkg/kctrl/cmd/app/init/file_util.go
```diff
@@ -0,0 +1,37 @@
...
+// IsFileExists checks if file exists
+func IsFileExists(filePath string) (bool, error) {
+	_, err := os.Stat(filePath)
+	if err == nil {
+		return true, nil
+	} else if os.IsNotExist(err) {
+		return false, nil
+	} else {
+		return false, fmt.Errorf("Checking file: %s", err.Error())
+	}
```
praveenrewar:
```suggestion
	if err == nil {
		return true, nil
	} 
	if os.IsNotExist(err) {
		return false, nil
	} 
	return false, fmt.Errorf("Checking file: %s", err.Error())
```

cli/pkg/kctrl/cmd/app/init/file_util.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// WriteFile writes binary content to file
+func WriteFile(filePath string, data []byte) error {
+	// Create creates or truncates the named file. If the file already exists, it is truncated.
```
praveenrewar: nit: Comment not required.

cli/pkg/kctrl/cmd/app/init/file_util.go
```diff
@@ -0,0 +1,37 @@
...
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	_, err = file.Write(data)
+	if err != nil {
+		return err
+	}
+	return nil
```
praveenrewar:
```suggestion
	return err
```

cli/pkg/kctrl/cmd/app/init/interfaces.go
```diff
@@ -0,0 +1,37 @@
...
+	GetObjectMeta() *metav1.ObjectMeta
+	SetObjectMeta(*metav1.ObjectMeta)
+	SetExport(export *[]Export)
+	GetExport() *[]Export
+}
+
+type Step interface {
+	PreInteract() error
+	Interact() error
+	PostInteract() error
```
praveenrewar, line 24: Should we just have one function called `Run()` because `PreInteract()` doesn't seem to be doing anything for all the steps and mostly only one step is doing "something" in most of the steps. Another reason to have a single `Run()` method is that there's no clear picture as to what each of these different stages mean.

cli/pkg/kctrl/cmd/app/init/fetch.go
```diff
@@ -0,0 +1,102 @@
...
+		ui:                        ui,
+		build:                     build,
+		isAppCommandRunExplicitly: isAppCommandRunExplicitly,
+	}
+}
+
+func (f *FetchStep) PreInteract() error { return nil }
+
+func (f *FetchStep) Interact() error {
+	f.ui.PrintHeaderText("Content (Step 2/3)")
```
praveenrewar, line 43: It seems a bit weird to start with Step 2, maybe we should have a common place from where all the steps are run?

cli/pkg/kctrl/cmd/app/init/fetch.go
```diff
@@ -0,0 +1,102 @@
...
+
+	options := []string{FetchFromLocalDirectory, FetchFromGithubRelease, FetchFromHelmRepo, FetchFromGit, FetchChartFromGit}
+	previousFetchOptionIndex := getPreviousFetchOptionIndex(options, previousFetchOptionSelected)
+	defaultFetchOptionIndex := previousFetchOptionIndex
+	choiceOpts := ui.ChoiceOpts{
+		Label:   "Enter configuration source",
+		Default: defaultFetchOptionIndex,
+		Choices: options,
+	}
+	currentFetchOptionIndex, err := f.ui.AskForChoice(choiceOpts)
```
praveenrewar, line 61: currentFetchOptionIndex -> selectedOpt ?

cli/pkg/kctrl/cmd/app/init/fetch.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return err
+	}
+	currentFetchOptionSelected := options[currentFetchOptionIndex]
+
+	if previousFetchOptionSelected != "" && currentFetchOptionSelected != previousFetchOptionSelected {
```
praveenrewar, line 67: If we are not allowing to transition from one fetch option to another, would it make sense to just use the previous choice if it's present and not ask the user anything instead of asking and then erroring out saying transition is not allowed?

cli/pkg/kctrl/cmd/app/release/release.go
```diff
@@ -0,0 +1,90 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package release
+
+import (
+	"os"
+
+	"github.com/cppforlife/go-cli-ui/ui"
+	"github.com/spf13/cobra"
+	cmdappbuild "github.com/vmware-tanzu/carvel-kapp-controller/cli/pkg/kctrl/cmd/app/init"
```
praveenrewar: package name is init but we are calling it cmdappbuild?

cli/pkg/kctrl/cmd/package/release/release.go
```diff
@@ -0,0 +1,194 @@
...
+func (o *ReleaseOptions) loadExportData(pkgBuild *cmdpkgbuild.PackageBuild) error {
+	if len(pkgBuild.Spec.Template.Spec.Export) == 0 {
+		pkgBuild.Spec.Template.Spec.Export = []appinit.Export{
+			{
+				ImgpkgBundle: &appinit.ImgpkgBundle{
+					UseKbldImagesLock: true,
+				},
+			},
+		}
+	}
```
praveenrewar: We could try to make this inline?

cli/pkg/kctrl/cmd/package/release/release.go
```diff
@@ -0,0 +1,194 @@
...
+		return err
+	}
+
+	// To be removed and moved to a question in case we have more config/variations around this
+	if pkgBuild.Spec.Release == nil || len(pkgBuild.Spec.Release) == 0 {
+		pkgBuild.Spec.Release = []appinit.Release{
+			{
+				Resource: &appinit.ReleaseResource{},
+			},
+		}
```
praveenrewar: inline

cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
```diff
@@ -0,0 +1,134 @@
...
+	depsFactory      cmdcore.DepsFactory
+	logger           logger.Logger
+	ui               cmdcore.AuthoringUI
+	opts             AppSpecBuilderOpts
+}
+
+type AppSpecBuilderOpts struct {
+	BuildTemplate []kcv1alpha1.AppTemplate
+	BuildDeploy   []kcv1alpha1.AppDeploy
+	BuildExport   []appinit.Export
```
praveenrewar, line 30: Maybe these should just be `Template`, `Deploy` and `Export`?

cli/pkg/kctrl/cmd/app/init/fetch.go
```diff
@@ -0,0 +1,102 @@
...
+		return fmt.Errorf("Transitioning from one fetch option to another is not allowed. Earlier option selected: %s, Current Option selected: %s", previousFetchOptionSelected, currentFetchOptionSelected)
+	}
+
+	buildObjectMeta := f.build.GetObjectMeta()
+	if buildObjectMeta.Annotations == nil {
+		buildObjectMeta.Annotations = make(map[string]string)
+	}
+	buildObjectMeta.Annotations[FetchContentAnnotationKey] = currentFetchOptionSelected
+	f.build.SetObjectMeta(buildObjectMeta)
+	// For a local directory, we will be including everything.
```
praveenrewar: Should we make the comment more descriptive? 
(Right now we are saying that we will include everything for a local directory and then we return nil, so it's not very evident as to how we will include everything.)

cli/pkg/kctrl/cmd/app/init/git.go
```diff
@@ -0,0 +1,99 @@
...
+			return err
+		}
+	}
+
+	err := g.configureGitURL()
+	if err != nil {
+		return err
+	}
+	return g.configureGitRef()
+	return nil
```
praveenrewar: unreachable code

cli/pkg/kctrl/cmd/app/init/git.go
```diff
@@ -0,0 +1,99 @@
...
+		err := g.initializeContentWithGit()
+		if err != nil {
+			return err
+		}
+	} else if contents[0].Git == nil {
+		err := g.initializeGit()
+		if err != nil {
+			return err
+		}
+	}
```
praveenrewar, line 40: How about?
```suggestion
	if contents == nil {
		g.vendirConfig.Directories[0].Contents = append(g.vendirConfig.Directories[0].Contents, vendirconf.DirectoryContents{})
	} 
	err := g.initializeGit()
	if err != nil {
		return err
	}
```

cli/pkg/kctrl/cmd/app/init/git.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func (g *GitStep) initializeGit() error {
+	g.vendirConfig.Directories[0].Contents[0].Git = &vendirconf.DirectoryContentsGit{}
+	return SaveVendir(g.vendirConfig)
+}
+
+func (g *GitStep) configureGitURL() error {
+	gitContent := g.vendirConfig.Directories[0].Contents[0].Git
+	defaultURL := gitContent.URL
```
praveenrewar, line 64: could be inline

cli/pkg/kctrl/cmd/app/init/interfaces.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+type Build interface {
+	Save() error
+	GetAppSpec() *v1alpha12.AppSpec
+	SetAppSpec(*v1alpha12.AppSpec)
+	GetObjectMeta() *metav1.ObjectMeta
+	SetObjectMeta(*metav1.ObjectMeta)
+	SetExport(export *[]Export)
+	GetExport() *[]Export
```
praveenrewar, line 18: We typically don't use getters and setters in Go. We can have `AppSpec` instead of `GetAppSpec` and so on.
Also, we are using the `.` notation to get the required variables from Build in a couple of places which kind of defeats the purpose of having a method for just _get_.

praveenrewar (commented):
Still need to go through a couple of more files.

cli/pkg/kctrl/cmd/app/init/interfaces.go
100mik: We will be refactoring after achieving current milestones 🚀

cli/pkg/kctrl/cmd/app/init/app_build.go
100mik: I think this part gets resolved when we move away from the interfaces and unstopped functions we use to interact with files. So I am leaving this for now.
(cc: @praveenrewar correct me if I am wrong)

cli/pkg/kctrl/cmd/app/init/file_util.go
100mik: I think we won't need this at all? 
(Judging from our conclusions when we were refactoring)

Leaving it up.

cli/pkg/kctrl/cmd/app/init/fetch.go
100mik: This looks like work for refactoring as well

cli/pkg/kctrl/cmd/app/init/app_build.go
100mik: The other use seems to be in another package though

cli/hack/test-all-local-registry.sh
```diff
@@ -0,0 +1,15 @@
...
+
+function cleanup {
+  docker stop registry-"$PORT"
+  docker rm -v registry-"$PORT"
+}
+trap cleanup EXIT
+
+docker run -d -p "$PORT":5000 -e REGISTRY_VALIDATION_MANIFESTS_URLS_ALLOW='- ^https?://' --restart always --name registry-"$PORT" registry:2
+export KCTRL_E2E_IMAGE="localhost:$PORT/local-tests/test-repo"
+./hack/test-all.sh $@
```
praveenrewar: nit: add newline at end.

cli/pkg/kctrl/cmd/app/init/app_build.go
praveenrewar: The other usage is to pass the the constant as a parameter to a function of this package. Seems a bit weird.

cli/pkg/kctrl/cmd/app/init/app_build.go
praveenrewar: Yeah, but we have mentioned that part twice.

cli/pkg/kctrl/cmd/app/init/fetch.go
praveenrewar: Maybe we can just move the print statements to wherever fetch is being used?

PR #785, opened 2022-07-13
Add sidecarexec sandboxing

cppforlife:
- executes fetcher/templaters in a limited environment (within a new user/mount namespace on linux)
- TODO we need to run test on all the clouds
  - [x] Minikube (gh action runs e2e in minikube; also use it locally)
  - [x] Kind (gh action runs e2e in kind)
  - [ ] GKE
  - [ ] GKE Autopilot
  - [ ] Azure
  - [ ] EKS
  - [ ] Openshift
  - [ ] TKG
- resolve outstanding TODOs in code

n/a

no

```release-note
- Execute fetcher/templates in a limited environment
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -5,9 +5,11 @@ package main
 
 import (
 	"flag"
+	"fmt"
 	"os"
 	"time"
 
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/sidecarexec"
 	"k8s.io/klog/v2"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
@@ -18,7 +20,7 @@ var Version = "develop"
 
 func main() {
 	ctrlOpts := Options{}
-	var sidecarexec bool
+	var isSidecarexec, isSidecarexecWrap bool
 
 	flag.IntVar(&ctrlOpts.Concurrency, "concurrency", 10, "Max concurrent reconciles")
 	flag.StringVar(&ctrlOpts.Namespace, "namespace", "", "Namespace to watch")
@@ -27,11 +29,21 @@ func main() {
 	flag.BoolVar(&ctrlOpts.EnablePprof, "dangerous-enable-pprof", false, "If set to true, enable pprof on "+PprofListenAddr)
 	flag.DurationVar(&ctrlOpts.APIRequestTimeout, "api-request-timeout", time.Duration(0), "HTTP timeout for Kubernetes API requests")
 	flag.BoolVar(&ctrlOpts.APIPriorityAndFairness, "enable-api-priority-and-fairness", true, "Enable/disable APIPriorityAndFairness feature gate for apiserver. Recommended to disable for <= k8s 1.19.")
-	flag.BoolVar(&sidecarexec, "sidecarexec", false, "Run sidecarexec")
+	flag.BoolVar(&isSidecarexec, "sidecarexec", false, "Run sidecarexec")
+	flag.BoolVar(&isSidecarexecWrap, "sidecarexecwrap", false, "Run sidecarexecwrap")
 	flag.Parse()
 
-	if sidecarexec {
+	if isSidecarexec {
 		sidecarexecMain()
+		return
+	}
+	if isSidecarexecWrap {
+		err := sidecarexec.SandboxWrap{}.ExecuteCmd(flag.Args())
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "sidecarexecwrap: Error: %s\n", err)
+			os.Exit(1)
+		}
+		os.Exit(0)
 		return
 	}
 
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -20,7 +20,7 @@ var Version = "develop"
 
 func main() {
 	ctrlOpts := Options{}
-	var isSidecarexec, isSidecarexecWrap bool
+	var isSidecarexec, isSidecarexecDebug, isSidecarexecWrap bool
 
 	flag.IntVar(&ctrlOpts.Concurrency, "concurrency", 10, "Max concurrent reconciles")
 	flag.StringVar(&ctrlOpts.Namespace, "namespace", "", "Namespace to watch")
@@ -30,11 +30,12 @@ func main() {
 	flag.DurationVar(&ctrlOpts.APIRequestTimeout, "api-request-timeout", time.Duration(0), "HTTP timeout for Kubernetes API requests")
 	flag.BoolVar(&ctrlOpts.APIPriorityAndFairness, "enable-api-priority-and-fairness", true, "Enable/disable APIPriorityAndFairness feature gate for apiserver. Recommended to disable for <= k8s 1.19.")
 	flag.BoolVar(&isSidecarexec, "sidecarexec", false, "Run sidecarexec")
+	flag.BoolVar(&isSidecarexecDebug, "sidecarexecdebug", false, "Run sidecarexecdebug")
 	flag.BoolVar(&isSidecarexecWrap, "sidecarexecwrap", false, "Run sidecarexecwrap")
 	flag.Parse()
 
-	if isSidecarexec {
-		sidecarexecMain()
+	if isSidecarexec || isSidecarexecDebug {
+		sidecarexecMain(isSidecarexecDebug, flag.Args())
 		return
 	}
 	if isSidecarexecWrap {
--- a/cmd/controller/sidecarexec.go
+++ b/cmd/controller/sidecarexec.go
@@ -4,6 +4,9 @@
 package main
 
 import (
+	"fmt"
+	"os"
+	goexec "os/exec"
 	"syscall"
 	"time"
 
@@ -13,16 +16,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
-func sidecarexecMain() {
-	mainLog := zap.New(zap.UseDevMode(false)).WithName("kc-sidecarexec")
-	mainLog.Info("start sidecarexec", "version", Version)
-
-	go reapZombies(mainLog)
-
+func sidecarexecMain(debug bool, debugArgs []string) {
 	localCmdRunner := exec.NewPlainCmdRunner()
 	sandboxCmdRunner := sidecarexec.NewSandboxCmdRunner(localCmdRunner, sidecarexec.SandboxCmdRunnerOpts{
 		RequiresPosix: map[string]bool{
 			"vendir": true,
+			"bash":   true, // for debugging
 		},
 		RequiresNetwork: map[string]bool{
 			"vendir": true,
@@ -30,7 +29,27 @@ func sidecarexecMain() {
 		},
 	})
 
-	opts := sidecarexec.ServerOpts{
+	if debug {
+		cmd := goexec.Command("bash")
+		if len(debugArgs) > 0 {
+			cmd = goexec.Command(debugArgs[0], debugArgs[1:]...)
+		}
+		cmd.Stdin = os.Stdin
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		err := sandboxCmdRunner.Run(cmd, exec.RunOpts{})
+		if err != nil {
+			fmt.Printf("Exit error: %s\n", err)
+		}
+		return
+	}
+
+	mainLog := zap.New(zap.UseDevMode(false)).WithName("kc-sidecarexec")
+	mainLog.Info("start sidecarexec", "version", Version)
+
+	go reapZombies(mainLog)
+
+	serverOpts := sidecarexec.ServerOpts{
 		AllowedCmdNames: []string{
 			// Fetch (calls impgkg and others internally)
 			"vendir",
@@ -38,8 +57,7 @@ func sidecarexecMain() {
 			"ytt", "kbld", "sops", "helm", "cue",
 		},
 	}
-
-	server := sidecarexec.NewServer(sandboxCmdRunner, opts, mainLog)
+	server := sidecarexec.NewServer(sandboxCmdRunner, serverOpts, mainLog)
 
 	err := server.Serve()
 	if err != nil {
--- a/cmd/controller/sidecarexec.go
+++ b/cmd/controller/sidecarexec.go
@@ -20,6 +20,16 @@ func sidecarexecMain() {
 	go reapZombies(mainLog)
 
 	localCmdRunner := exec.NewPlainCmdRunner()
+	sandboxCmdRunner := sidecarexec.NewSandboxCmdRunner(localCmdRunner, sidecarexec.SandboxCmdRunnerOpts{
+		RequiresPosix: map[string]bool{
+			"vendir": true,
+		},
+		RequiresNetwork: map[string]bool{
+			"vendir": true,
+			"kbld":   true,
+		},
+	})
+
 	opts := sidecarexec.ServerOpts{
 		AllowedCmdNames: []string{
 			// Fetch (calls impgkg and others internally)
@@ -29,7 +39,7 @@ func sidecarexecMain() {
 		},
 	}
 
-	server := sidecarexec.NewServer(localCmdRunner, opts, mainLog)
+	server := sidecarexec.NewServer(sandboxCmdRunner, opts, mainLog)
 
 	err := server.Serve()
 	if err != nil {
--- a/config/deployment.yml
+++ b/config/deployment.yml
@@ -15,6 +15,10 @@ spec:
   revisionHistoryLimit: 0
   template:
     metadata:
+      annotations:
+        #! Necessary at least on GKE (which includes a default apparmor policy)
+        #! which seems to prevent execution of mount syscall even within a user namespace
+        container.apparmor.security.beta.kubernetes.io/kapp-controller-sidecarexec: unconfined
       labels:
         app: kapp-controller
     spec:
@@ -66,6 +70,8 @@ spec:
             cpu: 120m
             memory: 100Mi
         env:
+        - name: KAPPCTRL_SIDECAREXEC_LIMIT_DIR
+          value: /etc/kappctrl-mem-tmp
         - name: KAPPCTRL_SIDECAREXEC_SOCK
           value: /etc/kappctrl-mem-tmp/sidecarexec.sock
         volumeMounts:
--- a/pkg/sidecarexec/sandbox_cmd_runner_darwin.go
+++ b/pkg/sidecarexec/sandbox_cmd_runner_darwin.go
@@ -1,0 +1,24 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package sidecarexec
+
+import (
+	goexec "os/exec"
+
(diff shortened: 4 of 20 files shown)


config/deployment.yml
```diff
@@ -15,6 +15,10 @@ spec:
   revisionHistoryLimit: 0
   template:
     metadata:
+      annotations:
+        #! Necessary at least on GKE (which includes a default apparmor policy)
+        #! which seems to prevent execution of mount syscall even within a user namespace
+        container.apparmor.security.beta.kubernetes.io/kapp-controller-sidecarexec: unconfined
```
neil-hickey, line 21: PR description talks about testing on all the IaaS, and GKE isn't checked. Did you run it there?

Could you explain what `unconfined` is ? Or if this has any implications?

cmd/controller/main.go
```diff
@@ -27,11 +29,22 @@ func main() {
...
+	flag.BoolVar(&isSidecarexecWrap, "sidecarexecwrap", false, "Run sidecarexecwrap")
 	flag.Parse()
 
-	if sidecarexec {
-		sidecarexecMain()
+	if isSidecarexec || isSidecarexecDebug {
+		sidecarexecMain(isSidecarexecDebug, flag.Args())
+		return
+	}
+	if isSidecarexecWrap {
```
neil-hickey, line 41: What's the `wrap` part for? When would I choose `sidecarexecwrap` over `sidecarexec` seen as they are exclusive

pkg/sidecarexec/sandbox_cmd_runner_darwin.go
```diff
@@ -0,0 +1,24 @@
...
+package sidecarexec
+
+import (
+	goexec "os/exec"
+
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/exec"
+)
+
+func (SandboxCmdRunner) Run(cmd *goexec.Cmd, opts exec.RunOpts) error {
+	panic("Not implemented on this OS")
```
neil-hickey, line 13: We should mention this (not supported on OSX) somewhere in the docs / helper text of the flag

neil-hickey:
- Why are we sandboxing exactly? 
- What is the least privileged sandbox for a linux namespace? 
- I see some TODOs like // TODO syscall.CLONE_NEWUTS | syscall.CLONE_NEWIPC | syscall.CLONE_NEWPID? which makes me think we don't know exactly what stuff our NS requires for min setup?
- Are we ignoring other OS's for now? Plans for the future? 
- Intention to test across all the IaaS's?

pkg/sidecarexec/sandbox_cmd_runner_linux.go
```diff
@@ -0,0 +1,282 @@
...
+			// TODO get rid of 'dev/null'? (used by vendir today)
+			{Path: "dev", NonMount: true},
+			{Path: "dev/null", NonMount: true, File: true},
+		}...)
+	}
+
+	if args.Network {
+		mounts = append(mounts, []mountPoint{
+			// Cannot mount whole 'etc' directory directly (mount returns exit status 32)
+			// Mounting individual child files/directories makes it work.
```
neil-hickey, line 164: on photon? I guess I am wondering if all the assumptions we bake in would work around any "linux / posix" like system. Let's say we swap to using a different base OS like scratch, or ubuntu or whatever.

pkg/sidecarexec/sandbox_cmd_runner_linux.go
```diff
@@ -0,0 +1,282 @@
...
+	for _, mount := range mounts {
+		if mount.File {
+			file, err := os.Create(filepath.Join(newRootDir, mount.Path))
+			if err != nil {
+				return fmt.Errorf("Creating mount file '%s': %s", mount.Path, err)
+			}
+			if len(mount.Content) > 0 {
+				_, err = file.Write([]byte(mount.Content))
+				if err != nil {
+					return fmt.Errorf("Writing into file '%s': %s", mount.Path, err)
```
neil-hickey, line 203: do we need to close the file if there's an error while writing?

pkg/sidecarexec/sandbox_cmd_runner_linux.go
```diff
@@ -0,0 +1,282 @@
...
+		srcPath := string(filepath.Separator) + mount.Path
+		dstPath := filepath.Join(newRootDir, mount.Path)
+
+		// Equivalent to: mount --bind /usr/sbin usr/sbin
+		err := syscall.Mount(srcPath, dstPath, "bind", syscall.MS_BIND, "")
+		if err != nil {
+			return fmt.Errorf("Mounting path '%s': %s", mount.Path, err)
+		}
+
+		// TODO readonly for files?
```
neil-hickey, line 232: Seems like this TODO is covered by `syscall.MS_RDONLY` right?

neil-hickey (commented):
I feel like I only scratched the surface of what this PR doing. I might need to re-read it again and again :)

cmd/controller/sidecarexec.go
```diff
@@ -13,23 +16,48 @@ import (
...
+			"vendir": true,
+			"bash":   true, // for debugging
+		},
+		RequiresNetwork: map[string]bool{
+			"vendir": true,
+			"kbld":   true,
+		},
+	})
+
+	if debug {
```
joe-kimmel-vmw, line 32: it looks like the use-case of this debug mode is "I want to see the output of a command run in the side-car" ?     can you also add that to someplace like
https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/docs/dev.md#troubleshooting-tips

and/or the docs/website repo?  (eg. https://carvel.dev/kapp-controller/docs/v0.38.0/debugging-kc/ )

pkg/sidecarexec/sandbox_cmd_runner_darwin.go
cppforlife: the only reason why i added darwin is so that code compiles successfully on a mac. no intention to run sandboxing on a mac any time soon.

pkg/sidecarexec/sandbox_cmd_runner_linux.go
```diff
@@ -0,0 +1,282 @@
...
+
+		VisiblePaths: opts.VisiblePaths,
+	}
+
+	// Roughly equivalent to: unshare -Urm -R / ...
+	cmd.Path = "/usr/bin/kapp-controller"
+	cmd.Args = []string{"kapp-controller", "--sidecarexecwrap", wrapArgs.AsString()}
+	cmd.Dir = "/"
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		// TODO syscall.CLONE_NEWUTS | syscall.CLONE_NEWIPC | syscall.CLONE_NEWPID?
```
joe-kimmel-vmw, line 48: agree with @neil-hickey that this TODO looks like it's not meant to be merged?   maybe you're planning to resolve it through testing in different iaas/infra/providers?

pkg/sidecarexec/sandbox_cmd_runner_linux.go
cppforlife: for commands that require posix env (e.g. git binary) there would have to be something available (currently it's photon). etc mounting stuff is more specific to k8s than photon. k8s sets up /etc in a particular way.

pkg/sidecarexec/sandbox_cmd_runner_linux.go
cppforlife: if we want to be good citizens (technically process dies so all fds are closed by the kernel). ill make the change.

pkg/sidecarexec/sandbox_cmd_runner_linux.go
```diff
@@ -0,0 +1,282 @@
...
+			{Path: "etc/passwd", File: true, NonMount: true, Content: "root:x:0:0:/home/kapp-controller:/usr/sbin/nologin\n"},
+			{Path: "bin"}, // e.g. bin/sh was referenced by git executables
+			{Path: "lib"},
+			{Path: "lib64"}, // some dynamic libraries
+			{Path: "var/lib"},
+			{Path: "usr/lib"},
+			{Path: "usr/libexec"}, // internal git binaries
+			{Path: "usr/share"},   // internal git data
+			{Path: "usr/sbin"},
+			// TODO get rid of 'dev/null'? (used by vendir today)
```
joe-kimmel-vmw, line 155: just wondering why this is a goal -- is the idea just that every piece of surface area here increases risk?

pkg/sidecarexec/sandbox_cmd_runner_linux.go
cppforlife: conditional below uses !mount.File. i had it working with minikube but got stuck on GKE. need to revisit why it was not happy for mounting files.

joe-kimmel-vmw, line 232: just checking on TODOs again - are you going to resolve this prior to merging

cppforlife:
> Why are we sandboxing exactly?

to further restrict what individual binaries have access to (especially since we are a multi-tenant system). in general it's good to be paranoid here and limit as much as we can regardless if we trust the code/binaries or not (never know if there is a bug or some way escalation). this also gets us closer to extending support for other methods of templating such as user brought binaries, kustomize and its plugins, etc.

> What is the least privileged sandbox for a linux namespace? I see some TODOs like // TODO syscall.CLONE_NEWUTS | syscall.CLONE_NEWIPC | syscall.CLONE_NEWPID? which makes me think we don't know exactly what stuff our NS requires for min setup?

this comment was more of an open consideration. i think im happy with the existing setup but interested to hear anyone else.

> Are we ignoring other OS's for now? Plans for the future?

files we are mounting are not os specific (typically), but generally we only work with a specific base image (photon today) that we reference in kapp-controller container image. if you are thinking of windows, i dont think there is any intention to run kapp-controller on windows nodes.

> Intention to test across all the IaaS's?

i was going to try to hit the list mentioned in the description and run ./hack/test-e2e.sh on each one.

cmd/controller/sidecarexec.go
cppforlife: yup, this would be a good addition to the docs. the intent was to allow to debug command execution in a same way that kapp-controller is executing it.

pkg/sidecarexec/sandbox_cmd_runner_linux.go
cppforlife: yup. updated PR description

config/deployment.yml
cppforlife: wdym GKE isnt checked? there 2 check boxes for GKE.

cppforlife: i did run it there already, though i need to rerun after latest changes to double check.

cmd/controller/main.go
cppforlife: sidecarexecwrap flag is used internally by https://github.com/vmware-tanzu/carvel-kapp-controller/pull/785/files#diff-8289313d09c3e91a49ad1d1a7230003f8b93e1b75dbdbe2cc1454b31e51d0956R45. 

effectively sidecarexec server creates a new sandboxed process (linux namespaces applied), and wrap command is what gets executed to continue sandboxing process (it does mounting of directories and pivot_root) before executing the actual command we care about.

may be a different name would make this a bit better? sidecarexecCmdInit?

pkg/sidecarexec/sandbox_cmd_runner_linux.go
cppforlife: imho the more minimal the better. most things dont write things to /dev/null (or even need /dev directory). the only reason why we have it for vendir is because we run `vendir sync -f ... --lock-output /dev/null` (https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/pkg/fetch/vendir.go#L388)

config/deployment.yml
neil-hickey: I guess I don't see it marked as "checked" off (just minikube) - so I wasn't sure where you were at in your testing

Screenshot 2022-07-15 at 12 43 42

DennisDenuto (commented):
How has this been tested?

What are the consequences if a sandbox command can modify the mounted /etc/resolv.conf for e.g.

If the consequences are bad, then this may warrant a specific test.

Edit: Ignore this. I left this comment to a specific code line

pkg/sidecarexec/sandbox_cmd_runner_linux.go
```diff
@@ -0,0 +1,282 @@
...
+		// Equivalent to: mount --bind /usr/sbin usr/sbin
+		err := syscall.Mount(srcPath, dstPath, "bind", syscall.MS_BIND, "")
+		if err != nil {
+			return fmt.Errorf("Mounting path '%s': %s", mount.Path, err)
+		}
+
+		// TODO readonly for files?
+		if !mount.Writeable && !mount.File {
+			// Equivalent to: mount -o remount,ro,bind etc
+			err = syscall.Mount("", dstPath, "", syscall.MS_REMOUNT|syscall.MS_BIND|syscall.MS_RDONLY, "")
```
DennisDenuto, line 235: How has this been tested?

What are the consequences if a sandbox command can modify the mounted /etc/resolv.conf for e.g.

If the consequences are bad, then this may warrant a specific test.

cppforlife: that's actually one topic i wanted feedback on. the current e2e tests of course exercise the positive scenario that the code works, however, they do not exercise that sandboxing itself is forbidding certain access. any thougths? the only idea i have is to execute tests directly against sidecarexec (which is a bit unfortunate).

neil-hickey: Yeh, I was wondering the same thing. I think we should have something, if nothing else to prevent accidental regression. I think tests against the sidecarexec are probably the most reasonable way to do it. Especially that if we one day change out the base OS, how do we know this stuff still works or not?

pkg/sidecarexec/sandbox_cmd_runner_linux.go
```diff
@@ -0,0 +1,282 @@
...
+	File      bool
+	Writeable bool
+	NonMount  bool
+	Content   string
+}
+
+// ExecuteCmd mounts various directories needed to execute sandboxed command.
+// As a last step it exec-s into given command, replacing the current process.
+func (w SandboxWrap) ExecuteCmd(rawArgs []string) error {
+	runtime.LockOSThread() // as a safety precaution for syscalls below
```
DennisDenuto, line 125: I'm still groking the architecture here so i'm making a few assumptions...

There are some syscalls made below. mounting for e.g. 

If the syscalls below block, we may reach the max os threads. Would that cause our sandbox cmd runner to essentially freeze up / not process commands?

iirc GOMAXPROCS defaults to the number of CPUs, so there may not be many OS threads by default too.

I wonder if perhaps `runtime.LockOSThread() / runtime.UnLockOSThread()` should be done only during a syscall to minimize the time spent locked...

and/or

whether some timeouts should be introduced to the syscalls made.

cppforlife: SandboxWrap class gets executed in a standalone process (called via kapp-controller --sidecarexecwrap flag) and the end of the function does an exec so nothing else should be executing within this process.

pkg/sidecarexec/sandbox_cmd_runner_linux.go
joe-kimmel-vmw: > [...] exercise that sandboxing itself is forbidding certain access. any thougths? the only idea i have is to execute tests directly against sidecarexec [...]

seems like the e2e way of doing this would be to take a known exploit and confirm that it's not exploitable, yeah?  like i'm looking at[ this one](https://apiiro.com/blog/malicious-kubernetes-helm-charts-can-be-used-to-steal-sensitive-information-from-argo-cd-deployments/) where the heart of the exploit was that they changed a web URL to a local filesystem URI,  which should be one of the things you're locking down with these namespaces, right?    So we'd kinda  white-hat hack our own tools and then show that this change locks it down (e.g. if it's a helm exploit you'd want the template step to fail,  so the e2e test would assert that failure.)
   does that make any sense?

cppforlife: imho, not the best approach, since it would not exhaustively assert on properties we expect from current implementation. i imagine we want to specifically assert on:
- only expected locations are visible
- readonly areas are indeed readonly
- when networking is not enabled, cannot communicate out

pkg/sidecarexec/sandbox_cmd_runner_linux.go
DennisDenuto: low level linux stuff here that's hurting my head because I don't know enough about it all, but *IF* someone were TODO `syscall.CLONE_NEWPID` (creating a process ns for sidecarexecwrap process) would that play nicely with the sidecarexecserver `reapZombies`? I think it will still be a child process so maybe?

depending on your answer we might wanna `s/TODO/DONTDO/` or have a bit more clarity around this comment after your testing

cppforlife: CLONE_NEWPID might actually nicely eliminate reaping of zombies since all procs (zombies or not) would be collected upon exit of the root of the pid namespace.

config/deployment.yml
cppforlife: i only checked off minikube since ive been actively developing against it. ill be checking off other things as i try to run against them. (if anybody else wants to take it for a spin on any env, just edit the description marking in progress?).

pkg/sidecarexec/sandbox_cmd_runner_linux.go
DennisDenuto: Ah ok I see! Yeah I thought I was in the long running process. Getting a better handle on the arch now thx...

pkg/sidecarexec/sandbox_cmd_runner_linux.go
DennisDenuto: ah! nice! if we can let the OS seamlessly handle cleaning up resources i'm in favor of that.

I don't think it's a blocker for this PR. But leaving the TODO's here as-is I think as time goes on, is less likely to be done.

pkg/sidecarexec/sandbox_cmd_runner_linux.go
DennisDenuto: Yeah, I don't have a clean way

Perhaps leveraging the `debug` flag and running specific bash commands within the "Sandbox"

Then you can write a test to `echo should-not-work > /etc/resolv.conf` and expect it to fail.

(Not sure if thats what you meant by "execute tests directly against sidecarexec" which in my head is referring to the server that has a whitelist of commands allowed)

cmd/controller/main.go
DennisDenuto: yeah I got a lil confused with the naming at first tbh

As of right now, in my head:
`sidecarexec` == A container that is a RPC server that allows a whitelist of commands run in an isolated way.

Therefore, `sidecarexecCmdInit` to me, sounds like the Init command for the above container.

my 2 cents:
- keep it sidecarexecwrap
- sidecarexecLiteChroot
- sidecarexecJail
- sidecarexecVirtualEnv
- sidecarexecLiteContainer

lol naming is hard... not convinced myself with the above. posting anyway.

cppforlife: another option sidecarexecPreExec

PR #795, opened 2022-07-19
Add Namespace isolation

### LINKED ISSUE #792: Different deployment mode to support good performance for 10k Apps
I need to use kapp-controller in a cluster to support deployment for 200 customers, and each customer has about 50 Apps. Totally we have more than 10000 Apps in the cluster. 

1. The queue is blocked because of so many apps
2.  More pods can't help because each pod shares the same queue, they will conflict.
3. If one customer start lots of App, other customers will wait, I hope each customer can has they own queue and quota.

Add deployment mode support for kapp-controller, there are normal, cluster and namespace deployment mode support
1) Normal as current implementation
2) Cluster Model only reconcile the resource for cluster resource, like Package and PackageRepository
3) In Namespace deployment model, each deployment can define the work namespace environment parameter in deployment, and the deployment only reconcile the resource in the give namespace, like App and 
In this way, the namespace deployment will process the CRDs in its own namespace, and cluster deployment process the cluster CRDs

There are some performance test result:
1) In Normal deployment, If there are 300 Apps is deployed, the reconcile will be very slow and new App will be triggered after more than 5m.  The deployment is to deploy to another Kubernetes cluster.
2) I did some change for CRD App to support namespace (https://github.com/showpune/carvel-kapp-controller/tree/cloud-version), we still can get good performance even there are 5000 apps in 100 namespaces
Current situation
image
Namespace deployment after change
image

There is similar issue for kpack and cartographer.

---
Vote on this request

showpune:
Add two env RECONCILE_SCOPE and RECONCILE_NAMESPACE to control the reconcile scope
1) Add env RECONCILE_SCOPE to specify if reconcile only for cluster resource ( cluster ), namespace resource (namespace) or both (not set RECONCILE_SCOPE )
2) Add env RECONCILE_NAMESPACE to specify the reconcile namespace 

Fixes #792

##### Review Checklist:

- [X] Relevant docs in this repo added or updated
  a link to that PR
  change

```docs

```


cmd/controller/run.go
```diff
@@ -34,8 +34,10 @@ import (
 )
 
 const (
-	PprofListenAddr       = "0.0.0.0:6060"
-	kappctrlAPIPORTEnvKey = "KAPPCTRL_API_PORT"
+	PprofListenAddr             = "0.0.0.0:6060"
+	kappctrlAPIPORTEnvKey       = "KAPPCTRL_API_PORT"
+	kappctrlWorkNamespaceEnvKey = "WORK_NAMESPCE"
+	cluster                     = "cluster"
```
neil-hickey: In the case where a namespaces is called "cluster" this won't work. This probably ought to be a separate flag instead of a "magic string" here.

In general we prefer flags over environment variables, this is true for the namespace stuff below also

cmd/controller/run.go
```diff
@@ -59,6 +61,22 @@ func Run(opts Options, runLog logr.Logger) error {
 		restConfig.Timeout = opts.APIRequestTimeout
 	}
 
+	var workForClusterResource bool = true
```
neil-hickey: workForClusterResource := true

cmd/controller/run.go
```diff
@@ -59,6 +61,22 @@ func Run(opts Options, runLog logr.Logger) error {
...
+	var workForNamespaceResource bool = true
+	if namespaceWorkedFor, ok := os.LookupEnv(kappctrlWorkNamespaceEnvKey); ok {
+		opts.Namespace = namespaceWorkedFor
+		if namespaceWorkedFor == cluster {
+			workForClusterResource = true
```
neil-hickey: The pattern of `workFor`... doesn't seem to fit quite right. Maybe some other options might be:

- ReconcileScope...
- Shard...
- ReconcileBy...
- FilterBy...

Any other opinions / options?

As mentioned above these names don't make things clear that you must have at least one `workForClusterResource` for the system to work (i.e the APIServer to be up and running to provide the CRDs to the cluster)

For example, the word singleton implies that at least one "cluster" thing is required

cmd/controller/run.go
```diff
@@ -59,6 +61,22 @@ func Run(opts Options, runLog logr.Logger) error {
...
+			workForNamespaceResource = false
+		} else if namespaceWorkedFor != "" && namespaceWorkedFor != "None" {
```
neil-hickey: This shouldn't be necessary: `namespaceWorkedFor != "None"`

This should be covered by having a flag instead.

cmd/controller/run.go
```diff
@@ -214,7 +235,7 @@ func Run(opts Options, runLog logr.Logger) error {
 		}
 	}
 
-	{ // add controller for pkgrepositories
+	if workForClusterResource { // add controller for pkgrepositories
```
neil-hickey: Is there a reason why PackageRepostories are not also handled like the other namespaced reconcilers?

config-namespace/0-namespace.yml
```diff
@@ -0,0 +1,12 @@
+#@ load("@ytt:data", "data")
+
+#! Prefix filename with "0-" so that generated config
+#! has Namespace first so that kubectl can can install kc
+#! (kapp of course perm automatic ordering)
+
+#@ if/end data.values.namespace != "default" and data.values.create_namespace:
```
neil-hickey: Why is this file changed as part of this PR?

config-namespace/build.yml
```diff
@@ -0,0 +1,23 @@
+#@ load("@ytt:data", "data")
```
neil-hickey: Did you mean to commit this as a new file instead of changing the `config` folder?

config-namespace/deployment.yml
```diff
@@ -0,0 +1,108 @@
...
+      - name: kapp-controller
+        image: kapp-controller
+        args:
+        - #@ "-enable-api-priority-and-fairness={}".format(data.values.enable_api_priority_and_fairness)
+        env:
+        - name: KAPPCTRL_MEM_TMP_DIR
+          value: /etc/kappctrl-mem-tmp
+        - name: KAPPCTRL_SIDECAREXEC_SOCK
+          value: /etc/kappctrl-mem-tmp/sidecarexec.sock
+        - name: WORK_NAMESPCE
```
neil-hickey: It looks like you just added this WORK_NAMESPACE environment variable, but the whole file is new. Did you mean to change the `config/deployment.yml` file instead?

config-namespace/values.yml
```diff
@@ -0,0 +1,17 @@
+#@data/values
```
neil-hickey: Did you mean to commit this as a new file instead of changing the `config` folder?

config-namespace/rbac.yml
```diff
@@ -0,0 +1,22 @@
+#@ load("@ytt:data", "data")
```
neil-hickey: Did you mean to commit this as a new file instead of changing the `config` folder?

hack/deploy-namespace.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
neil-hickey: This isn't needed for the PR.

config/rbac.yml
```diff
@@ -67,6 +67,48 @@ rules:
 - apiGroups: ["flowcontrol.apiserver.k8s.io"]
   resources: ["prioritylevelconfigurations", "flowschemas"]
   verbs: ["list", "watch"]
+
+---
+kind: ClusterRole
```
neil-hickey, line 72: This will cause a conflict when deploying multiple kapp-controllers as it is currently a Cluster-wide role.

This should be a `Role` with relevant `RoleBindings` per namespace.

neil-hickey (dismissed):

cmd/controller/run.go
showpune: When I left empty in values.yaml, the value is "None" and I tried to keep compatible with it.

I can remove it

cmd/controller/run.go
showpune: I handle the PackageResository as a cluster resource (if workForClusterResource), what's your concern?

cmd/controller/run.go
showpune: Updated the key name

cmd/controller/run.go
showpune: Updated

cmd/controller/run.go
showpune: Use RECONCILE_NAMESPCE and RECONCILE_SCOPE to avoid magic string

config-namespace/0-namespace.yml
showpune: I use the config-namespace for namespaced deployment, I will remove it

config-namespace/build.yml
showpune: I will update it in config folder

config-namespace/deployment.yml
showpune: OK

config-namespace/rbac.yml
showpune: I agree. I planned to put all namespaced deployment in a seperate folder

config-namespace/values.yml
showpune: OK and I will remove the config-namespace folder

hack/deploy-namespace.sh
showpune: I will remove it

config/rbac.yml
showpune: I think it RBAC to monitor to cluster resoource is needed. The clustorRole can be kept and we need to creat clusterrolebinding for the SA kapp-controller-sa in each namespace

cmd/controller/run.go
joe-kimmel-vmw: @showpune is it ok to put the 'A' back into `RECONCILE_NAMESPACE`(in 'space') ?   This looks like it was done consistently throughout so just checking whether there's a strong preference or it was a typo that got copy/pasted. thanks!

config/values.yml
```diff
@@ -9,10 +9,12 @@ enable_api_priority_and_fairness: true
 
 dangerous_enable_pprof: false
 
-push_images: false
+push_images: true
```
joe-kimmel-vmw: in the final submission please revert this `push_images` back to `false`

joe-kimmel-vmw:
Another wrinkle with this feature as implemented is that as soon as I have made a kapp-controller that has namespace scope, now (a) it will introduce errors and race conditions if i start a kapp-controller also in "cluster" scope, and (b) it will require me to add one kapp-controller per namespace for any/every new namespace on this cluster. 

One thought I have is that we could do some regular expression matching (or regex-lite, ie just "*" wildcard) but that introduces other complexities as well.

config/build.yml
```diff
@@ -1,24 +1,24 @@
 #@ load("@ytt:data", "data")
 
 apiVersion: kbld.k14s.io/v1alpha1
-kind: Sources
+kind: Config
```
neil-hickey, line 4: Is this file should be reverted back to the original for it be mergable / tests to run

cmd/controller/run.go
```diff
@@ -34,8 +34,12 @@ import (
...
 
 const (
-	PprofListenAddr       = "0.0.0.0:6060"
-	kappctrlAPIPORTEnvKey = "KAPPCTRL_API_PORT"
+	PprofListenAddr                  = "0.0.0.0:6060"
+	kappctrlAPIPORTEnvKey            = "KAPPCTRL_API_PORT"
+	kappctrlReconcileNamespaceEnvKey = "RECONCILE_NAMESPCE"
+	kappctrlReconcileScopeEnvKey     = "RECONCILE_SCOPE"
+	cluster                          = "cluster"
+	namespace                        = "namespace"
```
joe-kimmel-vmw: per the comment I left earlier here is a slightly more fleshed out thought:
Here, rather than the "ReconcileScope" flag, we could introduce a boolean flag for "should i run the aggregate api server?" (yes/no) and an orthogonal environment variable for "what is the scope of the namespaces i'm reconciling?"    this second env var could use empty string ("") to indicate "all namespaces" or a legal namespace name and do an exact string match.  that way we could avoid regular expressions. :) 

so then the env vars would become something more like 
```
KAPPCTRL_START_API_SERVER
KAPPCTRL_RECONCILE_NAMESPACE
```

further, as we do prefer flags, and it turns out there's already a namespace flag: 
https://github.com/vmware-tanzu/carvel-kapp-controller/blob/8dba0472c9ebdca83d3dd981f783ff9afa75ac2b/cmd/controller/main.go#L24

then it's only necessary to add a boolean flag indicating whether or not to start the aggregate api server. 

Note that in this arrangement we would also be namespacing the PKGR resource - this was called out in a different comment but maybe now it's more clear why that's preferred, since we're slicing the functionality slightly differently.

showpune: If we use KAPPCTRL_START_API_SERVER and KAPPCTRL_RECONCILE_NAMESPACE, it means we need to add another constant for namespace, see the sheet below:
image
Is it right?

For my point of view, reconcile scope and start api server just have indirect relationship. Is reconcile scope more straightforward to user?

joe-kimmel-vmw: Hi @showpune - I don't think the above diagram makes sense, no.  I don't think we need that middle row, because I think all the reconcilers should be scoped. By scoping all the reconcilers together: app, pkgi, pkgr, pkg; all of those either are directed towards one namespace or all namespaces,  then the only question is "should this process also start the apiserver?"   

We definitely want to avoid magic strings, as for example in your example above a namespace called "None" would create problems.

If you're concerned that the process running the aggregate api server shouldn't also be reconciling anything at all, one option could be to point it at a namespace where we don't expect any resources to be created (such as the `kapp-controller` namespace of kapp-controller's deployment). 

Another option is to use regular expressions to specify which namespaces to match. This would introduce regular expressions with all the problems they bring, but it could also allow for an administrator to avoid needing to add an additional kapp-controller for every single namespace. (for example, maybe you'd want a single kapp-controller to reconcile all namespaces matching `dev-.*` to simplify developer onboarding?) 

I agree that reconcile scope and whether or not start the API server are unrelated. I think that's a point in favor of slicing the parameters this way, as it avoids having dependencies across flags.  As far as names of the flags, I think you would have better "user perspective" than I do, but we can also ask in our slack channel or community meeting.

does this make more sense, what I'm trying to propose?  What do you think?

showpune: Hi @joe-kimmel-vmw，

It is OK that we use KAPPCTRL_START_API_SERVER and KAPPCTRL_RECONCILE_NAMESPACE. 

I want to confirm with you that if KAPPCTRL_START_API_SERVER is set to false, should we start reconciler for cluster resource like pkgr and pkg? The deployments are like below:
image
There should be a controller in each namespace, if we start reconciler in each namespace, if anything changed in cluster resource, it will be reconciled by all namespace with reconciler and there should be some data conflict issue

So the solution should be:
KAPPCTRL_START_API_SERVER=true: start the api server, reconcile the cluster resource like pkg and pkgr
KAPPCTRL_START_API_SERVER=false: not start the api server, not reconcile the cluster resource like pkg and pkgr

Is it right?

joe-kimmel-vmw: @showpune I think there's still a misunderstanding. The way i see it, _all_ resources that are reconciled are namespaced. I'm not sure that we need the notion of a "cluster resource".  Am I missing something?

showpune: Hi @joe-kimmel-vmw ,

I understood your idea and I commit the change, is it what you expect?

cmd/controller/run.go
```diff
@@ -34,8 +34,10 @@ import (
 )
 
 const (
-	PprofListenAddr       = "0.0.0.0:6060"
-	kappctrlAPIPORTEnvKey = "KAPPCTRL_API_PORT"
+	PprofListenAddr                  = "0.0.0.0:6060"
+	kappctrlAPIPORTEnvKey            = "KAPPCTRL_API_PORT"
```
joe-kimmel-vmw: I think we still prefer these to be flags rather than env vars, see e.g. :
https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/cmd/controller/main.go#L24

config/build.yml
```diff
@@ -18,7 +18,7 @@ sources:
 #@ if/end data.values.push_images:
 ---
 apiVersion: kbld.k14s.io/v1alpha1
-kind: ImageDestinations
+kind: Config
```
joe-kimmel-vmw, line 21: do we want this change?

cmd/controller/run.go
```diff
@@ -59,6 +61,23 @@ func Run(opts Options, runLog logr.Logger) error {
...
+		if kappctrlStartApiServerString == "false" {
+			kappctrlStartApiServer = false
+		}
+		runLog.Info("Start with api server", "kappctrlStartApiServer", kappctrlStartApiServer)
+	} else {
+		runLog.Info("Start with api server with default value", "kappctrlStartApiServer", kappctrlStartApiServer)
+	}
+
+	if namespaceWorkedFor, ok := os.LookupEnv(kappctrlReconcileNamespaceEnvKey); ok {
+		opts.Namespace = namespaceWorkedFor
```
joe-kimmel-vmw: i'm still curious whether you think there's a use-case for reconciling for multiple namespaces at all?   Do you think it would always be desirable to enumerate every single namespace and basically require a new instance of kapp-controller each time a namespace is added?  Although maybe the way you've implemented is great from a multi-tenancy isolation perspective.

joe-kimmel-vmw (commented):
I suspect we also don't want you to delete [config-release/values.yml](https://github.com/vmware-tanzu/carvel-kapp-controller/pull/795/files#diff-09d86c4f70732b46a0ce953b97120419794501b14bb685325c3fe9766e52ae2d)

config/build.yml
showpune: I used this schema suggested by @neil-hickey 
image

and I checked the doc from https://carvel.dev/kbld/docs/v0.34.0/config/, is it the latest schema?

cmd/controller/run.go
showpune: For now, I want to one controller in one namespace for one customer instance, I can define the total resource quota for a customer in one namespace.  I prefer to use a new PR for multiple tenant support

joe-kimmel-vmw: fair!

joe-kimmel-vmw:
thanks @showpune -  Please also fix any remaining linter complaints and restore the file [config-release/values.yml](https://github.com/vmware-tanzu/carvel-kapp-controller/pull/795/files#diff-09d86c4f70732b46a0ce953b97120419794501b14bb685325c3fe9766e52ae2d). 

Then I also asked @cppforlife to take a look since he's the official approver for this project.

cmd/controller/main.go
```diff
@@ -28,6 +28,7 @@ func main() {
 	flag.DurationVar(&ctrlOpts.APIRequestTimeout, "api-request-timeout", time.Duration(0), "HTTP timeout for Kubernetes API requests")
 	flag.BoolVar(&ctrlOpts.APIPriorityAndFairness, "enable-api-priority-and-fairness", true, "Enable/disable APIPriorityAndFairness feature gate for apiserver. Recommended to disable for <= k8s 1.19.")
 	flag.BoolVar(&sidecarexec, "sidecarexec", false, "Run sidecarexec")
+	flag.BoolVar(&ctrlOpts.StartAPIServer, "startAPIServer", true, "start apiserver")
```
cppforlife: nit: startAPIServer => start-api-server

config/build.yml
cppforlife: i think it's fine to adopt `Config`

cppforlife (commented):
im in favor of merging this PR once the flag casing it corrected.

cmd/controller/main.go
showpune: Done

showpune:
create new PR https://github.com/vmware-tanzu/carvel-kapp-controller/pull/826

PR #798, opened 2022-07-26
Packages can constrain k8s and kc versions

### LINKED ISSUE #484: Support k8s version compatibility
We would like a way to check if upgrading a kubernetes cluster would render some of the installed packages incompatible. When building tooling on top of Carvel, it would also be nice to prevent users from attempting to install the Carvel packages on incompatible kubernetes clusters.

Standardize a format for package authors to specify k8s version compatibility range in the PackageMetadata, for tooling on top of it to parse.  

---
Vote on this request

### LINKED ISSUE #577: package should be able to specify acceptable kapp-controller and k8s version required

package authors want ability to specify that particular package can only be installed on kapp-controllers version X and above. example case: im using particular new ytt feature that it's only available in kc v0.45.0+ and want to prevent somebody from trying to install it on kapp-controller v0.43.0 because it will fail or do something terrible.

we do have some prior art in other carvel tools (vendir, kbld, kapp, ytt) e.g. https://github.com/vmware-tanzu/carvel-vendir/blob/develop/pkg/vendir/config/config.go#L25

questions/related areas:
- should we introduce max version as well?
- is this a App CR feature as well?
- similar topic: min/max kubernetes version for a package

---
Vote on this request

joe-kimmel-vmw:
- kubernetes and kapp-controller version constraints can be overridden by annotations
- range version selection of a Package chooses the highest version of
  that package which also satisfies kc and k8s version constraints
- Error messages if no package satisfies constraints provides detail of
  which constraints failed.
- if PKGI specifies a different cluster, that cluster's version of k8s
  is checked for the constraints (manually verified - no automated test
  of this)

Fixes #484 
Fixes #577 

```release-note
Packages can constrain kubernetes and kapp-controller versions
```

note that we def. punted on the multicluster case test.  We did verify this manually several times throughout development and again after our last change.  However, open to discussion re. using some shell scripts or introducing an e2e dependency on CAPI for spinning up a second cluster and making that cluster's kubeconfig available to kc via a Secret.

##### Review Checklist:

  a link to that PR
  change

design proposal doc: https://hackmd.io/6FQhzRIPR_uceqtTqSIXaA

carvel.dev docs pending

```docs

```


joe-kimmel-vmw:
note that @neil-hickey coauthored this PR with me.

config/crds.yml
```diff
@@ -108,6 +108,32 @@ spec:
...
+                properties:
+                  constraints:
+                    type: string
+                  prereleases:
+                    properties:
+                      identifiers:
+                        items:
+                          type: string
+                        type: array
+                    type: object
```
cppforlife: lets collect various versions of k8s distros so that we know what is the versions 'data set' we are working with. on gke all versions are prereleases and so it would be somewhat awkward to ask all users to set `prereleases: {}`

hack/dev-deploy.sh
```diff
@@ -2,9 +2,13 @@
 
 set -ex
 
+source $(dirname "$0")/version-util.sh
+
+#ytt -f config/ -f config-release -v kapp_controller_version=" --data-values-env=KCTRL | kbld -f- > ./tmp/release.yml
```
cppforlife: remove

pkg/packageinstall/packageinstall.go
```diff
@@ -201,6 +205,98 @@ func (pi *PackageInstallCR) reconcileAppWithPackage(existingApp *kcv1alpha1.App,
 	return reconcile.Result{}, nil
 }
 
+func (pi *PackageInstallCR) getClusterVersion() (*version.Info, error) {
```
cppforlife: let's move this back into deploy factory (as a new method: KubernetesEnv?) so that private details of it are not exposed to PackageInstall.

pkg/packageinstall/packageinstall.go
```diff
@@ -201,6 +205,98 @@ func (pi *PackageInstallCR) reconcileAppWithPackage(existingApp *kcv1alpha1.App,
...
+		return nil, err
+	}
+	return vi, nil
+}
+
+func (pi *PackageInstallCR) clusterVersionConstraints(pkg *datapkgingv1alpha1.Package) error {
+	const kubernetesVersionOverrideAnnotation = "packaging.carvel.dev/ignore-kubernetes-version-selection"
+	_, found := pi.model.Annotations[kubernetesVersionOverrideAnnotation]
+	if found {
+		pi.log.Info("Found k8s Version override annotation; not applying version constraints")
```
cppforlife: nit: Version => version

pkg/packageinstall/packageinstall.go
```diff
@@ -201,6 +205,98 @@ func (pi *PackageInstallCR) reconcileAppWithPackage(existingApp *kcv1alpha1.App,
...
+
+		semverVersions := versions.NewRelaxedSemversNoErr([]string{vi.GitVersion})
+		matchedVers, err := semverVersions.FilterConstraints(pkg.Spec.KubernetesVersionSelection.Constraints)
+		if err != nil {
+			return err
+		}
+
+		if matchedVers.Len() == 0 {
+			return fmt.Errorf("Cluster is running kubernetes %s but package constrained versions to: %s",
+				vi.GitVersion, pkg.Spec.KubernetesVersionSelection.Constraints)
```
cppforlife: should probably use Description() -- https://github.com/vmware-tanzu/carvel-vendir/blob/develop/pkg/vendir/versions/v1alpha1/config.go#L39

pkg/packageinstall/packageinstall.go
```diff
@@ -241,7 +346,10 @@ func (pi *PackageInstallCR) referencedPkgVersion() (datapkgingv1alpha1.Package,
 
 	selectedVersion, err := versions.HighestConstrainedVersion(versionStrs, verConfig)
 	if err != nil {
-		return datapkgingv1alpha1.Package{}, err
+		retErr := fmt.Errorf(
+			"Pre-Filter: Number of Packages excluded due to k8s or kc constraints: %d. (details: %v)\nPost-Exclusion Package Version Selection: %v",
+			len(rejectPkgReasons), rejectPkgReasons, err.Error())
+		return datapkgingv1alpha1.Package{}, retErr
 	}
```
cppforlife: ideally error would be something like this: `Expected to find at least one version, but did not (details: all=0 -> after-prereleases-filter=0 -> after-constraints-filter=0 -> after-k8s-compat-filter=0 -> after-kapp-controller-compat-filter=0)`

pkg/packageinstall/reconciler.go
```diff
@@ -22,6 +22,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+// ControllerVersion holds the semver string for the version of kapp-controller that is running, so that we can compare it against version constraints in packages
+var ControllerVersion string
```
cppforlife: lets get rid of a global variable in favor of injecting data into necessary class.

test/e2e/kappcontroller/packageinstall_test.go
```diff
@@ -172,6 +173,234 @@ stringData:
...
+metadata:
+  name: pkg-demo-values
+stringData:
+  values.yml: |
+    hello_msg: "hi"
+`
+	return installPkgYAML
+}
+
+func Test_PackageInstalled_FromPackageInstall_VersionConstraints(t *testing.T) {
```
cppforlife, line 262: given the scope of the selection logic, it makes sense to push most of the variations into unit tests (PackageInstallCR class) and probably just keep one high level e2e test.

pkg/packageinstall/packageinstall.go
joe-kimmel-vmw: fwiw we put it into deploy module but it still feels awkward to claim that it's part of a "factory" imo, - I already apologized to Neil for not just doing exactly what you said and causing more discussion, so - sorry, but i see your point re. scope of concern, but not re. "factory" specifically.

pkg/packageinstall/packageinstall.go
joe-kimmel-vmw: pending https://github.com/vmware-tanzu/carvel-vendir/pull/175

config/crds.yml
joe-kimmel-vmw: azure looks like they just use major.minor.patch: https://docs.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli

joe-kimmel-vmw: EKS also looks like it uses major.minor.patch: https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html

go.mod
```diff
@@ -4,13 +4,12 @@ go 1.18
...
 require (
 	github.com/fatih/color v1.12.0 // indirect
-	github.com/ghodss/yaml v1.0.0 // indirect
 	github.com/gogo/protobuf v1.3.2
 	github.com/google/go-cmp v0.5.6 // indirect
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0 // indirect
 	github.com/prometheus/client_golang v1.11.1
 	github.com/stretchr/testify v1.7.5
-	github.com/vmware-tanzu/carvel-vendir v0.24.0
+	github.com/vmware-tanzu/carvel-vendir v0.29.1-0.20220727190601-9dd41a08ed60
```
neil-hickey: This fairly large fast forward causes us to pull in more vendored things, hence why there are so many file changes...

config/crds.yml
neil-hickey: - TKGm : (don't seem to use a pre-release, but do include metadata about the build version)
```
NAME   STATUS   ROLES    AGE   VERSION
x      Ready    <none>    8d   v1.22.5+vmware.1
```
- AKS (azure): 
```
NAME   STATUS   ROLES    AGE   VERSION
x      Ready    <none>    8d   v1.23.8
```
- GKE (google):
```
NAME   STATUS   ROLES    AGE   VERSION
x      Ready    <none>    8d   v1.21.12-gke.2200
```
- Minikube:
```
NAME   STATUS   ROLES    AGE   VERSION
x      Ready    <none>    8d   v1.24.1
```
- EKS (amazon):
```
NAME  STATUS   ROLES    AGE   VERSION
x     Ready    <none>   10m   v1.21.12-eks-5308cf7
```
- OpenShift on vSphere:
- `v1.23.3+e419edf`

pkg/packageinstall/packageinstall_version_handler.go
```diff
@@ -105,6 +105,7 @@ func (ipvh *PackageInstallVersionHandler) isEligibleForVersionUpgrade(version st
 
 	semverConfig := installedPkg.Spec.PackageRef.VersionSelection
 
+	// TODO: we should  make this apply the same kc and k8s constraints as the other one to avoid unnecessary reconciles
```
neil-hickey: So this watcher doesn't get the PKGI resource,  only the PKG. This means:

- We can't check for the absense / presense of the ignore annotations
- We also have to duplicate / make public the constraint checking functions
- It introduces more complexity into this version handler and another API call / cross-reference check

Will probably want to chat this one through as a group

pkg/packageinstall/packageinstall.go
```diff
@@ -216,6 +289,9 @@ func (pi *PackageInstallCR) referencedPkgVersion() (datapkgingv1alpha1.Package,
 	if err != nil {
 		return datapkgingv1alpha1.Package{}, err
 	}
+	if len(pkgList.Items) == 0 {
```
joe-kimmel-vmw, line 276: we could add a test for this case.

pkg/deploy/service_accounts.go
```diff
@@ -52,6 +55,66 @@ func (s *ServiceAccounts) Find(genericOpts GenericOpts, saName string) (Processe
 	return pgoForSA, nil
 }
 
+// GetClusterVersion returns the kubernetes API version for the cluster which has been supplied to kapp-controller via a kubeconfig
```
cppforlife: lets move this code out to dedicated file cluster_version.go

pkg/deploy/service_accounts.go
```diff
@@ -52,6 +55,66 @@ func (s *ServiceAccounts) Find(genericOpts GenericOpts, saName string) (Processe
...
+	if len(saName) > 0 {
+		return cc.Discovery().ServerVersion()
+	}
+
+	processedGenericOpts, err := processOpts(
+		saName,
+		specCluster,
+		GenericOpts{Name: objMeta.Name, Namespace: objMeta.Namespace},
+		NewServiceAccounts(cc, log),
+		NewKubeconfigSecrets(cc))
```
cppforlife: service accounts object should be shared with the one in the factory to take advantage of token cache.

pkg/packageinstall/packageinstall.go
```diff
@@ -26,6 +27,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
```
cppforlife: nit: remove newline

pkg/packageinstall/packageinstall.go
```diff
@@ -201,6 +205,74 @@ func (pi *PackageInstallCR) reconcileAppWithPackage(existingApp *kcv1alpha1.App,
...
+		pi.log.Info("Found kubernetes version override annotation; not applying version constraints")
+		return true
+	}
+
+	vi, err := deploy.GetClusterVersion(pi.coreClient, pi.model.Spec.ServiceAccountName, pi.model.Spec.Cluster, &pi.model.ObjectMeta, pi.log)
+	if err != nil {
+		pi.log.Error(err, "Unable to retrieve cluster kubernetes version")
+		return false
+	}
+	v, err := dropPreAndBuild(vi.GitVersion)
```
cppforlife: dropPreAndBuild => dropVersionPrereleaseAndBuild might be a clearer name

pkg/packageinstall/packageinstall.go
```diff
@@ -201,6 +205,74 @@ func (pi *PackageInstallCR) reconcileAppWithPackage(existingApp *kcv1alpha1.App,
...
+	if err != nil {
+		pi.log.Error(err, "Unable to parse cluster kubernetes version", "version string", vi.GitVersion)
+		return false
+	}
+	constraints, err := semver.ParseRange(pkg.Spec.KubernetesVersionSelection.Constraints)
```
cppforlife: lets add a validation for constraints on a package: https://github.com/vmware-tanzu/carvel-kapp-controller/blob/e5667d3102f34e05df58e42670689b3d6ab5ee29/pkg/apiserver/apis/datapackaging/validation/validations.go#L37. with that addition, it's fine that we would ignore the error here (and log).

pkg/packageinstall/packageinstall.go
```diff
@@ -201,6 +205,74 @@ func (pi *PackageInstallCR) reconcileAppWithPackage(existingApp *kcv1alpha1.App,
...
+	}
+	const kubernetesVersionOverrideAnnotation = "packaging.carvel.dev/ignore-kubernetes-version-selection"
+
+	_, found := pi.model.Annotations[kubernetesVersionOverrideAnnotation]
+	if found {
+		pi.log.Info("Found kubernetes version override annotation; not applying version constraints")
+		return true
+	}
+
+	vi, err := deploy.GetClusterVersion(pi.coreClient, pi.model.Spec.ServiceAccountName, pi.model.Spec.Cluster, &pi.model.ObjectMeta, pi.log)
```
cppforlife: clusterVersionConstraintsSatisfied gets called per every package, so we should move out GetClusterVersion outside to avoid N calls.

test/e2e/kappcontroller/packageinstall_test.go
```diff
@@ -172,6 +173,165 @@ stringData:
...
+		installPkgYAML := generatePkgiYAML(name, "", "<1.0.0", []string{}) + sas.ForNamespaceYAML()
+		out, err := kapp.RunWithOpts([]string{"deploy", "-a", name, "-f", "-"},
+			e2e.RunOpts{
+				StdinReader: strings.NewReader(installPkgYAML),
+				AllowError:  true})
+		assert.Contains(t, out, "after-kubernetes-version-check=0")
+		require.Error(t, err)
+		// Without calling cleanup after each stage, somehow it seems like the controller doesn't
+		// get the updated annotation.
+		cleanUp()
```
cppforlife, line 290: lets try to figure out since sounds like real users will run into this situation => update to annotation should "fix" the PackageInstall

cppforlife: => leave comment about metadata.generation not being incr when annotations change

joe-kimmel-vmw: @cppforlife can we go over this one more time - still a little unclear to us exactly what is breaking and why it is/isn't a user-facing concern?

cmd/controller/run.go
```diff
@@ -165,6 +166,9 @@ func Run(opts Options, runLog logr.Logger) error {
 	refTracker := reftracker.NewAppRefTracker()
 	updateStatusTracker := reftracker.NewAppUpdateStatus()
 
+	// initialize deploy factory once - the deploy factory contains a cache which should be only setup once.
```
joe-kimmel-vmw: maybe worth saying "cache of _"  (service accounts? right?) to assist future readers in their journey

neil-hickey: Yeh, we can make it more specific.

pkg/packageinstall/packageinstall.go
```diff
@@ -239,7 +295,31 @@ func (pi *PackageInstallCR) referencedPkgVersion() (datapkgingv1alpha1.Package,
 
 	verConfig := verv1alpha1.VersionSelection{Semver: semverConfig}
 
-	selectedVersion, err := versions.HighestConstrainedVersion(versionStrs, verConfig)
+	kcConstraint := func(pkgVer string) bool {
+		pkg := versionToPkg[pkgVer]
+		return pi.kcVersionConstraintsSatisfied(&pkg)
+	}
+
+	vi, err := pi.deployFactory.GetClusterVersion(pi.model.Spec.ServiceAccountName, pi.model.Spec.Cluster, &pi.model.ObjectMeta, pi.log)
```
neil-hickey: Do we make the tradeoff of a two-pass filter vs a call to the cluster for k8s info on each reconcile....

By two pass I mean: first pass --> check if any packageinstall has constraints (if not, skip over this code), second pass as is currently.

pkg/app/app_factory.go
```diff
@@ -29,13 +29,12 @@ type CRDAppFactory struct {
 }
 
 // NewCRDApp creates a CRDApp injecting necessary dependencies.
-func (f *CRDAppFactory) NewCRDApp(app *kcv1alpha1.App, log logr.Logger) *CRDApp {
+func (f *CRDAppFactory) NewCRDApp(app *kcv1alpha1.App, log logr.Logger, deployFactory deploy.Factory) *CRDApp {
```
cppforlife: instead of passing down deployFactory into NewCRDApp, let's make CRDAppFactory constructor accept it. this way whoever calls NewCRDApp does not need to know anything about deployFactory.

pkg/app/reconciler.go
```diff
@@ -27,14 +28,21 @@ type Reconciler struct {
 	appClient       kcclient.Interface
 	log             logr.Logger
 	crdAppFactory   CRDAppFactory
+	deployFactory   deploy.Factory
 	appRefTracker   *reftracker.AppRefTracker
 	appUpdateStatus *reftracker.AppUpdateStatus
 }
 
 // NewReconciler constructs new Reconciler.
-func NewReconciler(appClient kcclient.Interface, log logr.Logger, crdAppFactory CRDAppFactory,
```
cppforlife, line 35: with change to CRDAppFactory accepting deployFactory, this class would not have to change (since deployFactory is an implementation detail of how CRDAppFactory works).

pkg/deploy/factory.go
```diff
@@ -38,32 +41,42 @@ func NewFactory(coreClient kubernetes.Interface,
...
+		return nil, err
+	}
+	const suffix string = ".app"
+	return NewKapp(suffix, opts, *processedGenericOpts,
+		f.globalKappDeployRawOpts(), cancelCh, f.cmdRunner), nil
+}
+
+// processOpts takes generic opts and a ServiceAccount Name, and returns a populated kubeconfig that can connect to a cluster.
+// if the saName is empty then you'll connect to a cluster via the clusterOpts inside the genericOpts, otherwise you'll use the specified SA.
+func processOpts(saName string, clusterOpts *v1alpha1.AppCluster, genericOpts GenericOpts, serviceAccounts *ServiceAccounts, kubeconfigSecrets *KubeconfigSecrets) (*ProcessedGenericOpts, error) {
```
cppforlife: since we have placed GetClusterVersion onto Factory, we should be able to reduce processOpts scope as well but placing it onto Factory. by doing that it will also not have to accept serviceAccounts and kubeconfigSecrets and just read off f.

pkg/packageinstall/packageinstall.go
cppforlife: or it can also be memoized so that we get the best of both worlds.

pkg/app/app_factory.go
neil-hickey: By constructor do you mean just add it as another public field on the struct? There is no constructor right now for CRDAppFactory. We could make one, and then keep it as a private field?

neil-hickey: @cppforlife

pkg/packageinstall/packageinstall.go
neil-hickey: Maybe it can be in the case of the on-cluster version. But should we really "cache" the version for an external cluster? Wouldn't that be prone to stale-ness?

pkg/deploy/factory.go
```diff
@@ -38,12 +41,24 @@ func NewFactory(coreClient kubernetes.Interface,
...
+		return nil, err
+	}
+	const suffix string = ".app"
+	return NewKapp(suffix, opts, *processedGenericOpts,
+		f.globalKappDeployRawOpts(), cancelCh, f.cmdRunner), nil
+}
+
+// processOpts takes generic opts and a ServiceAccount Name, and returns a populated kubeconfig that can connect to a cluster.
+// if the saName is empty then you'll connect to a cluster via the clusterOpts inside the genericOpts, otherwise you'll use the specified SA.
+func (f Factory) processOpts(saName string, clusterOpts *v1alpha1.AppCluster, genericOpts GenericOpts) (*ProcessedGenericOpts, error) {
```
cppforlife: nit: doesnt have to be pointer afaik

pkg/deploy/factory.go
```diff
@@ -92,3 +105,33 @@ func (f Factory) globalKappDeployRawOpts() []string {
 	}
 	return nil
 }
+
+// GetClusterVersion returns the kubernetes API version for the cluster which has been supplied to kapp-controller via a kubeconfig
+func (f Factory) GetClusterVersion(saName string, specCluster *v1alpha1.AppCluster, objMeta *metav1.ObjectMeta, log logr.Logger) (*version.Info, error) {
+	// this logic is duplicated here and in ProcessOpts: if the serviceAccount name is present, we will be deploying to this cluster
+	if len(saName) > 0 {
```
cppforlife: we should use a switch statement as above to catch the case when it's neither saName or kubeconfig

pkg/packageinstall/packageinstall.go
```diff
@@ -216,15 +273,22 @@ func (pi *PackageInstallCR) referencedPkgVersion() (datapkgingv1alpha1.Package,
...
 
+	requiresClusterVersion := false
 	for _, pkg := range pkgList.Items {
 		if pkg.Spec.RefName == pi.model.Spec.PackageRef.RefName {
 			versionStrs = append(versionStrs, pkg.Spec.Version)
 			versionToPkg[pkg.Spec.Version] = pkg
 		}
+		if pkgHasK8sConstraint(&pkg) {
+			requiresClusterVersion = true
+		}
```
cppforlife: this if should be within above if since we only care about a subset of packages.

pkg/packageinstall/packageinstall.go
```diff
@@ -237,9 +301,35 @@ func (pi *PackageInstallCR) referencedPkgVersion() (datapkgingv1alpha1.Package,
...
+		return pi.clusterVersionConstraintsSatisfied(&pkg, v)
+	}
+	kcConstraint := func(pkgVer string) bool {
+		pkg := versionToPkg[pkgVer]
+		return pi.kcVersionConstraintsSatisfied(&pkg)
+	}
 
-	selectedVersion, err := versions.HighestConstrainedVersion(versionStrs, verConfig)
+	vcc := []versions.ConstraintCallback{
+		{Constraint: k8sConstraint, Name: "kubernetes-version-check"},
```
cppforlife: should this constraint be only conditionally added when requiresClusterVersion is true? i dunno the impact of running clusterVersionConstraintsSatisfied with a blank version.

neil-hickey: I guess the only consequence is that we will always have a constraint showing in the error message for the k8s version check, even if the user never specified that constraint,

neil-hickey: I added a test to show this Test_Package_ConstraintNotGiven_ErrorDoesNotContainMessage

PR #799, opened 2022-07-26
Surface namespace and GK resources in AppCR status

### LINKED ISSUE #535: surface in App CR status set of namespaces used for the deployment

it's useful for users of App CR and PackageInstall CR to know which namespaces hold deployed resources. internally `kapp ls` presents a list of used namespaces, however such information would only be accessible to direct kapp CLI users. it would be nice to somehow surface this info in App CR status, e.g.

```yaml
status:
  deploy:
     usedNamespaces:
     - "" # empty to represent cluster-scope
     - knative-serving
     - knative-eventing
```

this problem might be also generally related to letting users identify what are resources that are part of this app/packageinstall.

---
Vote on this request

DennisDenuto:
TODO:

- [x] do code TODO's
- [x] remove any hardcoding left over
- [x] fix up commits
- [x] fix linter errors / code styling errors
- [x] remove kapp linux binary change from PR

- Allow the ability to discover what k8s resources belong to an App CR

Fixes #535

```release-note
- namespaces and GK's are set on AppCR.KappDeployStatus.AssociatedResources
```

##### Review Checklist:

  a link to that PR
  change

```docs

```


100mik:
Oooh! This will really help us put together something like `kctrl  app inspect -a app-name` which lets users inspect resources created by kapp for an app

pkg/app/app_reconcile.go
```diff
@@ -164,7 +164,28 @@ func (a *App) updateLastDeploy(result exec.CmdRunResult) exec.CmdRunResult {
 		UpdatedAt: metav1.NewTime(time.Now().UTC()),
 	}
 
-	a.updateStatus("marking last deploy")
+	defer func() {
+		a.updateStatus("marking last deploy")
```
neil-hickey: This can just be `defer a.updateStatus("marking last deploy")`

pkg/deploy/kapp.go
```diff
@@ -38,15 +41,20 @@ var _ Deploy = &Kapp{}
...
 
 // Deploy takes the output from templating, and the app name,
 // it shells out, running kapp deploy ...
 func (a *Kapp) Deploy(tplOutput string, startedApplyingFunc func(),
 	changedFunc func(exec.CmdRunResult)) exec.CmdRunResult {
 
-	args, err := a.addDeployArgs([]string{"deploy", "--prev-app", a.oldManagedName(), "-f", "-"})
+	metadataFile := fmt.Sprintf("/etc/kappctrl-mem-tmp/metadata-%s", a.genericOpts.Name)
+	defer func() {
+		os.Remove(metadataFile)
```
neil-hickey: can be just `defer os.Remove(metadataFile)`

pkg/deploy/kapp.go
```diff
@@ -38,15 +41,20 @@ var _ Deploy = &Kapp{}
...
+	return &Kapp{appSuffix, opts, genericOpts, globalDeployRawOpts, cancelCh, cmdRunner, nil}
 }
 
 // Deploy takes the output from templating, and the app name,
 // it shells out, running kapp deploy ...
 func (a *Kapp) Deploy(tplOutput string, startedApplyingFunc func(),
 	changedFunc func(exec.CmdRunResult)) exec.CmdRunResult {
 
-	args, err := a.addDeployArgs([]string{"deploy", "--prev-app", a.oldManagedName(), "-f", "-"})
+	metadataFile := fmt.Sprintf("/etc/kappctrl-mem-tmp/metadata-%s", a.genericOpts.Name)
```
cppforlife: use memdir.NewTmpDir since we dont want to be spilling out details about special /etc/kappctrl-mem-tmp location in multiple places.

pkg/deploy/kapp.go
```diff
@@ -231,3 +250,13 @@ func (a *Kapp) addGenericArgs(args []string, appName string) ([]string, []string
 
 	return args, env
 }
+
+// tryCacheAppConfigMap if unable to cache the kapp configmap continue and do not fail the deploy.
+func (a *Kapp) tryCacheAppConfigMap(metadataFileName string) {
+	metadataFile, err := os.ReadFile(metadataFileName)
```
neil-hickey, line 277: I suppose here if we somehow fail to read the file:

We will end up with stale metadata right (could be leftover in our cache from previous deploy)? Do we want to leave it as the old one? Or wipe it on failure?

pkg/deploy/kapp.go
cppforlife: nit: `a.genericOpts.Name` would not have been unique since apps can be in multiple namespaces but have same name.

test/e2e/kappcontroller/app_status_test.go
```diff
@@ -124,6 +125,129 @@ spec:
...
+
+	cleanUpApp()
+	defer cleanUpApp()
+
+	logger.Section("deploy", func() {
+		out, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, e2e.RunOpts{StdinReader: strings.NewReader(appYaml), AllowError: true})
+		assert.NoError(t, err)
+		assert.Contains(t, out, "ok: reconcile role/kappctrl-e2e-ns-role")
+		assert.Contains(t, out, "ok: reconcile rolebinding/kappctrl-e2e-ns-role-binding")
+		assert.Contains(t, out, "ok: reconcile serviceaccount/kappctrl-e2e-ns-sa")
```
cppforlife: no need to check for output

test/e2e/kappcontroller/app_status_test.go
```diff
@@ -124,6 +125,129 @@ spec:
...
+	cleanUpApp := func() {
+		kapp.Run([]string{"delete", "-a", name})
+	}
+
+	cleanUpApp()
+	defer cleanUpApp()
+
+	logger.Section("deploy", func() {
+		out, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, e2e.RunOpts{StdinReader: strings.NewReader(appYaml), AllowError: true})
+		assert.NoError(t, err)
```
cppforlife: no need to check for success if you remove AllowError: true from opts

test/e2e/kappcontroller/app_status_test.go
```diff
@@ -124,6 +125,129 @@ spec:
...
+            kind: ConfigMap
+            metadata:
+              name: configmap
+            data:
+              key: value
+  template:
+    - ytt: {}
+  deploy:
+    - kapp:
+        inspect: {}
```
cppforlife: no need to have `inspect` configuration

test/e2e/kappcontroller/app_status_test.go
```diff
@@ -124,6 +125,129 @@ spec:
...
+  fetch:
+    - inline:
+        paths:
+          file.yml: |
+            apiVersion: v1
+            kind: ConfigMap
+            metadata:
+              name: configmap
+            data:
+              key: value
```
cppforlife: i wonder if we can add any other "cheap" resource to see non-empty group in GKs array.

pkg/app/app_reconcile.go
```diff
@@ -164,7 +164,28 @@ func (a *App) updateLastDeploy(result exec.CmdRunResult) exec.CmdRunResult {
 		UpdatedAt: metav1.NewTime(time.Now().UTC()),
 	}
 
-	a.updateStatus("marking last deploy")
+	defer func() {
```
neil-hickey: can be just `defer a.updateStatus("marking last deploy")`

pkg/app/app_deploy.go
```diff
@@ -122,6 +124,39 @@ func (a *App) inspect() exec.CmdRunResult {
...
+		Namespaces []string `yaml:"namespaces"`
+	} `yaml:"lastChange"`
+	UsedGKs []struct {
+		Group string `yaml:"Group"`
+		Kind  string `yaml:"Kind"`
+	} `yaml:"usedGKs"`
+}
+
+// tryCacheMetadata if unable to cache the kapp configmap into an App meta continue and do not fail the deploy.
+func (a *App) tryCacheMetadata(kapp *ctldep.Kapp) {
```
neil-hickey: I think the word `cache` threw me off a bit. I was expecting a cache to be something that has a TTL / get's invalidated. I don't know if `save` is better or not.... It feels more like a variable than any sort of caching. It's holding data until something overwrites it, rather than holding data so access is faster / easier. I don't think this cache saves any hits on the API (kapp deploy each time is required) so I guess that's why I am feeling this way.

This was longer than I expected... I hope it makes sense

pkg/app/app_deploy.go
```diff
@@ -122,6 +124,39 @@ func (a *App) inspect() exec.CmdRunResult {
...
+	LabelKey   string `yaml:"labelKey"`
+	LabelValue string `yaml:"labelValue"`
+	LastChange struct {
+		Namespaces []string `yaml:"namespaces"`
+	} `yaml:"lastChange"`
+	UsedGKs []struct {
+		Group string `yaml:"Group"`
+		Kind  string `yaml:"Kind"`
+	} `yaml:"usedGKs"`
+}
```
cppforlife: the structure and unmarshaling is more of an implementation detail of kapp. dont think higher level layer (like app) should concern itself with that. propose we return this struct from kapp class's InternalAppConfigMap (though with a different name).

pkg/app/app.go
```diff
@@ -37,6 +37,7 @@ type App struct {
 
 	pendingStatusUpdate   bool
 	flushAllStatusUpdates bool
+	Metadata              *Meta
```
cppforlife: seems like this field keeps intermediate state, if that's best we can do (vs passing that data between funcs directly), it should at least be private.

pkg/app/app_deploy.go
```diff
@@ -122,6 +124,39 @@ func (a *App) inspect() exec.CmdRunResult {
...
+// Meta contains app meta allowing for an AppCR to surface the associated namespaces and GKs
+type Meta struct {
+	LabelKey   string `yaml:"labelKey"`
+	LabelValue string `yaml:"labelValue"`
+	LastChange struct {
+		Namespaces []string `yaml:"namespaces"`
+	} `yaml:"lastChange"`
+	UsedGKs []struct {
+		Group string `yaml:"Group"`
+		Kind  string `yaml:"Kind"`
```
cppforlife: is first letter really uppercased as returned by kapp?

test/e2e/kappcontroller/app_status_test.go
```diff
@@ -124,6 +125,129 @@ spec:
...
+		Deploy: &v1alpha1.AppStatusDeploy{
+			ExitCode: 0,
+			Finished: true,
+			Error:    "",
+			KappDeployStatus: v1alpha1.KappDeployStatus{
+				AssociatedResources: v1alpha1.AssociatedResources{
+					Label:      expectedAppLabel,
+					Namespaces: []string{"kappctrl-test"},
+					GroupKinds: []metav1.GroupKind{
+						{"", "ConfigMap"},
```
cppforlife: can we explicitly call out field name on this struct just to make it easier to see that it's not an array.

pkg/deploy/kapp.go
cppforlife: /etc/kappctrl-mem-tmp is more of an implementation detail of memdir.TmpDir. lets use that class to create a directory. this will also fix a bug above where we only use name of an app vs name+namespace.

pkg/app/app_reconcile.go
```diff
@@ -164,7 +164,28 @@ func (a *App) updateLastDeploy(result exec.CmdRunResult) exec.CmdRunResult {
...
+	usedGKs := []metav1.GroupKind{}
+	for _, gk := range a.Metadata.UsedGKs {
+		usedGKs = append(usedGKs, metav1.GroupKind{
+			gk.Group, gk.Kind,
+		})
+	}
+
+	a.app.Status.Deploy.KappDeployStatus = v1alpha1.KappDeployStatus{
+		AssociatedResources: v1alpha1.AssociatedResources{
+			Label:      fmt.Sprintf("%v=%v", a.Metadata.LabelKey, a.Metadata.LabelValue),
```
cppforlife: %s since it's just strings

pkg/app/app_deploy.go
praveenrewar: Surprisingly, yes. We actually don't have json tags for GVs and GKs. I am thinking that we should add those, but then again we will have to think about backwards compatibility 😞

pkg/app/app_deploy.go
DennisDenuto: sgtm; `s/cache/save/`

pkg/app/app_reconcile.go
DennisDenuto: 👍

pkg/deploy/kapp.go
DennisDenuto: I updated it to wipe on failure. I think it will help signal the error case a bit better.

> could be leftover in our cache from previous deploy

In practice I don't think the state would be left over from previous delpoy since a new kapp object is created. However that may change in the future and the api should be robust enough to handle it

pkg/deploy/kapp.go
DennisDenuto: 👍

pkg/app/app_reconcile.go
DennisDenuto: 👍

pkg/app/app_reconcile.go
DennisDenuto: changed

pkg/deploy/kapp.go
DennisDenuto: great catch! updated to use tmpdir.

if the tmp dir fails to create however, it will cause the deploy to fail.

pkg/app/app_deploy.go
DennisDenuto: fair enough! changed. went with the name `AppMeta` instead of `InternalAppConfigMap`

pkg/app/app.go
DennisDenuto: made it private

pkg/deploy/kapp.go
DennisDenuto: great catch! updated to use tmpdir.

if the tmp dir fails to create however, it will cause the deploy to fail.

cppforlife: i think that is a good behaviour.

DennisDenuto: Yeah, it will fail with an error message closer to the root cause (rather than the error message returned by `kapp deploy --app-meta-output`)

pkg/deploy/kapp.go
```diff
@@ -38,15 +43,25 @@ var _ Deploy = &Kapp{}
...
-	args, err := a.addDeployArgs([]string{"deploy", "--prev-app", a.oldManagedName(), "-f", "-"})
+	tmpMetadataDir := memdir.NewTmpDir("app_metadata")
+	defer tmpMetadataDir.Remove()
+
+	err := tmpMetadataDir.Create()
+	if err != nil {
+		return exec.NewCmdRunResultWithErr(err)
+	}
+
+	metadataFile := filepath.Join(tmpMetadataDir.Path(), a.genericOpts.Name)
```
cppforlife: i think it's probably better to have a stable name to avoid implying that using a name provides uniqueness here (tmp dir is unique in itself). i would switch to `filepath.Join(tmpMetadataDir.Path(), "app-metadata.yml")` or something like this.

pkg/deploy/kapp.go
```diff
@@ -231,3 +270,20 @@ func (a *Kapp) addGenericArgs(args []string, appName string) ([]string, []string
...
+func (a *Kapp) trySaveAppMeta(metadataFileName string) {
+	metadataFile, err := os.ReadFile(metadataFileName)
+	if err != nil {
+		a.appMeta = nil
+		return
+	}
+
+	appMetadata := Meta{}
+	err = yaml.Unmarshal(metadataFile, &appMetadata)
+	if err != nil {
```
cppforlife, line 284: reset a.appMeta = nil here as well? or should we just do it at the top of the Deploy() function so that we dont have to think about all the error cases.

test/e2e/kappcontroller/app_status_test.go
cppforlife: @DennisDenuto ^ wdyt?

test/e2e/kappcontroller/app_status_test.go
```diff
@@ -124,6 +125,117 @@ spec:
...
+			KappDeployStatus: v1alpha1.KappDeployStatus{
+				AssociatedResources: v1alpha1.AssociatedResources{
+					Label:      expectedAppLabel,
+					Namespaces: []string{"kappctrl-test"},
+					GroupKinds: []metav1.GroupKind{
+						{Group: "", Kind: "ConfigMap"},
+					},
+				},
+			},
+		},
```
cppforlife: since this test specifically focues on kapp deploy status, let's simplify assertion to only care about status.Deploy section or even just status.Deploy.KappDeployStatus

pkg/apis/kappctrl/v1alpha1/types.go
```diff
@@ -162,6 +162,22 @@ type AppStatusDeploy struct {
 	StartedAt metav1.Time `json:"startedAt,omitempty"`
 	// +optional
 	UpdatedAt metav1.Time `json:"updatedAt,omitempty"`
+	// +optional
+	KappDeployStatus KappDeployStatus `json:"kapp,omitempty"`
```
cppforlife: lets make this a pointer since technically we could be doing some other type of deploys (non-kapp; existing api leaves that as a possibility)

DennisDenuto: ok done

test/e2e/kappcontroller/app_status_test.go
DennisDenuto: yup! done

pkg/deploy/kapp.go
DennisDenuto: Nice! added it to the top of Deploy()

pkg/deploy/kapp.go
DennisDenuto: changed to use a stable name

test/e2e/kappcontroller/app_status_test.go
DennisDenuto: Ok! changed it to -> `rbac.authorization.k8s.io/v1` / `Role`

The test asserts on a non-empty Group (rbac.authorization.k8s.io)

pkg/app/app_deploy.go
```diff
@@ -122,6 +123,20 @@ func (a *App) inspect() exec.CmdRunResult {
 	return result
 }
 
+// trySaveMetadata if unable to save the kapp metadata into an App meta continue and do not fail the deploy.
+func (a *App) trySaveMetadata(kapp *ctldep.Kapp) {
+	if a.metadata != nil {
+		return
+	}
```
cppforlife: nit: i dont think we need this, since we probably would want to always use the "latest".

pkg/deploy/kapp.go
```diff
@@ -5,9 +5,13 @@ package deploy
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/memdir"
```
cppforlife: nit: per our usual style, non-std goes into a separate section.

test/e2e/kappcontroller/app_status_test.go
```diff
@@ -124,7 +125,78 @@ spec:
...
+					Namespaces: []string{"kappctrl-test"},
+					GroupKinds: []metav1.GroupKind{
+						{Group: "rbac.authorization.k8s.io", Kind: "Role"},
+					},
+				},
+			},
+		},
+	}
+
+	require.Equal(t, expectedStatus.Deploy.KappDeployStatus, cr.Status.Deploy.KappDeployStatus)
```
cppforlife, line 201: idea: it would be nice to add to this test a second step that makes deploy fail (and some resources have been deployed). e.g. have a second deploy that removes rbac, adds configmap, and has a failing job that makes deploy fail.

pkg/app/app_deploy.go
DennisDenuto: ok removed

pkg/deploy/kapp.go
DennisDenuto: ok changed

test/e2e/kappcontroller/app_status_test.go
DennisDenuto: good idea. it caught an error case that wasn't handled in kapp. See the associated kapp PR: https://github.com/vmware-tanzu/carvel-kapp/pull/574

I've added the test you suggested. Once a new kapp release is made with the bug fix, this test should go green and hopefully be in a state to merge.

praveenrewar: kapp [v0.52.0](https://github.com/vmware-tanzu/carvel-kapp/releases/tag/v0.52.0) is out. The tests should go green once we bump kapp version.

neil-hickey: merged kapp bump, rebased the PR. 👍

PR #808, opened 2022-08-01
use pull_request_target to allow access to secrets

benmoss:
Allows the "add-to-project" github action to work for PRs from non-contributors. GitHub seems to allow access to secrets from these actions when they come from people with certain privileges, but will fail from other users.

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/


aaronshurley:
LGTM

PR #826, opened 2022-08-12
Namespace isolation

### LINKED ISSUE #792: Different deployment mode to support good performance for 10k Apps

showpune:
When to start the kapp-controller, two  start options are added
1) start-api-server with boolean type indicate if start the api server
2_ namespace with string to indicate which namespace will be reconciled
In a cluster, we will start multiple reconcile controller in each namespace to reconcile the resource in the specific namespace, but only keep one to run the api server

Fixes #792

```release-note
When to start the kapp-controller, two start options are added
1) start-api-server with boolean type indicate if start the api server in the instance
2_ namespace with string to indicate which namespace will be reconciled
In a cluster, we will start multiple reconcile controller in each namespace to reconcile the resource in the specific namespace, but only keep one to run the api server
```

##### Review Checklist:

  a link to that PR
  change

```docs

```


joe-kimmel-vmw (approved):
LGTM

neil-hickey (approved):
LGTM

PR #833, opened 2022-08-17
Versions Spike Exploration PR

neil-hickey:
This is intended to be a WIP - feedback only PR. It will not get merged, and is a poc / learning opportunity PR instead.


config/crds.yml
```diff
@@ -427,6 +427,12 @@ spec:
                                                 fieldPath:
                                                   description: 'Required: Selects a field of the app: only annotations, labels, uid, name and namespace are supported.'
                                                   type: string
+                                                kubernetesVersion:
+                                                  description: 'Optional: Get Kubernetes version, defaults (empty) to retrieving the version from the cluster. Can be manually overridden to a value instead.'
+                                                  properties:
+                                                    version:
```
neil-hickey, line 439: this exact UX is not what we will end up on, just did it for simplicity / speed

pkg/template/factory.go
```diff
@@ -14,20 +16,22 @@ import (
...
+	kbldAllowBuild bool, cmdRunner exec.CmdRunner, log logr.Logger) Factory {
 
-	return Factory{coreClient, fetchFactory, kbldAllowBuild, cmdRunner}
+	return Factory{coreClient, fetchFactory, deployFactory, kbldAllowBuild, cmdRunner, log}
 }
 
 // NewYtt returns ytt template.
 func (f Factory) NewYtt(opts v1alpha1.AppTemplateYtt, appContext AppContext) *Ytt {
-	return NewYtt(opts, appContext, f.coreClient, f.fetchFactory, f.cmdRunner)
+	return NewYtt(opts, appContext, f.coreClient, f.fetchFactory, f.deployFactory, f.cmdRunner, f.log)
```
neil-hickey: only ytt actually implements the k8s version stuff right now... so ignore the other's, threading the deployfactory through them all was just to get them to compile

pkg/template/cue.go
```diff
@@ -59,7 +63,7 @@ func (c *cue) template(dirPath string, input io.Reader) exec.CmdRunResult {
 		args = append(args, c.opts.Paths...)
 	}
 
-	vals := Values{c.opts.ValuesFrom, c.appContext, c.coreClient}
+	vals := Values{c.opts.ValuesFrom, c.appContext, c.coreClient, c.deployFactory, c.log}
```
cppforlife: this is actually a good example where extra indirection via dep-inject + factory will help. right now newCue has the responsbility of creating Values{} -- because of that Value's dependencies are leaking into newCue (e.g. coreClient and deployFactory). what i would recommend doing is newCue to take a ValuesFactory. ValuesFactory would accept coreClient and deployFactory and hide all that from cue's usage: `vals := c.valuesFactory.NewValues(c.opts.ValuesFrom, c.appContext)`

config/crds.yml
cppforlife: i would recommend following api contract:

```
- name: ver
  kubernetesVersion: {}
```
more advanced case where overriding happens
```
- name: ver
  kubernetesVersion:
    version: "1.2.3"
```

pkg/template/cue.go
```diff
@@ -63,7 +57,7 @@ func (c *cue) template(dirPath string, input io.Reader) exec.CmdRunResult {
 		args = append(args, c.opts.Paths...)
 	}
 
-	vals := Values{c.opts.ValuesFrom, c.appContext, c.coreClient, c.deployFactory, c.log}
+	vals := Values{c.opts.ValuesFrom, c.appContext, nil, nil}
```
neil-hickey: gotta revisit this, maybe moving the values into their own factory as suggested by DK

pkg/template/downward_api_values.go
```diff
@@ -18,13 +18,14 @@ import (
...
+	items                 []v1alpha1.AppTemplateValuesDownwardAPIItem
+	metadata              PartialObjectMetadata
+	kubernetesVersion     func() (semver.Version, error)
+	kappControllerVersion string
 }
 
-// AsYAMLs returns many key-values queried (using jsonpath) against an object metadata provided.
-func (a DownwardAPIValues) AsYAMLs() ([][]byte, error) {
+// AsYAML returns many key-values queried against AppCR metadata or fetch-able information (such as the kubernetes version)
+func (a DownwardAPIValues) AsYAML() ([][]byte, error) {
```
cppforlife, line 28: nit: it was named plural because it's an array of array of bytes. typically one would expect AsYAML to return just []byte

pkg/template/cue.go
```diff
@@ -6,33 +6,27 @@ package template
...
 }
 
 var _ Template = &cue{}
 
-func newCue(opts v1alpha1.AppTemplateCue, appContext AppContext,
-	coreClient kubernetes.Interface, cmdRunner exec.CmdRunner, deployFactory deploy.Factory, log logr.Logger) *cue {
-
-	return &cue{opts: opts, appContext: appContext,
-		coreClient: coreClient, cmdRunner: cmdRunner, deployFactory: deployFactory, log: log}
+func newCue(opts v1alpha1.AppTemplateCue, appContext AppContext, cmdRunner exec.CmdRunner, valuesFactory ValuesFactory, fetchFactory fetch.Factory) *cue {
```
cppforlife, line 28: why does cue need fetchFactory? cue is all about templating so shouldnt be fetching anything.

pkg/template/values.go
```diff
@@ -6,27 +6,40 @@ package template
...
-	deployFactory deploy.Factory
-	log           logr.Logger
+	appContext     AppContext
+	versionFetcher *fetch.VersionFetch
+	coreClient     kubernetes.Interface
 }
 
+// ValuesFactory abstracts the factories for fetching the values away from the template factory
+type ValuesFactory struct {
+	appContext   AppContext
```
cppforlife, line 32: there is no need to keep appContext on ValuesFactory. it's injected at the NewValues time. this also means leads towards a good direction where ValuesFactory can be injected into TemplateFactory (or just created once), instead of everytime NewCue/Ytt/etc is called.

pkg/template/ytt.go
```diff
@@ -9,32 +9,29 @@ import (
...
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/fetch"
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/memdir"
-	"k8s.io/client-go/kubernetes"
 )
 
 type Ytt struct {
 	opts          v1alpha1.AppTemplateYtt
 	appContext    AppContext
-	coreClient    kubernetes.Interface
+	clusterClient *clusterclient.ClusterClient
```
cppforlife, line 22: Ytt is a templating tool, it should not have access to the cluster.

pkg/template/values.go
```diff
@@ -6,27 +6,40 @@ package template
...
-	coreClient    kubernetes.Interface
-	deployFactory deploy.Factory
-	log           logr.Logger
+	appContext     AppContext
+	versionFetcher *fetch.VersionFetch
+	coreClient     kubernetes.Interface
 }
 
+// ValuesFactory abstracts the factories for fetching the values away from the template factory
+type ValuesFactory struct {
```
cppforlife, line 31: nit: i would move ValuesFactory into a new file (or at least before or after Values class). right now it's litterally splitting Values class into 2 pieces which makes it hard to read.

pkg/template/factory.go
```diff
@@ -4,34 +4,33 @@
...
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/clusterclient"
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/exec"
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/fetch"
-	"k8s.io/client-go/kubernetes"
 )
 
 // Factory allows to build various templaters e.g. ytt, cue.
 type Factory struct {
-	coreClient     kubernetes.Interface
+	clusterClient  *clusterclient.ClusterClient
```
cppforlife, line 15: general comment (but i already left a few specific once). since we now have ValuesFactory, it's nice to make template.Factory not have access to a cluster. ValuesFactory can be injected into template.Factory from outside.

pkg/pkgrepository/app_test.go
```diff
@@ -4,6 +4,7 @@
 package pkgrepository_test
 
 import (
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/clusterclient"
```
cppforlife, line 7: nit: move to non-std imports section (there are a few other places).

pkg/fetch/version_fetcher.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// VersionFetch allows retrieving versions for components of the system
+type VersionFetch struct {
+	clusterClient     *clusterclient.ClusterClient
+	controllerVersion string
+}
+
+// NewVersionFetcher creates a VersionFetch object based on the clusterClient (for kubernetes access, and the static controller version)
+func NewVersionFetcher(clusterclient *clusterclient.ClusterClient, controllerVersion string) *VersionFetch {
```
cppforlife, line 29: generally it's a good to avoid do-er objects:
- https://www.yegor256.com/2015/03/09/objects-end-with-er.html
- http://objology.blogspot.com/2011/09/one-of-best-bits-of-programming-advice.html

pkg/fetch/version_fetcher.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+// VersionFetcher supplies a generic interface for retrieving versions for components of the system
+type VersionFetcher interface {
+	GetKappControllerVersion() string
+	GetKubernetesVersion() semver.Version
+}
+
+// VersionFetch allows retrieving versions for components of the system
+type VersionFetch struct {
```
cppforlife, line 23: not a fan of putting version discovery into a fetch package since fetch package is more about fetching deployable content which version is not. i think naming it "Fetcher" leads to believe that it's a good fit, but i think that's more of a coincidence in name. everything in theory could be "fetched" from something.

pkg/deploy/factory.go
```diff
@@ -33,18 +26,16 @@ type KappConfiguration interface {
...
 }
 
 // NewKapp configures and returns a deployer of type Kapp
 func (f Factory) NewKapp(opts v1alpha1.AppDeployKapp, saName string,
-	clusterOpts *v1alpha1.AppCluster, genericOpts GenericOpts, cancelCh chan struct{}) (*Kapp, error) {
+	clusterOpts *v1alpha1.AppCluster, genericOpts clusterclient.GenericOpts, cancelCh chan struct{}) (*Kapp, error) {
 
-	processedGenericOpts, err := f.processOpts(saName, clusterOpts, genericOpts)
+	processedGenericOpts, err := f.clusterClient.ProcessOpts(saName, clusterOpts, genericOpts)
+	fmt.Println(processedGenericOpts.Kubeconfig)
```
cppforlife, line 38: printing...

pkg/clusterclient/cluster_client.go
```diff
@@ -0,0 +1,60 @@
...
+		coreClient:        coreClient,
+		kubeconfigSecrets: NewKubeconfigSecrets(coreClient),
+		serviceAccounts:   NewServiceAccounts(coreClient, log),
+		log:               log,
+	}
+}
+
+// ProcessOpts takes generic opts and a ServiceAccount Name, and returns a populated kubeconfig that can connect to a cluster.
+// if the saName is empty then you'll connect to a cluster via the clusterOpts inside the genericOpts, otherwise you'll use the specified SA.
+func (c ClusterClient) ProcessOpts(saName string, clusterOpts *v1alpha1.AppCluster, genericOpts GenericOpts) (ProcessedGenericOpts, error) {
```
cppforlife, line 39: try to find a different name for this method. right now it doesnt mean anything to the consumers of this class. (it was private purposely because it's an auxiliary/internal functionality and contextual to the deployFactory, but if you plan to elevate to public is has to have meaning. what is the noun we are working with?).

pkg/clusterclient/cluster_client.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"fmt"
+	"github.com/go-logr/logr"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// ClusterClient provides access to the kubernetes cluster
+// It initializes the service-account token cache
+type ClusterClient struct {
```
cppforlife, line 15: if it's indeed a client consumers shouldn't ask it for another client (CoreClient). right now this class is behaving like a factory for CoreClients, which is totally fine but then name should reflect that -- ClusterClientFactory. to that, i would also recommend renaming CoreClient() to NewThisClusterCoreClient() to indicate that it's the client where this code is running.

cppforlife: tbh seeing usage of this factory in code below (for the sake of CoreClient) does not seem compelling as most things use "this cluster core client" so factory does not add much. i would be in favor of injecting coreClient directly (as it was done) and only inject this factory in places where comm with "other" cluster is necessary.

pkg/fetch/version_fetcher.go
```diff
@@ -0,0 +1,76 @@
...
+// NewVersionFetcher creates a VersionFetch object based on the clusterClient (for kubernetes access, and the static controller version)
+func NewVersionFetcher(clusterclient *clusterclient.ClusterClient, controllerVersion string) *VersionFetch {
+	return &VersionFetch{
+		clusterClient:     clusterclient,
+		controllerVersion: controllerVersion,
+	}
+}
+
+// GetKappControllerVersion returns the live kapp-controller version
+func (vf VersionFetch) GetKappControllerVersion() string {
```
cppforlife, line 37: nit: drop "Get" naming. it's redundant.

pkg/fetch/version_fetcher.go
cppforlife: i think a better place is in clusterclient package since it's cluster related functionality.

pkg/app/app_deploy.go
```diff
@@ -5,6 +5,7 @@ package app
 
 import (
 	"fmt"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/clusterclient"
```
joaopapereira, line 8: nit: should this import be near the group below?

pkg/app/app_reconcile_test.go
```diff
@@ -5,6 +5,7 @@ package app
 
 import (
 	"fmt"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/clusterclient"
```
joaopapereira, line 8: nit: should this import be near the group below?

pkg/app/app_test.go
```diff
@@ -4,6 +4,7 @@
 package app_test
 
 import (
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/clusterclient"
```
joaopapereira, line 7: nit: should this import be near the group with non-stdlib imports?

pkg/clusterclient/cluster_client.go
joaopapereira: After thinking a bit about this:
- I do not think we should mix `kubernetes.Interface` getter with kubeconf generator. They feel like 2 different concepts.
- Roughly looking at the code I think that we would be ok with passing `kubernetes.Interface` and `clusterclient.ClusterConnectionGenerator` (this last one is just an interface that given `saName string, clusterOpts *v1alpha1.AppCluster, genericOpts GenericOpts` can generate the `ProcessedGenericOpts` for you)
-

pkg/clusterclient/cluster_client.go
joaopapereira: What about `ClusterConnectionConfiguration` and also change the `ProcessedGenericOpts` to `ClusterConnectionConfiguration`?

pkg/deploy/kapp.go
```diff
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/clusterclient"
```
joaopapereira, line 10: nit: should this import be near the group with non-stdlib imports?

pkg/fetch/version_fetcher.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/k14s/semver/v4"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/clusterclient"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// VersionFetcher supplies a generic interface for retrieving versions for components of the system
+type VersionFetcher interface {
```
joaopapereira, line 17: I could not find a place where this interface was used.
According to go's best practices the interfaces should be defined near the place they are used. Which will allow you to have smaller interfaces

pkg/fetch/version_fetcher.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+}
+
+// GetKappControllerVersion returns the live kapp-controller version
+func (vf VersionFetch) GetKappControllerVersion() string {
+	return vf.controllerVersion
+}
+
+// GetKubernetesVersion returns the live kubernetes version (local or external cluster based on the spec)
+func (vf VersionFetch) GetKubernetesVersion(saName string, specCluster *v1alpha1.AppCluster, genericOpts clusterclient.GenericOpts) (semver.Version, error) {
```
joaopapereira, line 42: nit: drop "Get" naming. it's redundant.

pkg/packageinstall/packageinstall.go
```diff
@@ -248,12 +237,15 @@ func (pi *PackageInstallCR) kcVersionConstraintsSatisfied(pkg *datapkgingv1alpha
 		return true
 	}
 
-	v, err := dropVersionPreReleaseAndBuild(pi.controllerVersion)
+	v, err := semver.ParseTolerant(pi.fetchFactory.NewVersionFetcher().GetKappControllerVersion())
```
joaopapereira, line 240: I think we should not provide a factory in this case. What we should provide is an interface that looks like this

```go
type KappControllerVersionRetriever interface {
  KappControllerVersion() (semver.Version, error)
  KubernetesVersion(saName string, specCluster *v1alpha1.AppCluster, genericOpts clusterclient.GenericOpts) (semver.Version, error)
}
```
This will allow you to inject into this struct the struct that implements this interface instead of having to build a full factory that will create the struct for you.

pkg/packageinstall/packageinstall_deletion_test.go
```diff
@@ -4,14 +4,15 @@
 package packageinstall
 
 import (
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/clusterclient"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/fetch"
```
joaopapereira, line 8: nit: should this import be near the group with non-stdlib imports?

pkg/template/cue.go
```diff
@@ -6,29 +6,27 @@ package template
 import (
 	"bytes"
 	"fmt"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/fetch"
```
joaopapereira, line 9: nit: should this import be near the group with non-stdlib imports?

pkg/template/cue.go
joaopapereira: Also doesn't look like it is used

pkg/template/helm_template.go
```diff
@@ -6,21 +6,23 @@ package template
 import (
 	"bytes"
 	"fmt"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/clusterclient"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/fetch"
```
joaopapereira, line 10: nit: should this import be near the group with non-stdlib imports?

joaopapereira (commented):
I left some comments of the PR, but I am more than happy to sit down and look at the code together with you

neil-hickey:
https://github.com/vmware-tanzu/carvel-kapp-controller/pull/846

Issue #837, opened 2022-08-22
Feedback on kctrl from `Graham Dumpleton`


rohitagg2020:
Feedback on kctrl from `Graham Dumpleton`

- [x] [Doc] Cert-manager example misses `kctrl pkg release -v 1.0.0` command. Fixed before release.
- [x] [Doc] Add the `kctrl pkg init` command in case of cert-manager example.
- [x] [Bug] If `kctrl pkg init` is not run successfully, then running `kctrl pkg release` fail with panic. Fixed before release.
- [x] `--local` and `--build` flags are missing while running `kctrl dev` command.
- [x] [Doc] Enhance `kctrl dev` documentation. https://kubernetes.slack.com/archives/CH8KCCKA5/p1660799313014129?thread_ts=1660781243.782389&cid=CH8KCCKA5
- [x] [Doc]There are 404 errors while clicking on links.
- [x] #934
- [x] [Doc] In case a pkg init is rerun, explicitly mentions what app spec will get overwritten(of packageBuild or package in package-resources.yml).
- [x] [Code] Pkg init rerun overwrite PackageMetadata changes.
- [x] #933
- [x] [Code] `kctrl pkg release` cmd will fail with error `no .imgpkg/images.yml file` even if YAML supplied is broken
- [x] [Doc/Code] In the above scenario, creating an empty `.imgpkg/images.yml` file will make the cmd succeed but no carvel-artifacts directory is created.
- [x] [Code/Doc] `kctrl` consumer flow: PackageInstall fails as it is not able to authenticate with registry.
- [x] [Doc] Add proper documentation on creating service account and rbac for `kctrl dev` flow.  
- [x] #932
- [x] While consuming the package, it is trying to install in the default namespace. This gives the error: kctrl: Error: Creating sensitive resources in a shared namespace (default)(hint: Specify a namespace using the '-n' flag or use kubeconfig to change default namespace 'kubectl config set-context --current --namespace=private-namespace'.Or use '--dangerous-allow-allow-use-of-shared-namespace=default' to allow use of shared namespace)
- [x] #932

100mik:
Branched out from #831

100mik:
> [Doc] In case a pkg init is rerun, explicitly mentions what app spec will get overwritten(of packageBuild or package in package-resources.yml).

[...]

PR #842, opened 2022-08-23
kctrl: Handle errors in status after successful local reconciliation.

### LINKED ISSUE #700: `kctrl`: `Released-at` column should show user friendly value in case the value is not available
**What steps did you take:**
`kctrl package available list -p contour.tanzu.vmware.com`

**What happened:**
```bash
NAMESPACE  NAME                      VERSION                RELEASED-AT
test-ns    contour.tanzu.vmware.com  1.15.1+vmware.1-tkg.1  0001-01-01 00:00:00 +0000 UTC
```

**What did you expect:**
A more user friendly value for released-at or empty string when it's not available.

---
Vote on this request

100mik:
We need to check for errors that might have happened after successful reconciliations (which won't return an error) explicitly. This PR also adds a check that formats `ReleasedAt` values better.

Fixes parts of #837, namely :
- kctrl pkg release cmd will fail with error no .imgpkg/images.yml file even if YAML supplied is broken

Resolves #700 

```release-note
Handle errors while building images for releases better
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
+++ b/cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
@@ -4,9 +4,11 @@
 package release
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	appinit "github.com/vmware-tanzu/carvel-kapp-controller/cli/pkg/kctrl/cmd/app/init"
@@ -14,6 +16,7 @@ import (
 	cmdlocal "github.com/vmware-tanzu/carvel-kapp-controller/cli/pkg/kctrl/local"
 	"github.com/vmware-tanzu/carvel-kapp-controller/cli/pkg/kctrl/logger"
 	kcv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
+	fakekc "github.com/vmware-tanzu/carvel-kapp-controller/pkg/client/clientset/versioned/fake"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -83,9 +86,13 @@ func (b *AppSpecBuilder) Build() (kcv1alpha1.AppSpec, error) {
 	reconciler := cmdlocal.NewReconciler(b.depsFactory, cmdRunner, b.logger)
 
 	err = reconciler.Reconcile(buildConfigs, cmdlocal.ReconcileOpts{
-		Local:     true,
-		KbldBuild: true,
+		Local:             true,
+		KbldBuild:         true,
+		AfterAppReconcile: b.checkForErrorsAfterReconciliation,
 	})
+	if err != nil {
+		return kcv1alpha1.AppSpec{}, err
+	}
 
 	bundleURL := ""
 	useKbldImagesLock := false
@@ -130,4 +137,17 @@ func (b *AppSpecBuilder) Build() (kcv1alpha1.AppSpec, error) {
 	}
 
 	return appSpec, nil
+}
+
+func (b *AppSpecBuilder) checkForErrorsAfterReconciliation(app kcv1alpha1.App, fakeClient *fakekc.Clientset) error {
+	existingApp, err := fakeClient.KappctrlV1alpha1().Apps(app.Namespace).Get(context.Background(), app.Name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	// TODO: How can we prevent reconciler from trying to prepare kapp?
+	if existingApp.Status.UsefulErrorMessage != "" && !strings.Contains(existingApp.Status.UsefulErrorMessage, "Preparing kapp") {
+		return fmt.Errorf("Reconciling: %s", existingApp.Status.UsefulErrorMessage)
+	}
+	return nil
 }
--- a/cli/pkg/kctrl/cmd/package/available/list.go
+++ b/cli/pkg/kctrl/cmd/package/available/list.go
@@ -194,11 +194,18 @@ func (o *ListOptions) listPackages() error {
 			cmdcore.NewValueNamespace(pkg.Namespace),
 			uitable.NewValueString(pkg.Spec.RefName),
 			cmdcore.NewValueSemver(pkg.Spec.Version),
-			uitable.NewValueString(pkg.Spec.ReleasedAt.String()),
+			uitable.NewValueString(formatTimestamp(pkg.Spec.ReleasedAt)),
 		})
 	}
 
 	o.ui.PrintTable(table)
 
 	return err
+}
+
+func formatTimestamp(timestamp metav1.Time) string {
+	if timestamp.IsZero() {
+		return "-"
+	}
+	return timestamp.String()
 }
--- a/cli/pkg/kctrl/cmd/package/available/get.go
+++ b/cli/pkg/kctrl/cmd/package/available/get.go
@@ -173,7 +173,7 @@ func (o *GetOptions) show(client pkgclient.Interface, pkgName, pkgVersion string
 
 		row = append(row, []uitable.Value{
 			uitable.NewValueString(pkg.Spec.Version),
-			uitable.NewValueString(pkg.Spec.ReleasedAt.String()),
+			uitable.NewValueString(formatTimestamp(pkg.Spec.ReleasedAt)),
 			uitable.NewValueString(wordwrap.WrapString(pkg.Spec.CapactiyRequirementsDescription, 80)),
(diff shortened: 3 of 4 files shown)


cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
```diff
@@ -39,14 +41,17 @@ func NewAppSpecBuilder(depsFactory cmdcore.DepsFactory, logger logger.Logger, ui
 const (
 	LockOutputFolder = ".imgpkg"
 	LockOutputFile   = "images.yml"
+
+	inMemoryAppName      = "kctrl-builder"
```
100mik: I thought I would need these constants while checking for errors again, but apparently the "run after" function receives a copy of the app, so it is not needed.

100mik: So I am happy to remove the new constants if y'all think it is unnecessary

cli/pkg/kctrl/cmd/package/available/values_schema.go
```diff
@@ -15,10 +15,11 @@ import (
 // DataValueProperty holds the details of each property under Carvel package.spec.valuesSchema.openAPIv3.properties.
 // The example of the schema could be found at: https://carvel.dev/kapp-controller/docs/latest/packaging/#package-1
 // From above example, we would have the following:
-//   DataValueProperty.Key = "namespace"
-//   DataValueProperty.Type = "string"
-//   DataValueProperty.Description = "Namespace where fluent-bit will be installed."
-//   DataValueProperty.Default = "fluent-bit"
+//
```
100mik: Looks like my formatter added some pretty spaces in comments, happy to remove these as well if y'all think it is not needed

100mik (commented):
Might be nice to have a test for this as well

cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
```diff
@@ -131,3 +141,16 @@ func (b *AppSpecBuilder) Build() (kcv1alpha1.AppSpec, error) {
...
 }
+
+func (b *AppSpecBuilder) checkForErrorsAfterReconciliation(app kcv1alpha1.App, fakeClient *fakekc.Clientset) error {
+	existingApp, err := fakeClient.KappctrlV1alpha1().Apps(app.Namespace).Get(context.Background(), app.Name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	// TODO: How can we prevent reconciler from trying to prepare kapp?
+	if existingApp.Status.UsefulErrorMessage != "" && !strings.Contains(existingApp.Status.UsefulErrorMessage, "Preparing kapp") {
```
100mik, line 149: Looks like kapp-ctrl reconciler tries to "prepare kapp"([here](https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/pkg/deploy/factory.go#L59-L80)) even though `ReleaseCmdRunner` ignores kapp. This ensures that the an error thrown due to the absence of a service account does not make it fail.

Choosing to ignore this error in favour of ensuring release is not dependent on a service account on the cluster even though `kapp` is not being used.

cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
praveenrewar: > So I am happy to remove the new constants if y'all think it is unnecessary

It would definitely reduce some noise from the PR.

cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
```diff
@@ -131,3 +141,16 @@ func (b *AppSpecBuilder) Build() (kcv1alpha1.AppSpec, error) {
 
 	return appSpec, nil
 }
+
+func (b *AppSpecBuilder) checkForErrorsAfterReconciliation(app kcv1alpha1.App, fakeClient *fakekc.Clientset) error {
```
praveenrewar, line 142: `checkForErrorsAfterReconciliation` -> `afterAppReconcile` ?

cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
praveenrewar: I will have to take a closer look at how the reconciler works to be able to suggest something.

cli/pkg/kctrl/cmd/package/available/default_values.go
```diff
@@ -56,6 +56,7 @@ func (s PackageSchema) DefaultValues() ([]byte, error) {
 // This is adopted from k8s.io/apiextensions-apiserver/pkg/apiserver/schema/defaulting with 2 changes
 // 1. Prevent deep copy of int as it panics
 // 2. For type object depth first search to see if there is any property with default
+//
```
praveenrewar: nit: remove this to reduce noise in the PR

cli/pkg/kctrl/cmd/package/available/list.go
```diff
@@ -194,11 +194,18 @@ func (o *ListOptions) listPackages() error {
...
+			uitable.NewValueString(formatTimestamp(pkg.Spec.ReleasedAt)),
 		})
 	}
 
 	o.ui.PrintTable(table)
 
 	return err
 }
+
+func formatTimestamp(timestamp metav1.Time) string {
```
praveenrewar, line 206: Should we have a new type defined for timestamp just like Semver in cmdcore or would that be over engineering as we don't require comparison on it yet?

cli/pkg/kctrl/cmd/package/available/values_schema.go
praveenrewar: Yeah, let's remove it to reduce the extra noise.

cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
100mik: I like how it reads on line 91 🤔 
Still happy to switch it up if you feel strongly about this

cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
100mik: How an instance of `kapp` is created is defined at a Factory level in kapp-ctrl
So we cannot propagate values "up the tree"

I can imagine a boolean which disables deploy which can be passed to the constructor used to create App CR instances. But that does not make a lot of sense from a kapp-ctrl perspective and would be a change solely for `kctrl`.

cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
100mik: Done!

cli/pkg/kctrl/cmd/package/available/values_schema.go
100mik: Removed!

cli/pkg/kctrl/cmd/package/available/default_values.go
100mik: Done!

cli/pkg/kctrl/cmd/package/available/list.go
100mik: I think it might be a bit overkill because we can directly use functions scoped to `metav1.Time` for most purposes.

cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
```diff
@@ -131,3 +138,16 @@ func (b *AppSpecBuilder) Build() (kcv1alpha1.AppSpec, error) {
...
+	existingApp, err := fakeClient.KappctrlV1alpha1().Apps(app.Namespace).Get(context.Background(), app.Name, metav1.GetOptions{})
+	if err != nil {
```
rohitagg2020, line 144: Can fakeClient return err ?

100mik: In this context I would not think so, but I would keep a check just in case something ever changes there.
I do not think we should leave errors unchecked

cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
praveenrewar: hehe, I actually wanted to add the comment on line 91. 
(No strong opinions on it)

rohitagg2020:
LGTM

praveenrewar (approved):

cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
100mik: Sticking with this for now!

PR #846, opened 2022-08-25
Pass additional information to downward API (k8s version, kc version, k8s g/v)

### LINKED ISSUE #781: Expose cluster versions and capabilities to templating steps
see design proposal doc: https://hackmd.io/6FQhzRIPR_uceqtTqSIXaA

per proposal doc, add to both Helm template specifically and ytt/cue templaters more generally

Helm shouldn't be talking to the cluster; we should be talking the cluster and passing through things to helm. 

---
Vote on this request

neil-hickey:

Fixes #781

```release-note
Templating stages (ytt, helm, cue) now have the kubernetes version, available group/version APIs and kapp-controller version passed to them via data values, and enabled capabilites in helm.
```

##### Review Checklist:

  change
  a link to that PR

```docs

```


pkg/app/app_reconcile_test.go
```diff
@@ -223,3 +225,16 @@ func Test_TemplateError_DisplayedInStatus_UsefulErrorMessageProperty(t *testing.
 		reflect.DeepEqual(expectedStatus, crdApp.app.Status()),
 		fmt.Sprintf("Status is not same:\nExpected:\n%#v\nGot:\n%#v\n", expectedStatus, crdApp.app.Status()))
 }
+
+type FakeComponentInfo struct {
```
neil-hickey, line 229: For now this is kinda duplicated across the tests.... Probably a better centralized place for it...

pkg/app/app.go
```diff
@@ -4,6 +4,7 @@
 package app
 
 import (
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/componentInfo"
```
neil-hickey: this behaviour of GoLand is annoying.. ignore the import formatting, I will go fix it later before proper review

cmd/controller/run.go
```diff
@@ -17,9 +17,10 @@ import (
 	pkgclient "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/client/clientset/versioned"
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/app"
 	kcclient "github.com/vmware-tanzu/carvel-kapp-controller/pkg/client/clientset/versioned"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/componentInfo"
```
joaopapereira:
```suggestion
	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/componentinfo"
```

Changing the `I` to `i` since we try to keep all the package names lowercase. This comment will apply to all imports and also the folder names

pkg/componentInfo/component_info.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/k14s/semver/v4"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/kubeconfig"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// Info provides information about components of the system
+type Info interface {
```
joaopapereira: I would move this interface into `pkg/app/app.ComponentInfo`.
Why: Interfaces generally should be nearer to the place where they are used. This allows you to separate implementation from the needs of a particular package. In this case, we will use all the functions on the interface but in other parts that might not happen.

joaopapereira: As a rule of thumb that I use, I only store interfaces in the same package that declares a struct that implements them when:
1. This would be the only package consuming this interface/struct (very rare in occurrence)
2. This package is part of a library that is exported for others to use and implement their own structs. Normal case an example of that is [this library](https://github.com/vmware-tanzu/carvel-imgpkg/blob/develop/vendor/github.com/cppforlife/go-cli-ui/ui/interfaces.go#L7) 
3. There is some sort of dependency that can be circular an example of that is [this interface](https://github.com/vmware-tanzu/carvel-imgpkg/blob/develop/pkg/imgpkg/internal/util/level_logger.go#L10). In this example the function [NewIndentedLevelLogger](https://github.com/vmware-tanzu/carvel-imgpkg/blob/develop/pkg/imgpkg/internal/util/level_logger.go#L33) constructs a new logger based on another logger

pkg/deploy/kapp.go
```diff
@@ -16,7 +15,9 @@ import (
 
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/exec"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/kubeconfig"
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/memdir"
+	"gopkg.in/yaml.v2"
```
joaopapereira: What are we using this package for? (sorry couldn't find it.) Also doesn't Kubernetes use their own yaml library?

pkg/componentInfo/component_info.go
```diff
@@ -0,0 +1,88 @@
...
+
+// ComponentInfo provides information about components of system
+type ComponentInfo struct {
+	coreClient            kubernetes.Interface
+	clusterAccess         *kubeconfig.Kubeconfig
+	kappControllerVersion string
+}
+
+// NewComponentInfo returns a ComponentInfo
+func NewComponentInfo(coreClient kubernetes.Interface, clusterAccess *kubeconfig.Kubeconfig, kappControllerVersion string) *ComponentInfo {
```
joaopapereira: The `clusterAccess` parameter could also be an interface.

cmd/controller/run.go
```diff
@@ -170,20 +171,22 @@ func Run(opts Options, runLog logr.Logger) error {
 	refTracker := reftracker.NewAppRefTracker()
 	updateStatusTracker := reftracker.NewAppUpdateStatus()
 
-	// initialize deploy factory once - the deploy factory contains a service account token cache which should be only setup once.
-	deployFactory := deploy.NewFactory(coreClient, kcConfig, sidecarCmdExec, runLog)
+	// initialize cluster access once - it contains a service account token cache which should be only setup once.
+	kubeconf := kubeconfig.NewKubeconfig(coreClient, runLog)
+	compInfo := componentInfo.NewComponentInfo(coreClient, kubeconf, Version)
```
joe-kimmel-vmw: lol ok i know "real go programmers don't use whole words" or something,  but this definitely feels like i'm suddenly in the HR portal checking on my compensation info.   anyways it's probably unambiguous given the context.

pkg/componentInfo/component_info.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	return v, nil
+}
+
+// KubernetesVersion returns the running K8s version depending on AppSpec
+// If AppSpec points to external cluster, we use that k8s version instead
+func (ci *ComponentInfo) KubernetesVersion(serviceAccountName string, specCluster *v1alpha1.AppCluster, objMeta *metav1.ObjectMeta) (semver.Version, error) {
+	switch {
+	case len(serviceAccountName) > 0:
+		v, err := ci.coreClient.Discovery().ServerVersion()
```
joe-kimmel-vmw: here on this line,  we're always going to call the same `coreClient.Discovery` , right?   I wonder if we could cache this call the first time we make it so like :

`ci.GetK8sVersion()`

and then 
```
func ci.GetK8sVer() (semver.Version, error){
  if ci.cachedVer != "" {
     return ci.cachedVer, nil
 }
  // the logic that you currently have, but then store the result into ci.cachedVer, which i just made up.
```


point being i _think_ it's a legit assumption that if we're calling th version of the cluster we're hosted on, it can't change without us getting restarted, so we can cache the version of our own cluster and thus minimize API calls.

pkg/componentInfo/component_info.go
```diff
@@ -0,0 +1,88 @@
...
+		if err != nil {
+			return semver.Version{}, err
+		}
+
+		clientset, err := kubernetes.NewForConfig(config)
+		if err != nil {
+			return semver.Version{}, err
+		}
+
+		v, err := clientset.Discovery().ServerVersion()
```
joe-kimmel-vmw: here we could contemplate whether we could do a cache based on the clientset, but i think we would have no way of subscribing to like, "tell me when the cluster i'm managing gets upgraded" , yeah? so might be a lot harder.

one ~heuristic is to keep a cache that's valid for e.g. ~1 minute or some short duration at a time, so you basically say, "ok but if there's a ton of these all at once (e.g. when the controller restarts on an existing cluster and has to re-reconcile everything) then i'll limit my API requests"

pkg/componentInfo/component_info.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+		return semver.ParseTolerant(v.String())
+	default:
+		return semver.Version{}, fmt.Errorf("Expected service account or cluster specified")
+	}
+}
+
+// KubernetesAPIs returns the available kubernetes Group/Version resources
+func (ci *ComponentInfo) KubernetesAPIs() ([]string, error) {
+	groups, err := ci.coreClient.Discovery().ServerGroups()
```
joe-kimmel-vmw: similar question/comment as above - i think this is hard for us to generally cache, but is it worth caching for short time durations (on order 1 minute) since generally this kind of thing doesn't change very often.

pkg/deploy/factory.go
```diff
@@ -34,69 +28,40 @@ type KappConfiguration interface {
...
+	return Factory{coreClient, kappConfig, kubeconfig, cmdRunner}
 }
 
 // NewKapp configures and returns a deployer of type Kapp
 func (f Factory) NewKapp(opts v1alpha1.AppDeployKapp, saName string,
-	clusterOpts *v1alpha1.AppCluster, genericOpts GenericOpts, cancelCh chan struct{}) (*Kapp, error) {
+	clusterOpts *v1alpha1.AppCluster, cancelCh chan struct{}, location kubeconfig.AccessLocation) (*Kapp, error) {
 
-	processedGenericOpts, err := f.processOpts(saName, clusterOpts, genericOpts)
+	clusterAccess, err := f.kubeconfig.ClusterAccess(saName, clusterOpts, location)
```
joe-kimmel-vmw, line 41: congratulations on refactoring away that `processedGenericOpts` biz.

pkg/componentInfo/component_info.go
```diff
@@ -0,0 +1,88 @@
...
+		return semver.Version{}, err
+	}
+	return v, nil
+}
+
+// KubernetesVersion returns the running K8s version depending on AppSpec
+// If AppSpec points to external cluster, we use that k8s version instead
+func (ci *ComponentInfo) KubernetesVersion(serviceAccountName string, specCluster *v1alpha1.AppCluster, objMeta *metav1.ObjectMeta) (semver.Version, error) {
+	switch {
+	case len(serviceAccountName) > 0:
```
joe-kimmel-vmw: it's slightly a bummer that this ~unintuitive check is still duplicated here and in the ClusterAccess code.  I'm guessing you already wrestled with this knot and this was the best solution you saw.  maybe worth thinking if there's a simplification that puts this check into exactly one place.

pkg/componentInfo/component_info.go
```diff
@@ -0,0 +1,88 @@
...
+		v, err := ci.coreClient.Discovery().ServerVersion()
+		if err != nil {
+			return semver.Version{}, err
+		}
+		return semver.ParseTolerant(v.String())
+	case specCluster != nil:
+		accessInfo, err := ci.clusterAccess.ClusterAccess(serviceAccountName, specCluster, kubeconfig.AccessLocation{Name: objMeta.Name, Namespace: objMeta.Namespace})
```
joe-kimmel-vmw: hmmm especially considering that in the second case here we _do_ delegate to ClusterAccess.   ...  is there a way to delegate the whole thing to cluster access so that there's only one file that checks the saname length, even if that one file has two codepaths that check it?    i guess that would look like moving the bulk of this function into the clusterAccess ...   how would that feel?

pkg/packageinstall/packageinstall.go
```diff
@@ -248,12 +236,14 @@ func (pi *PackageInstallCR) kcVersionConstraintsSatisfied(pkg *datapkgingv1alpha
 		return true
 	}
 
-	v, err := dropVersionPreReleaseAndBuild(pi.controllerVersion)
+	v, err := pi.compInfo.KappControllerVersion()
```
joe-kimmel-vmw, line 244: this is cleaner, nice

pkg/template/downward_api_values.go
```diff
@@ -9,33 +9,69 @@ import (
...
+		switch {
+		case item.FieldPath != "":
+			var fieldPathExpression string
+			fieldPathExpression, err = relaxedJSONPathExpression(item.FieldPath)
+			if err != nil {
+				return nil, err
+			}
+			keyValueContent, err = a.extractFieldPathAsKeyValue(item.Name, fieldPathExpression)
+		case item.KubernetesVersion != nil:
+			if item.KubernetesVersion.Version == "" { // I wish there was a ternary operator in Go
```
joe-kimmel-vmw, line 53: +1

pkg/template/values_factory.go
```diff
@@ -0,0 +1,22 @@
...
+
+package template
+
+import (
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
+	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/fetch"
+	"k8s.io/client-go/kubernetes"
+)
+
+// ValuesFactory abstracts the factories for fetching the values away from the template factory
```
joe-kimmel-vmw: lol wat

joe-kimmel-vmw (approved):
I have some comments and questions but i wouldn't argue with merging this as-is.  we can catch up real quick tomorrow AM about whether it's worth addressing any of these in this iteration.

pkg/template/helm_template.go
```diff
@@ -71,8 +75,15 @@ func (t *HelmTemplate) template(dirPath string, input io.Reader) exec.CmdRunResu
 
 	args := []string{"template", name, chartPath, "--namespace", namespace, "--include-crds"}
 
+	if t.opts.KubernetesVersion != nil {
+		args = append(args, []string{"--kube-version", t.valuesFactory.additionaDownwardAPIValues.KubernetesVersion.String()}...)
```
joaopapereira:
```suggestion
		args = append(args, []string{"--kube-version", vals.DownwardAPI().KubernetesVersion}...)
```

My suggestion here is to promote the AdditionalDownwardAPIValues to a primary concern in the Values struct.
This struct should contain all the values that you might eventually need to add to the templating phase.
Also given that the values are supposed to be all strings we should not need to do a `.String()` call when we are using this from the Values struct

pkg/template/helm_template.go
```diff
@@ -71,8 +75,15 @@ func (t *HelmTemplate) template(dirPath string, input io.Reader) exec.CmdRunResu
...
+	}
+	if t.opts.KubernetesAPIs != nil {
+		args = append(args, []string{"--api-versions", strings.Join(t.valuesFactory.additionaDownwardAPIValues.KubernetesAPIs, ",")}...)
+	}
+
 	{ // Add values files
-		vals := Values{t.opts.ValuesFrom, t.appContext, t.coreClient}
+		vals := t.valuesFactory.NewValues(t.opts.ValuesFrom, t.appContext)
```
joaopapereira: Move this line to line 77 to accommodate my previous comment

pkg/template/values.go
```diff
@@ -18,8 +18,9 @@ import (
...
 	ValuesFrom []v1alpha1.AppTemplateValuesSource
 
-	appContext AppContext
-	coreClient kubernetes.Interface
+	appContext                  AppContext
+	coreClient                  kubernetes.Interface
+	additionalDownwardAPIValues AdditionalDownwardAPIValues
 }
 
 func (t Values) AsPaths(dirPath string) ([]string, func(), error) {
```
joaopapereira: Just a comment about this function and maybe not be actionable in this PR.
This signature feels strange, the main reason for that is the return of that `func()` that you need to read the code to understand that it is the removal of the created folders.
It is the responsibility of the creator of `dirPath` to delete that folder, while this function is only responsible for creating the folders it needs inside the `dirPath`.

joaopapereira (commented):
Added some comments

pkg/template/values_factory.go
neil-hickey: those are definitely words I must have written. They make no sense. LOL

pkg/componentInfo/component_info.go
neil-hickey: interesting! That means we would need to save this cache to the disk - for the usecase of restarts.

pkg/componentInfo/component_info.go
neil-hickey: yup, I couldn't really find one sadly. Even if we refactored it out, we still *need* the branching in both places (though maybe it could be a shared function? I dunno if that helps)

pkg/template/helm_template.go
neil-hickey: @joaopapereira I moved things back into the Values struct, lemme know if I misunderstood the ask or if it's better now. Honestly, I am fine with it either way, I guess this one is more intentional though

pkg/app/app_template.go
```diff
@@ -26,20 +26,40 @@ func (a *App) template(dirPath string) exec.CmdRunResult {
...
+		result.AttachErrorf("%s", fmt.Errorf("Unable to get kapp-controller version before templating"))
+		return result
+	}
+
+	k8sAPIs, err := a.compInfo.KubernetesAPIs()
+	if err != nil {
+		result.AttachErrorf("%s", fmt.Errorf("Unable to list all server apigroups/version before templating"))
+		return result
+	}
+
```
cppforlife: perf: we should not have to retrieve any of this info if it's not being consumed by values.

pkg/app/app_template.go
```diff
@@ -26,20 +26,40 @@ func (a *App) template(dirPath string) exec.CmdRunResult {
 	var result exec.CmdRunResult
 	var isStream bool
 
+	kubernetesVersion, err := a.compInfo.KubernetesVersion(a.app.Spec.ServiceAccountName, a.app.Spec.Cluster, &a.app.ObjectMeta)
+	if err != nil {
+		result.AttachErrorf("%s", fmt.Errorf("Unable to get kubernetes version before templating"))
```
cppforlife: should be including underlying err

pkg/packageinstall/packageinstall.go
```diff
@@ -248,12 +241,14 @@ func (pi *PackageInstallCR) kcVersionConstraintsSatisfied(pkg *datapkgingv1alpha
...
 	if err != nil {
-		pi.log.Error(err, "Unable to parse kapp-controller version", "version string", pi.controllerVersion)
 		return false
 	}
 
+	v.Pre = semver.PRVersion{}
+	v.Build = semver.BuildMeta{}
```
cppforlife, line 250: should this move into ComponentInfo? seems its more related to version parsing rather than package matching

pkg/apis/kappctrl/v1alpha1/types_template.go
```diff
@@ -53,6 +53,11 @@ type AppTemplateHelmTemplate struct {
 	Path string `json:"path,omitempty" protobuf:"bytes,3,opt,name=path"`
 	// One or more secrets, config maps, paths that provide values (optional)
 	ValuesFrom []AppTemplateValuesSource `json:"valuesFrom,omitempty" protobuf:"bytes,4,rep,name=valuesFrom"`
+	// Optional: Get Kubernetes version, defaults (empty) to retrieving the version from the cluster.
+	// Can be manually overridden to a value instead.
+	KubernetesVersion *Version `json:"kubernetesVersion,omitempty" protobuf:"bytes,5,opt,name=kubernetesVersion"`
+	// Optional: Use kubernetes group/versions resources available in the live cluster
+	KubernetesAPIs *Version `json:"kubernetesAPIs,omitempty" protobuf:"bytes,6,opt,name=kubernetesAPIs"`
```
cppforlife: seems like our test isnt catching that this shouldnt be a *Version.

pkg/apis/kappctrl/v1alpha1/types_template.go
```diff
@@ -68,16 +73,35 @@ type AppTemplateValuesSourceRef struct {
...
+	KappControllerVersion *Version `json:"kappControllerVersion,omitempty" protobuf:"bytes,4,opt,name=kappControllerVersion"`
+	// Optional: Get running KubernetesAPIs from cluster, defaults (empty) to retrieving the APIs from the cluster.
+	// Can be manually supplied instead, e.g ["group/version", "group2/version2"]
+	KubernetesAPIs *GroupVersion `json:"kubernetesAPIs,omitempty" protobuf:"bytes,5,opt,name=kubernetesAPIs"`
 }
 
 // +k8s:openapi-gen=true
-type AppTemplateValuesDownwardAPI struct {
-	Items []AppTemplateValuesDownwardAPIItem `json:"items,omitempty" protobuf:"bytes,1,opt,name=items"`
+type Version struct {
```
cppforlife, line 98: i would recommend renaming this to KubernetesVersion and KubernetesGroupVersions

pkg/app/app_template.go
neil-hickey: Hmm, that's an interesting ask. So the reason it's like it is currently is:

1. I wanted to call each of these APIs only once before every templating step. Mainly because we have no good way to cache the k8s version on an external cluster, nor the k8s apis. We could do as Joe suggested with a one minute cache. 
2. The calls would have to get moved down into each factory individually (i.e duplicate the information gathering of something that ought not to change between templating steps.

pkg/packageinstall/packageinstall.go
neil-hickey: The constraint matching is specific to packageinstall (i.e the choice to drop the pre and build fields of KC version), so I think it makes more sense here. The kapp-controller version should always be as it is in the cluster, regardless of how / why we manipulate it in kapp-controller.

pkg/apis/kappctrl/v1alpha1/types_template.go
neil-hickey: Why exactly? I think `Version` can be used for both KC and K8s version info. If we make it explicitly `KubernetesVersion` wouldn't it look like this?

```
kubernetesVersion:
  kubernetesVersion: "version"

kappControllerVersion:
  kappControllerVersion: "version"
```

Otherwise we have two structs `KappControllerVersion` and `KubernetesVersion` which both match on the json `version` which feels a little redundant, i.e.

```
kubernetesVersion:
  version: "version"

kappControllerVersion:
  version: "version"
```

pkg/app/app_template.go
neil-hickey: I went ahead and memoized the calls, I added a test which validates the calls are made made only once with multiple template steps. pkg/app/app_template_test.go

pkg/apis/kappctrl/v1alpha1/types_template.go
neil-hickey: I renamed GroupVersions to KubernetesGroupVersions but I left the other as Version as it made more sense to me to keep them together / and it's a key underneath `KubernetesVersion` so it felt fine to leave as is

pkg/packageinstall/packageinstall.go
```diff
@@ -301,7 +301,18 @@ func (pi *PackageInstallCR) referencedPkgVersion() (datapkgingv1alpha1.Package,
 
 	// we only need to populate the versionInfo we know that the packages have constraints that will require this info.
 	if requiresClusterVersion {
-		v, err := pi.compInfo.KubernetesVersion(pi.model.Spec.ServiceAccountName, pi.model.Spec.Cluster, &pi.model.ObjectMeta)
+		kubernetesVersion := func() (semver.Version, error) {
+			memoizedKubernetesVersion := semver.Version{}
```
joe-kimmel-vmw: here the naming suggests you're memoizing but the variable doesn't have scope outside the function, right?  am I missing how this would be persisted across function calls?   (vs. a couple files up where you declared the variables in the outer scope and then assigned to them in the func)

pkg/packageinstall/packageinstall.go
joe-kimmel-vmw: I understand the point that it's not entirely semantically in-scope for the reader of PackageInstall to know about the effects of resetting the members of the version, `v` struct.   However it's less obvious to me where to move that logic, and i think it'll be a similar number of lines of code,  just maybe a little faster to read packageinstall.go until your concern is this logic specifically.     To the extent the concern is just readability, it feels like a refactor or an explanatory comment would have similar benefits.

pkg/template/values.go
neil-hickey: Not going to action on this one in this PR

PR #850, opened 2022-08-29
kctrl: Tightning  up the pkg authoring e2e testcases

sethiyash:

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/.github/workflows/test-kctrl-gh.yml
+++ b/.github/workflows/test-kctrl-gh.yml
@@ -49,7 +49,7 @@ jobs:
         ./hack/deploy-test.sh
 
         export KCTRL_E2E_NAMESPACE=kctrl-test
-        export KCTRL_E2E_IMAGE="localhost:5000/test-repo/testimage-$GITHUB_RUN_ID"
+        export KCTRL_E2E_IMAGE="`ifconfig | grep inet | grep -E '\b10\.' | awk '{ print $2}'`:5000/test-repo/testimage-$GITHUB_RUN_ID"
         kubectl create ns $KCTRL_E2E_NAMESPACE
 
         cd cli
--- a/cli/test/e2e/package_authoring_e2e_test.go
+++ b/cli/test/e2e/package_authoring_e2e_test.go
@@ -21,6 +21,9 @@ const (
 type E2EAuthoringTestCase struct {
 	Name                    string
 	InitInteraction         Interaction
+	ExpectedPkgBuild        string
+	ExpectedPkgResource     string
+	ExpectedVendir          string
 	ExpectedPackage         string
 	ExpectedPackageMetadata string
 }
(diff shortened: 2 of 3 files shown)


.github/workflows/test-kctrl-gh.yml
```diff
@@ -49,7 +49,7 @@ jobs:
         ./hack/deploy-test.sh
 
         export KCTRL_E2E_NAMESPACE=kctrl-test
-        export KCTRL_E2E_IMAGE="localhost:5000/test-repo/testimage-$GITHUB_RUN_ID"
+        export KCTRL_E2E_IMAGE="`ifconfig | grep inet | grep -E '\b10\.' | awk '{ print $2}'`:5000/test-repo/testimage-$GITHUB_RUN_ID"
```
rohitagg2020, line 52: Here port is 5000 but below it is 8080, not sure what is happening here. I think it should be same. Also, do we really need to export the variable at two places?

sethiyash: Yeah that's something missed by me actually in local my 5000 port was occupied so testing it with 8080 and then later forgot to update it. Thanks for bringing this up.

> do we really need to export the variable at two places?

We can start using `cli/hack/test-all-local-registry.sh` script in Github action instead of `cli/hack/test-all.sh` to not to have them in 2 places is it something that you are suggesting?

100mik, line 52: What is the motivation behind doing this, won't the registry always run on localhost?

cli/test/e2e/package_authoring_e2e_test.go
```diff
@@ -15,26 +15,105 @@ import (
...
 const (
-	workingDir = "kcrl-test"
+	workingDir = "kctrl-test"
 )
 
-func TestPackageInitAndRelease(t *testing.T) {
-	env := BuildEnv(t)
-	logger := Logger{}
-	kappCtrl := Kctrl{t, env.Namespace, env.KctrlBinaryPath, logger}
+type E2EAuthoringTestCase struct {
```
100mik, line 21: Instead of having a separate list of outputs, how about we add a key for `ExpectedOutput` in `E2ETestAuthoringCase`, and store expected output in the test case itself.

.github/workflows/test-kctrl-gh.yml
sethiyash: So this we have done to eliminate the imgpkg circular registry error we are getting when push and pulling from the same registry and that goes away once we stop using localhost and start doing this.

cli/test/e2e/package_authoring_e2e_test.go
rohitagg2020: +1

sethiyash: Done

cli/test/e2e/package_authoring_e2e_test.go
```diff
@@ -163,89 +193,413 @@ spec:
...
+			go releaseInteraction.Run(promptOutput)
+
+			kappCtrl.RunWithOpts([]string{"pkg", "release", "--version", "1.0.0", "--tty=true", "--chdir", workingDir},
+				RunOpts{NoNamespace: true, StdinReader: promptOutput.StringReader(),
+					StdoutWriter: promptOutput.BufferedOutputWriter(), Interactive: true})
+
+			keysToBeIgnored := []string{"creationTimestamp:", "releasedAt:", "image"}
+
+			// Verify PackageMetadata artifact
+			out, err := readFile("./carvel-artifacts/packages/testpackage.corp.dev/metadata.yml")
```
100mik: nit: could we use variables/constants for stuff we are using repetitively?
Maybe variables like `packageDir` and `pkgName`

cli/test/e2e/package_authoring_e2e_test.go
```diff
@@ -163,89 +193,413 @@ spec:
...
+				RunOpts{StdinReader: promptOutput.StringReader(), StdoutWriter: promptOutput.BufferedOutputWriter()})
+
+			kappCtrl.RunWithOpts([]string{"pkg", "available", "list"},
+				RunOpts{StdinReader: promptOutput.StringReader(), StdoutWriter: promptOutput.BufferedOutputWriter()})
+
+			kappCtrl.RunWithOpts([]string{"pkg", "install", "-p", "testpackage.corp.dev", "-i", "test", "--version", "1.0.0"},
+				RunOpts{StdinReader: promptOutput.StringReader(), StdoutWriter: promptOutput.BufferedOutputWriter()})
+
+			// clean ns created for dynatrace in GitHub release flow
+			switch testcase.Name {
```
100mik: Let's do this in the deferred clean up function?

100mik (commented):
Some nitty stuff, looking good!

cli/test/e2e/package_authoring_e2e_test.go
```diff
@@ -163,89 +193,413 @@ spec:
...
+			}
+
+			defer cleanUpInstalledPkg()
+
+			switch testcase.Name {
+			case "Github Release Flow":
+				kubectl.RunWithOpts([]string{"create", "ns", "dynatrace"}, RunOpts{NoNamespace: true})
+			}
+
+			kappCli.RunWithOpts([]string{"deploy", "-a", "test-package", "-f", fmt.Sprintf("%s/carvel-artifacts/packages/testpackage.corp.dev", workingDir), "-c"},
```
100mik, line 593: This should use `pkgDir` as well. It will be resolved in #852 as it is adding tests for that particular case

100mik (approved):
LGTM!

PR #852, opened 2022-08-29
Ensure that  For option helm Chart from Git, pkg init sync's properly

### LINKED ISSUE #848: kctrl - When syncing for helm Chart from Git, `pkg init` is not syncing the helm charts.

rohitagg2020:
Vendir sync was not happening properly for helmChart from Git and Manifest from Git.

Fixes #848

```release-note
Fix(es) these issues -
Vendir sync was not happening properly for helmChart from Git and Manifest from Git.
In case there is mismatch between AppSpec in PackageBuild and Package, it ask for overwrite behaviour via prompt.
```

##### Review Checklist:

- [x] Relevant tests are added or updated. Will be added by another PR being worked upon.
  a link to that PR
  change

```docs

```


cli/pkg/kctrl/cmd/package/init/init.go
```diff
@@ -114,15 +115,17 @@ func (o *InitOptions) Run() error {
 
 	// TODO: @praveenrewar Remove the header text as we are only saving data to files and doesn't require user input
 	// hence shouldn't be called as Step 3/3
```
praveenrewar: Can we remove this TODO as well? It seems it was added for some line which has been removed now.

cli/pkg/kctrl/cmd/app/init/github.go
```diff
@@ -118,3 +122,30 @@ func (g *GithubStep) initializeContentWithGithubRelease() error {
 	g.vendirConfig.Directories[0].Contents = append(g.vendirConfig.Directories[0].Contents, vendirconf.DirectoryContents{})
 	return g.initializeGithubRelease()
 }
+
+func (g *GithubStep) getIncludedPaths() error {
```
praveenrewar, line 126: Except for the small variance in the informational text, I don't see any difference in this and the getIncludedPaths for git. Perhaps this will be covered in the refactoring?

cli/pkg/kctrl/cmd/app/init/init.go
```diff
@@ -222,12 +222,16 @@ func (c CreateStep) createAppFromAppBuild() kcv1alpha1.App {
 func (c CreateStep) configureExportSection() {
 	fetchSource := c.build.GetObjectMeta().Annotations[FetchContentAnnotationKey]
 	exportSection := *c.build.GetExport()
+	// TODO not handling the case: In case of pkg init rerun with FetchFromLocalDirectory,
```
praveenrewar: Are we saying we are not handling these scenarios now and we would want to handle these in future? If so, can maybe we can reword it a little?

cli/pkg/kctrl/cmd/app/init/vendir.go
```diff
@@ -58,32 +56,17 @@ func (v *VendirStep) Interact() error {
...
 		helmStep := NewHelmStep(v.ui, v.config)
 		return Run(helmStep)
 	case FetchFromGit:
 		gitStep := NewGitStep(v.ui, v.config)
-		err := Run(gitStep)
-		if err != nil {
-			return err
-		}
+		return Run(gitStep)
 	case FetchChartFromGit:
```
praveenrewar: These 2 cases seems similar. Maybe we can just use a comma, like `case FetchFromGit, FetchChartFromGit:`?

cli/pkg/kctrl/cmd/app/init/fetch.go
```diff
@@ -49,7 +49,11 @@ func (f *FetchStep) Interact() error {
 	}
 	isHelmTemplateExistInPreviousOption := f.helmTemplateExistInAppBuild()
 	previousFetchOptionSelected := GetFetchOptionFromVendir(vendirConfig, isHelmTemplateExistInPreviousOption)
-
+	if previousFetchOptionSelected == "MultipleFetchOptions" {
```
100mik: Should this not be a constant defined someplace?

cli/pkg/kctrl/cmd/app/init/template.go
```diff
@@ -64,12 +65,12 @@ func (t *TemplateStep) Interact() error {
 		appTemplates := []v1alpha1.AppTemplate{}
 
 		// Add helmTemplate
-		if fetchSource == FetchFromHelmRepo || fetchSource == FetchChartFromGit {
```
100mik, line 67: No difference but any reason for switching these up?

cli/pkg/kctrl/cmd/app/init/template.go
```diff
@@ -103,8 +104,29 @@ func (t *TemplateStep) Interact() error {
...
 
+func (t *TemplateStep) getHelmAppTemplate(fetchSource string) (v1alpha1.AppTemplate, error) {
+	var pathFromVendir string
+	if fetchSource == FetchChartFromGit {
+		vendirConf, err := ReadVendirConfig()
+		if err != nil {
+			return v1alpha1.AppTemplate{}, err
+		}
+		pathFromVendir = vendirConf.Directories[0].Contents[0].IncludePaths[0]
+		// TODO Create regex and match against that.
```
100mik: I think this is good enough if there are  no additional possibilities. Are we sure about this?

Is "/**/*/" valid? Should that be taken care of? (I would think not but just making sure)

cli/pkg/kctrl/cmd/package/init/init.go
```diff
@@ -254,22 +257,50 @@ func (o *InitOptions) updatePackageInstall(pkgInstall *pkgv1alpha1.PackageInstal
...
+			o.ui.PrintInformationalText("AppSpec section of Package(inside package-resources.yml) and " +
+				"PackageBuild(inside package-build.yml) is different. " +
+				"Either choose to overwrite the Package AppSpec or leave as it is.")
+			overrideOptions := []string{"Yes", "No"}
+			choiceOpts := ui.ChoiceOpts{
+				Label:   "Overwrite the Package AppSpec from PackageBuild",
+				Default: 1,
+				Choices: overrideOptions,
+			}
+			selectedIndex, err := o.ui.AskForChoice(choiceOpts)
```
100mik, line 279: Should this not be ask for confirmation rather than `AskForChoice`, since the options are yes/no

cli/pkg/kctrl/cmd/package/init/init.go
```diff
@@ -132,7 +135,7 @@ func (o *InitOptions) Run() error {
 	if err != nil {
 		return err
 	}
-
+	o.ui.PrintHeaderText("Output")
```
100mik, line 135: Was this being done someplace else before?

100mik (dismissed):
Some thoughts

cli/pkg/kctrl/cmd/package/init/init.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/github.go
rohitagg2020: After much discussion, we have decided to add this to the individual fetch option. Anyway, it will be covered in refactoring.

cli/pkg/kctrl/cmd/app/init/template.go
rohitagg2020: no reason.

cli/pkg/kctrl/cmd/app/init/template.go
rohitagg2020: I am removing all the trailing `/` before using TrimSuffix to handle n no. of trailing `/`. e.g. `/**/*////`,

rohitagg2020: Adopted

cli/pkg/kctrl/cmd/app/init/vendir.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/init/init.go
rohitagg2020: `AskForChoice` if you select `no`, it exits today.

cli/pkg/kctrl/cmd/app/init/fetch.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/init/init.go
100mik: I think it returns an error today I believe, and that error if returned causes the exit.
Instead we could respond differently if there is an error.

Just suggesting this because putting "y" or "n" might be intuitive for a few folks

cli/pkg/kctrl/cmd/package/init/init.go
rohitagg2020: Yup, just moving it down closer to the text

cli/pkg/kctrl/cmd/app/init/init.go
rohitagg2020: Reworded

cli/pkg/kctrl/cmd/package/init/init.go
rohitagg2020: Thing is non interactive flow, `AskForConfirmation` will always return true and we want it to be false. Also, with `AskForChoice`, user will be entering `0` or `1` rather than `Yes` or `No`. If you still think we should move to `AskForConfirmation`, we can rephrase the question and do the error handling for `false` use case.

100mik:
Looks like a test is failing, does not look like a flake 🤔 
Re-running it though

100mik:
Was a flake ^

cli/pkg/kctrl/cmd/app/init/fetch.go
```diff
@@ -21,6 +21,7 @@ const (
 	FetchFromHelmRepo       string = "Helm Chart from Helm Repository"
 	FetchChartFromGit       string = "Helm Chart from Git Repository"
 	FetchFromLocalDirectory string = "Local Directory"
+	MultipleFetchOptionsSelected string = "MultipleFetchOptionsSelected"
```
100mik, line 24: This section seems to produce a diff when we run fmt

100mik (commented):
Quick fix!

PR #859, opened 2022-09-07
introduce min app sync period

cppforlife:
- provides a way to tune default sync period and minimum sync period for apps

Fixes #

```release-note
- Add `appDefaultSyncPeriod and appMinimumSyncPeriod durations in kapp-controller config`
```

##### Review Checklist:

  a link to that PR
  change

```docs

```


pkg/app/app.go
```diff
@@ -32,6 +33,12 @@ type Hooks struct {
 	WatchChanges    func(func(v1alpha1.App), chan struct{}) error
 }
 
+// AppOpts keeps reconciliation options
+type AppOpts struct {
```
joe-kimmel-vmw: idk feelfree to do a revive nolint annotation

pkg/app/app.go
```diff
@@ -55,11 +63,11 @@ type App struct {
 
 func NewApp(app v1alpha1.App, hooks Hooks,
 	fetchFactory fetch.Factory, templateFactory template.Factory,
-	deployFactory deploy.Factory, log logr.Logger, appMetrics *metrics.AppMetrics, compInfo ComponentInfo) *App {
+	deployFactory deploy.Factory, log logr.Logger, opts AppOpts, appMetrics *metrics.AppMetrics, compInfo ComponentInfo) *App {
```
joe-kimmel-vmw: this reminds me of last time this function's arguments list got too long and i asked you how many is too many

pkg/app/app_factory.go
```diff
@@ -41,5 +41,8 @@ func (f *CRDAppFactory) NewCRDApp(app *kcv1alpha1.App, log logr.Logger) *CRDApp
 	fetchFactory := fetch.NewFactory(f.CoreClient, vendirOpts, f.CmdRunner)
 	templateFactory := template.NewFactory(f.CoreClient, fetchFactory, f.KbldAllowBuild, f.CmdRunner)
 	deployFactory := deploy.NewFactory(f.CoreClient, f.Kubeconf, f.KcConfig, f.CmdRunner, log)
-	return NewCRDApp(app, log, f.AppMetrics, f.AppClient, fetchFactory, templateFactory, deployFactory, f.CompInfo)
+	return NewCRDApp(app, log, f.AppMetrics, f.AppClient, fetchFactory, templateFactory, deployFactory, AppOpts{
+		DefaultSyncPeriod: f.KcConfig.AppDefaultSyncPeriod(),
```
joe-kimmel-vmw, line 46: just for readability, idk about putting an inline struct in the middle of the 8th out of 9 args

pkg/app/reconcile_timer.go
```diff
@@ -61,11 +72,14 @@ func (rt ReconcileTimer) IsReadyAt(timeAt time.Time) bool {
 }
 
 func (rt ReconcileTimer) syncPeriod() time.Duration {
-	const defaultSyncPeriod = 30 * time.Second
-	if sp := rt.app.Spec.SyncPeriod; sp != nil && sp.Duration > defaultSyncPeriod {
-		return sp.Duration
+	selectedDur := rt.opts.DefaultSyncPeriod
+	if sp := rt.app.Spec.SyncPeriod; sp != nil {
```
joe-kimmel-vmw, line 79: hey i thought Dmitriy told me one time that we don't like one liners like this, stylistically speaking

pkg/config/config.go
```diff
@@ -150,14 +160,35 @@ func (gc *Config) ShouldSkipTLSForAuthority(candidateAuthority string) bool {
...
 	// Eventually kapp can be smart enough to keep minimal number of app changes.
 	// Set default first so that it can be overridden by user provided options.
-	return append([]string{"--app-changes-max-to-keep=5"}, gc.kappDeployRawOptions...)
+	return append([]string{"--app-changes-max-to-keep=5"}, gc.data.kappDeployRawOptions...)
+}
+
+// AppDefaultSyncPeriod returns duration that is used by Apps
+// that do not explicitly specify sync period.
+func (gc *Config) AppDefaultSyncPeriod() time.Duration {
+	const lowestDefault = 30 * time.Second
```
joe-kimmel-vmw, line 176: honest golang question:  does it make a difference to define this outside of the function? i would expect this kind of constant at the top of a file, from other languages.  -  are we going to allocate/evaluate this expression once per invocation,  or is there an implicit `static const` equivalent here that will optimize it out?

pkg/config/config.go
```diff
@@ -150,14 +160,35 @@ func (gc *Config) ShouldSkipTLSForAuthority(candidateAuthority string) bool {
...
+		return gc.data.appDefaultSyncPeriod
+	}
+	return lowestDefault
+}
+
+// AppMinimumSyncPeriod returns duration that is used as a lowest
+// sync period App would use for reconciliation. This value
+// takes precedence over any sync period that is lower.
+func (gc *Config) AppMinimumSyncPeriod() time.Duration {
+	const lowestMin = 30 * time.Second
```
joe-kimmel-vmw, line 187: same question about constants and allocation, especially since for now they're the same value....

pkg/config/config.go
```diff
@@ -168,31 +199,48 @@ func (gc *Config) addSecretDataToConfig(secret *v1.Secret) error {
...
 		// Allowed flags will be verified before kapp is invoked within Kapp class.
 		// (See pkg/deploy/kapp_restrict.go).
-		gc.kappDeployRawOptions = opts
-	} else {
-		gc.kappDeployRawOptions = nil
+		data.kappDeployRawOptions = opts
 	}
 
-	gc.skipTLSVerify = data["dangerousSkipTLSVerify"]
+	gc.dataLock.Lock()
```
joe-kimmel-vmw, line 240: would it still be safer to have the `defer gc.datalock.unlock` call immediately after you get the lock?   I realize currently there's only a single assignment in between and it seems like either that would cause some deep panic, or not,  but is it more 'future-friendly' somehow?

joe-kimmel-vmw (commented):
buncha little things but seems fine

pkg/app/app.go
cppforlife: ill keep it as is unless you have a strong opinion

pkg/app/app.go
cppforlife: im ok with the current set since that is indeed all the dependencies App requires. may be in a separate PR id like to see if we can merge CRDApp and App together. that might simplify some of this.

pkg/app/app_factory.go
cppforlife: 👍

pkg/app/reconcile_timer.go
cppforlife: generally for errors, this one was already there and it seems to scope variable.

pkg/config/config.go
cppforlife: > does it make a difference to define this outside of the function?

general recommendation i push for is to limit any code to their absolutely smallest scope required (lang independent recommendation). pulling out consts used in small scope just creates extra indirection (which is a burden to follow through). in a perfect world most variables are consts since majority do not have change, but go does not have such level of consts.

cppforlife: >  are we going to allocate/evaluate this expression once per invocation

that's up to go compiler to decide. the perf implication here is "extra" negligible.

pkg/config/config.go
cppforlife: they are superficially same (same value), but conceptually they are independent.

pkg/config/config.go
cppforlife: sure can change back (dont have strong opinion). commit that introduced the change explicitly forces us to change whole struct as a value, rather than individual things so in theory this line should not change for a long time.

cppforlife: done

pkg/app/app_factory.go
cppforlife: done

joe-kimmel-vmw (approved):

PR #862, opened 2022-09-08
never report the kubernetes version with the pre or buildmeta

joe-kimmel-vmw:


pkg/componentinfo/component_info.go
```diff
@@ -65,7 +65,13 @@ func (ci *ComponentInfo) KubernetesVersion(serviceAccountName string, specCluste
 		if err != nil {
 			return semver.Version{}, err
 		}
-		return semver.ParseTolerant(v.String())
+		retv, err := semver.ParseTolerant(v.String())
+		if err != nil {
+			return retv, err
+		}
+		retv.Pre = semver.PRVersion{}
+		retv.Build = semver.BuildMeta{}
```
cppforlife: gotta do it for service account case as well

pkg/componentinfo/component_info.go
```diff
@@ -38,13 +38,25 @@ func (ci *ComponentInfo) KappControllerVersion() (semver.Version, error) {
 // KubernetesVersion returns the running K8s version depending on AppSpec
 // If AppSpec points to external cluster, we use that k8s version instead
 func (ci *ComponentInfo) KubernetesVersion(serviceAccountName string, specCluster *v1alpha1.AppCluster, objMeta *metav1.ObjectMeta) (semver.Version, error) {
+
+	// parses and then scrubs Pre and Build from the version
+	parseAndScrub := func(s string) (semver.Version, error) {
```
cppforlife: nit: more conventionally this function would be turned into a method on ComponentInfo. i would also include "version" in the same so that it's cleat what it is doing => 

```
func (*ComponentInfo) parseAndScrubVersion(...
```

PR #864, opened 2022-09-09
add package details to child app cr annotations so that they can be used in app cr downward api

cppforlife:
change: add package details to child app cr annotations so that they can be used in app cr downward api

this info might be useful for template authors to propagate to some resources on the cluster. additionally it might be useful information to k8s operators that may decide certain functionality is available based on a version (e.g. imagine an operator that fails an upgrade of an operator before all of its CRs are on supported versions.)

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```


joe-kimmel-vmw (commented):
please make a corresponding docs change - this is going to be very hard for anyone to "discover" if we don't tell them it's there.

Issue #878, opened 2022-09-15
kapp-controller Pod stuck in CLBO and will never recover once the kapp-controller-sidecarexec container is SIGKILL’ed


scarlettqu:
**What steps did you take:**
Over the uptime, noticed kapp-controller-sidecarexec is in CLBO due to /etc/kappctrl-mem-tmp/sidecarexec.sock: bind: address already in use.
I was able to reproduce the issue by sending a SIGKILL signal to the container using `kill -9 <container pid>`.

**What happened:**
The Pod stuck in CLBO and the package resource is not recognized.
```
root@422f0e147ceebf2c4b41fd0b86093bc2 [ ~ ]# kubectl get po -n vmware-system-appplatform-operator-system
NAME                                                    READY   STATUS             RESTARTS       AGE
kapp-controller-786f8cf789-pv5m7                        1/2     CrashLoopBackOff   3 (28s ago)    32m

root@422f0e147ceebf2c4b41fd0b86093bc2 [ ~ ]# kubectl logs -n vmware-system-appplatform-operator-system kapp-controller-786f8cf789-pv5m7 kapp-controller-sidecarexec
{"level":"info","ts":1663028728.5450194,"logger":"kc-sidecarexec","msg":"start sidecarexec","version":"0.38.4"}
{"level":"info","ts":1663028728.5464313,"logger":"kc-sidecarexec","msg":"starting zombie reaper"}
{"level":"error","ts":1663028728.5473244,"logger":"kc-sidecarexec","msg":"Serving RPC","error":"Listening RPC: listen unix /etc/kappctrl-mem-tmp/sidecarexec.sock: bind: address already in use","stacktrace":"runtime.main\n\truntime/proc.go:250"}

root@422f0e147ceebf2c4b41fd0b86093bc2 [ ~ ]# kubectl get packages -A
error: the server doesn't have a resource type "packages"
```

**What did you expect:**
I expect the kapp-controller Pod can auto recover without any manual steps.

**Anything else you would like to add:**
1. Initial analysis: the issue comes from emptyDir: kapp-controller-sidecarexec is using emptyDir Volume for the /etc/kappctrl-mem-tmp/sidecarexec.sock. Since emptyDir persists among container crashes, the socket will not be released after the crash. Then the socket cannot bind after kapp-controller-sidecarexec restarts, which cause the CLBO.
[...]

PR #881, opened 2022-09-15
clean up sidecarexec socket file in case of previous unclean process termination

cppforlife:

addresses https://github.com/vmware-tanzu/carvel-kapp-controller/issues/878

no

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/sidecarexec/client.go
+++ b/pkg/sidecarexec/client.go
@@ -5,6 +5,7 @@ package sidecarexec
 
 import (
 	"net/rpc"
+	"sync"
 
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/exec"
 )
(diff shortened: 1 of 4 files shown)


pkg/sidecarexec/client.go
```diff
@@ -33,3 +30,53 @@ func (r Client) CmdExec() CmdExecClient {
...
+
+		err = refreshedClient.Call(serviceMethod, args, reply)
+	}
+	return err
+}
+
+// connectedClient takes a client which encountered a connection error
+// so that we only try to reconnect if we haven't already reconnected
+// due to another concurrent call.
+func (c *reconnectingRPCClient) connectedClient(disconnectedClient *rpc.Client) (*rpc.Client, error) {
```
joe-kimmel-vmw: nit re naming:  i'm trying to make this function a verb.  is it just `connect` ?   

Additional nit:  maybe the comment above would be more clear if you move the thread-safety to the front, e.g. 

``` 
// connect a client (which is nil or disconnected) in a thread-safe manner. 
  This is intended for clients which encountered a connection error, so that we only try to reconnect if we haven't already done so in a different thread.
func (c *rRPCC) connect(disconnectedClient *rpc.Client) .... 
```
 ?


or even could use technical language like `...concurrency-aware, reconnect a disconnected client at most once...`

cppforlife: done

joe-kimmel-vmw (approved):

PR #882, opened 2022-09-16
configurable tls cipher suites

joe-kimmel-vmw:
this change allows users of kapp-controller to customize the cryptographic algorithms that the API server will use for TLS. 

While the APIServer is by default only accessible within the cluster, users may still wish to customize this list especially if they expose the APIServer.

Manual tests show that if bad cipher suites are provided the pod promptly enters crashloopbackoff with an error message in the logs like:
```
{"level":"error","ts":1663351527.0906084,"logger":"kc.main","msg":"Exited run with error","error":"unable to parse TLSCipherSuites: Cipher suite FOO not supported or doesn't exist","stacktrace":"main.main\n\tgithub.com/vmware-tanzu/carvel-kapp-controller/cmd/controller/main.go:49\nruntime.main\n\truntime/proc.go:250"}
```

Additional manual tests showed that the default behavior is unchanged, but when a select set of cipher suites are applied then handshakes with other cipher algorithms cannot be completed.

### DIFF
--- a/cmd/controller/run.go
+++ b/cmd/controller/run.go
@@ -10,6 +10,7 @@ import (
 	_ "net/http/pprof" // Pprof related
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -28,6 +29,7 @@ import (
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/sidecarexec"
 	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp" // Initialize gcp client auth plugin
+	"k8s.io/component-base/cli/flag"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -49,6 +51,7 @@ type Options struct {
 	MetricsBindAddress     string
 	APIPriorityAndFairness bool
 	StartAPIServer         bool
+	TLSCipherSuites        string
 }
 
 // Based on https://github.com/kubernetes-sigs/controller-runtime/blob/8f633b179e1c704a6e40440b528252f147a3362a/examples/builtins/main.go
@@ -115,11 +118,18 @@ func Run(opts Options, runLog logr.Logger) error {
 		if err != nil {
 			return fmt.Errorf("Building pkg kappctrl client: %s", err)
 		}
+
+		cSuites, err := parseTLSCipherSuites(opts.TLSCipherSuites)
+		if err != nil {
+			return err
+		}
+
 		server, err := apiserver.NewAPIServer(pkgRestConfig, coreClient, pkgKcClient, apiserver.NewAPIServerOpts{
 			GlobalNamespace:              opts.PackagingGloablNS,
 			BindPort:                     bindPort,
 			EnableAPIPriorityAndFairness: opts.APIPriorityAndFairness,
 			Logger:                       runLog.WithName("apiserver"),
+			TLSCipherSuites:              cSuites,
 		})
 		if err != nil {
 			return fmt.Errorf("Building API server: %s", err)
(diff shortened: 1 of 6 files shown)


pkg/apiserver/apiserver.go
```diff
@@ -85,6 +85,12 @@ type NewAPIServerOpts struct {
 	// v1.19 and earlier clusters - our libraries use the beta version of those APIs but they used to be alpha.
 	EnableAPIPriorityAndFairness bool
 
+	// TlsCipherSuites is the kind of thing you wish you never had to think about but apparently there are wrong answers to everything
+	//                 in this life and the wrong answer to this question gets you flagged on somebody's compliance scan.
+	//                 I mean seriously I don't want to mislead you that I'm some kind of security expert. it's a list of acceptable Cipher algorithms I guess according to probably some bank's definition of acceptable.
```
joe-kimmel-vmw: (i'll clean these up)

cmd/controller/run.go
```diff
@@ -49,6 +51,7 @@ type Options struct {
 	MetricsBindAddress     string
 	APIPriorityAndFairness bool
 	StartAPIServer         bool
+	TlsCipherSuites        string
```
cppforlife: how can this field be configured by the user?

cmd/controller/run.go
```diff
@@ -266,3 +276,21 @@ func Run(opts Options, runLog logr.Logger) error {
 
 	return nil
 }
+
+// parseTlsCipherSuites tries to validate and return the user-input ciphers or returns a default list
+// defaults stolen from: https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/antrea/1.5.3/bundle/config/values.yaml#L48
```
aaronshurley: I have a slight maintenance concern about hardcoding the defaults. Primarily, how will we remember to update them?

A couple of mitigations/options that come to mind:
1. Keep the hardcoding
    - add a "release step" as a reminder to check these
    - or some other reminder (e.g. calendar/slack reminder, "when we bump go" reminder, etc.)
2. Remove the hardcoding and try to pull a "safe" list from somewhere. The best option that comes to mind for me is the CipherSuites function (https://go.dev/src/crypto/tls/cipher_suites.go). I think this would help keep the ciphers updated as we bump our deps.

This is a non-blocking comment. I'll leave it to y'all to figure out what you're comfortable with.

joe-kimmel-vmw: cool - i had a related conversation with DK in slack where he recommended keeping golang's defaults via `""` . 

one thing this list allowed me to do was to verify that it's working as desired without doing all the plumbing yet (by deploying the code, and then at the terminal trying openssl with an excluded cipher and seeing a handshake failure).   


I'm happy to leave this list behind as the 'prototype' phase ends

cmd/controller/run.go
```diff
@@ -266,3 +276,20 @@ func Run(opts Options, runLog logr.Logger) error {
 
 	return nil
 }
+
+// parseTLSCipherSuites tries to validate and return the user-input ciphers or returns a default list
+// implementation largely stolen from: https://github.com/antrea-io/antrea/blob/25ff93d8987c6b9e3a2062254da6d7d70c623410/pkg/util/cipher/cipher.go#L32
+func parseTLSCipherSuites(opts string) ([]string, error) {
+	csStrList := strings.Split(strings.ReplaceAll(opts, " ", ""), ",")
+	if len(csStrList) == 0 || (len(csStrList) == 1 && csStrList[0] == "") {
+		return []string{}, nil //{"TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256", "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256", "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384", "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384", "TLS_RSA_WITH_AES_256_GCM_SHA384"}, nil
```
cppforlife: nit: remove this comment as it's not clear what it's showing

cppforlife: lets also return nil instead of []string{} to be a bit more conventional.

cmd/controller/run.go
```diff
@@ -266,3 +276,20 @@ func Run(opts Options, runLog logr.Logger) error {
 
 	return nil
 }
+
+// parseTLSCipherSuites tries to validate and return the user-input ciphers or returns a default list
+// implementation largely stolen from: https://github.com/antrea-io/antrea/blob/25ff93d8987c6b9e3a2062254da6d7d70c623410/pkg/util/cipher/cipher.go#L32
+func parseTLSCipherSuites(opts string) ([]string, error) {
+	csStrList := strings.Split(strings.ReplaceAll(opts, " ", ""), ",")
+	if len(csStrList) == 0 || (len(csStrList) == 1 && csStrList[0] == "") {
```
cppforlife: no need to check `len(csStrList) == 0 ||`. Split will always returns array with at least one item (upstream code does not have it)

cmd/controller/run.go
```diff
@@ -266,3 +276,20 @@ func Run(opts Options, runLog logr.Logger) error {
...
+		return []string{}, nil //{"TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256", "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256", "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384", "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384", "TLS_RSA_WITH_AES_256_GCM_SHA384"}, nil
+	}
+
+	// check to make sure they all parse - this just a fail-fast
+	_, err := flag.TLSCipherSuites(csStrList)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse TLSCipherSuites: %v", err)
```
cppforlife: nit: %s instead of %v

Issue #874, opened 2022-09-14
Change tag in Package-repositry.yml file created by pkg repo release cmd to human readable tag.


rohitagg2020:
**Describe the problem/challenge you have**
Package-repository.yml file created by the pkg repo release command has sha digest as a tag for the pkg repo bundle. When we `kapp deploy` or `kubectl apply` this file and list the available repository on the cluster, we see the tag as `sha-digest` rather than human readable tag.

**Describe the solution you'd like**
We would like to see the sha-digest tag to be replaced with version provided while running `pkg repo release` command. Also, if we really dont want to miss out on the sha-digest, may be that can be added as metadata.annotation on the PackageRepository resource.

**Anything else you would like to add:**
[Additional information that will assist in solving the issue.]

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

PR #886, opened 2022-09-19
Update image with ver instead of sha in package-repository.yml

rohitagg2020:
Update image with ver instead of sha in package-repository.yml

Package-repository.yml file created by the pkg repo release command has sha digest as a tag for the pkg repo bundle. When we kapp deploy or kubectl apply this file and list the available repository on the cluster, we see the tag as sha-digest rather than human readable tag.

This PR adds the version as the tag rather than `sha-digest`. Also, it adds the sha-digest as metadata annotation.

#874

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/cli/pkg/kctrl/cmd/package/repository/release/artifact_writer.go
+++ b/cli/pkg/kctrl/cmd/package/repository/release/artifact_writer.go
@@ -22,7 +22,7 @@ func NewArtifactWriter(pkgRepoName string, directory string) *ArtifactWriter {
 	return &ArtifactWriter{PackageRepoName: pkgRepoName, TargetDir: directory}
 }
 
-func (w *ArtifactWriter) WritePackageRepositoryFile(imgpkgBundleLocation string) error {
+func (w *ArtifactWriter) WritePackageRepositoryFile(bundleURLWithSHARef, bundleURLWithVerRef string) error {
 	packageRepository := v1alpha1.PackageRepository{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "PackageRepository",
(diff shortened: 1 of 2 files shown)


cli/pkg/kctrl/cmd/package/repository/release/artifact_writer.go
```diff
@@ -31,11 +31,14 @@ func (w *ArtifactWriter) WritePackageRepositoryFile(imgpkgBundleLocation string)
 		ObjectMeta: metav1.ObjectMeta{
 			Name:              w.PackageRepoName,
 			CreationTimestamp: metav1.NewTime(time.Now()),
+			Annotations: map[string]string{
+				"sha-ref": bundleURLWithSHARef,
```
rohitagg2020, line 35: 2 questions:
1. Should we add this annotation? 
2. If yes, what should be the name?

cli/pkg/kctrl/cmd/package/repository/release/release.go
```diff
@@ -165,14 +165,16 @@ func (o *ReleaseOptions) Run() error {
...
-		bundleURL, err = imgpkgRunner.Run()
+		bundleURLWithSHARef, err = imgpkgRunner.Run()
 		if err != nil {
 			return err
 		}
 	}
 
 	artifactWriter := NewArtifactWriter(pkgRepoName, wd)
-	err = artifactWriter.WritePackageRepositoryFile(bundleURL)
+	repoURL := strings.Split(bundleURLWithSHARef, "@sha")[0]
```
rohitagg2020, line 175: Should we separate it on the basis of `@` or `@sha`. I think `@` is not a valid character for repo name. But not able to find concrete documentation

cli/pkg/kctrl/cmd/package/repository/release/artifact_writer.go
cppforlife: question 3, should this we flipped? (ie image uses sha reference and tag is in annotations), and if so, should we augment kctrl to display annotation value as an additional output in the cli?

rohitagg2020: @cppforlife I thought about it. Here are my observations:
1. If users are adding the pkg repo with `kctrl pkg repo add -r ... --url ... -n ....`, or the annotation is not present, for them, kctrl will not show anything in the `list` or `get` command.
2. Only when the annotation is provided, we will add the value.
Sample output:
```
./kctrl pkg repo list -A
Target cluster 'https://192.168.64.108:8443' (nodes: minikube)

Repositories in all namespaces

Namespace   Name   Source                                                                     Status
kctrl-test  demo   (imgpkg) rohitagg2020/demo-repo:1.0.0                                     Reconcile succeeded
^           demo2  (imgpkg) rohitagg2020/demo-repo@sha256-8a55a777f43f6755fbe782f7b45...     Reconcile succeeded
^           demo3  (imgpkg) (1.0.0) rohitagg2020/demo-repo@sha256-8a55a777f43f6755...        Reconcile succeeded
```

What are your thoughts?

PR #904, opened 2022-09-27
kctrl: Generating openapi schema from helm values file.

### LINKED ISSUE #899: kctrl: Generate the openapi schema for helm chart.
Today, when kctrl is used to package helm chart, it is not generating the openapi schema by reading values.yml. 

I would like kctrl to generate the openapi schema and be available as part of releasable package.

**Task**:
- [x] Proposal - https://hackmd.io/b436sCy1RnqTkSyjnzGwLg
- [ ] Add flag to disable the generation of openapi schema. FlagName: `generate-openapi-schema`
- [ ] Generate openAPI schema without description
- [ ] Generate openAPI schema with description by translating comments with the below formats. Also, all the comments have to be above the key name without any newline in-between.
- [ ] - "# <KEYNAME> Descritption"
- [ ] - "##  Descritption"
- [ ] - "# Descritption"
- [ ] Write unit test cases.
- [ ] Verify output with different scenarios - mongoDb enterprise operator.
       - Behavior in case of scalar node, if Value is empty.
- [ ] Update `TestE2EInitAndReleaseCases` e2e test cases.
- [ ] Update documentation w.r.t. flag.

---
Vote on this request

rohitagg2020:
kctrl: Generating openapi schema from helm values file.

Fixes #899

```release-note

```

##### Review Checklist:

  a link to that PR vmware-tanzu/carvel#568
  change

```docs

```


cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,373 @@
...
+	kcdatav1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/apis/datapackaging/v1alpha1"
+	yaml2 "gopkg.in/yaml.v2"
+	yaml3 "gopkg.in/yaml.v3"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/yaml"
+)
+
+// keys used when generating an OpenAPI Document
+const (
+	typeProp        = "type"
```
praveenrewar: Maybe use _Key_ instead of _Prop_? (For example, `typeKey`, `formatKey`, etc,.)

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,373 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &kcdatav1alpha1.ValuesSchema{
+		OpenAPIv3: runtime.RawExtension{Raw: jsonEncodedBytes},
+	}, nil
+}
+
+func (h HelmValuesSchemaGen) readValuesFile() ([]byte, error) {
+	helmChartDir := h.dirPath
```
praveenrewar: I think copying the dirPath is not required here. I noticed that you have used `helmChartDir` in some places and `h.dirPath` other places but couldn't figure out if there's a difference in those.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,373 @@
...
+	fileInfo, err := os.Stat(helmChartDir)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read %s: %s", h.dirPath, err.Error())
```
praveenrewar: os.Stat() always returns file path, so I think we can exclude it while wrapping the error. Also, `unable to read` might not be the best context for this error (example, it could be a not found error).
```suggestion
		return nil, fmt.Errorf("stat: %w", err.Error())
```

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+	if !fileInfo.IsDir() {
+		return nil, fmt.Errorf("expected %s to be directory", h.dirPath)
+	}
+
+	_, err = os.Stat(filepath.Join(helmChartDir, "values.yaml"))
```
praveenrewar: If os.ReadFile() also returns the os.IsNotExist() type error then I think we can directly use that instead of stat.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,373 @@
...
+		// It is possible that helm chart doesn't have values.yml file.
+		if os.IsNotExist(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	return os.ReadFile(filepath.Join(helmChartDir, "values.yaml"))
+}
+
+func (h HelmValuesSchemaGen) fileExists(filename string) bool {
```
praveenrewar: We don't seem to be using this anywhere, maybe you forgot to remove it?

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,373 @@
...
+				v := value.Content[i+1]
+				calculatedProperties, err := h.calculateProperties(k, v)
+				if err != nil {
+					return nil, err
+				}
+				valueItems := calculatedProperties.Items[0]
+				properties = append(properties, &MapItem{Key: valueItems.Key, Value: valueItems.Value})
+			}
+			apiKeys = append(apiKeys, &MapItem{Key: propertiesProp, Value: &Map{Items: properties}})
+		}
```
praveenrewar:
```suggestion
                var properties openAPIKeys
		for i := 0; i < len(value.Content); i += 2 {
			k := value.Content[i]
			v := value.Content[i+1]
			calculatedProperties, err := h.calculateProperties(k, v)
			if err != nil {
				return nil, err
			}
			valueItems := calculatedProperties.Items[0]
			properties = append(properties, &MapItem{Key: valueItems.Key, Value: valueItems.Value})
		}
		if len(value.Content) == 0 {
			apiKeys = append(apiKeys, &MapItem{Key: defaultProp, Value: "{}"})
		} else {
			apiKeys = append(apiKeys, &MapItem{Key: propertiesProp, Value: &Map{Items: properties}})
		}
```

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,373 @@
...
+
+		sort.Sort(apiKeys)
+		if key == nil {
+			return &Map{Items: apiKeys}, nil
+		}
+		mi := &MapItem{Key: key.Value, Value: &Map{Items: apiKeys}}
+		return &Map{Items: []*MapItem{mi}}, nil
```
praveenrewar: inline?

praveenrewar (commented):
Still going through it, just thought of hitting the submit button before continuing.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: In openapi Spec, they are called properties as well. Went with that convention.
If you still think to rename it as `key`, I am open to that.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Adopted. removed `helmChartDir` and replaced with `h.dirPath`.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Seems like os.stat error is pretty self explanatory. We need not wrap it or do anything. 
Sample error:
`stat /abc: no such file or directory`

rohitagg2020: Adopted, just returning the same error now.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Yes, Adopted

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: I prefer it like above, but If you still think we should do like below, I am open to do that.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Adopted.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
praveenrewar: Just having Prop doesn't seem to be that readable, and if we use the complete name _Property_ then one of the variable names would be `propertiesProperty` which seems a bit weird. So IMO _Key_ made more sense and it's also inline with your comment (`keys used when generating ...`) and other usage of them (for example `MapItem{Key: defaultKey}`)

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
praveenrewar: With the second approach we are relying on the condition in the for loop itself to keep the normal code path at minimal indentation (which is in general more readable)

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,362 @@
...
+	kcdatav1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/apis/datapackaging/v1alpha1"
+	yaml2 "gopkg.in/yaml.v2"
+	yaml3 "gopkg.in/yaml.v3"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/yaml"
+)
+
+// keys used when generating an OpenAPI Document
+const (
+	typeKey         = "type"
```
praveenrewar: nit: Missing proper indentation (running `hack/build.sh` should resolve it)

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,362 @@
...
+	formatKey      = "format"
+	descriptionKey = "description"
+	itemsKey      = "items"
+	propertiesKey = "properties"
+	defaultKey    = "default"
+)
+
+var propOrder = map[string]int{
```
praveenrewar: maybe `keyOrder` or `propertyOrder`?

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,362 @@
...
+	defaultKey:     6,
+}
+
+const (
+	objectVal = "object"
+	arrayVal  = "array"
+	floatVal  = "float"
+)
+
+// Yaml tags that differentiate the type of scalar object in the node
```
praveenrewar: nit: Yaml -> YAML (since it's an acronym)

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,362 @@
...
+	return len(o)
+}
+
+func (o openAPIKeys) Less(i, j int) bool {
+	return propOrder[o[i].Key.(string)] < propOrder[o[j].Key.(string)]
+}
+
+func (o openAPIKeys) Swap(i, j int) {
+	o[i], o[j] = o[j], o[i]
+}
```
praveenrewar: I think we can remove these.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,362 @@
...
+			OpenAPIv3: runtime.RawExtension{Raw: nil},
+		}, nil
+	}
+	if document.Kind != yaml3.DocumentNode {
+		// return proper error message
+		return nil, fmt.Errorf("invalid node kind supplied: %d", document.Kind)
+	}
+	if document.Content[0].Kind != yaml3.MappingNode {
+		return nil, fmt.Errorf("values file must resolve to a map (was %d)", document.Content[0].Kind)
+	}
```
praveenrewar: I think these can be replaced with switch case.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,362 @@
...
+		var apiKeys openAPIKeys
+		apiKeys = append(apiKeys, &MapItem{Key: typeKey, Value: arrayVal})
+		description, isPresent := h.getDescriptionFromNode(key)
+		if isPresent {
+			apiKeys = append(apiKeys, &MapItem{Key: descriptionKey, Value: description})
+		}
+
+		if len(value.Content) != 0 {
+			properties := &Map{[]*MapItem{}}
+			for _, v := range value.Content {
```
praveenrewar: Similar to above, this for loop can be out of the if condition.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,362 @@
...
+					calculatedProperties, err := h.calculateProperties(nil, v)
+					if err != nil {
+						return nil, err
+					}
+					for _, item := range calculatedProperties.Items {
+						if item.Key == propertiesKey {
+							properties.Items = append(properties.Items, item.Value.(*Map).Items...)
+						}
+					}
+				case yaml3.SequenceNode:
```
praveenrewar: Since we are doing the same thing for SequenceNode and AliasNode, we could have them in single case statement
```suggestion
				case yaml3.SequenceNode, yaml3.AliasNode:
```

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,362 @@
...
+		}
+
+		valType := h.openAPITypeFor(value.Tag, value.Value)
+		defaultVal, err := h.getDefaultValue(valType, value.Value)
+		if err != nil {
+			return nil, err
+		}
+		apiKeys = append(apiKeys, &MapItem{Key: typeKey, Value: valType})
+		apiKeys = append(apiKeys, &MapItem{Key: defaultKey, Value: defaultVal})
+		if value.Tag == floatTag {
```
praveenrewar, line 214: Seems a bit weird to just check for the floatTag 🤔

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,362 @@
...
+	comment = strings.ReplaceAll(comment, "\n#", "")
+	if strings.HasPrefix(comment, fmt.Sprintf("# %s", node.Value)) {
+		return strings.TrimSpace(strings.TrimPrefix(comment, fmt.Sprintf("# %s", node.Value))), true
+	} else if strings.HasPrefix(comment, "# --") {
+		return strings.TrimSpace(strings.TrimPrefix(comment, "# --")), true
+	} else if strings.HasPrefix(comment, "##") {
+		return strings.TrimSpace(strings.TrimPrefix(comment, "##")), true
+	} else if strings.HasPrefix(comment, "#") {
+		return strings.TrimSpace(strings.TrimPrefix(comment, "#")), true
+	}
```
praveenrewar: Sounds like this can be replaced with a switch case?
```
switch {
case strings.HasPrefix(comment, fmt.Sprintf("# %s", node.Value)):
         // do something
...
}
```

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,362 @@
...
+	return "", false
+}
+
+func (h HelmValuesSchemaGen) openAPITypeFor(tag, value string) string {
```
praveenrewar: `For` at the end sounds like an incomplete name, were you trying to add something else as well at the end?

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,362 @@
...
+	case nullTag:
+		if value == "null" {
+			return "null"
+		}
+	}
+	return "string"
+
+}
+
+func (h HelmValuesSchemaGen) getDefaultValue(tag, value string) (interface{}, error) {
```
praveenrewar, line 270: `getDefaultValue` is always used in conjunction with `openAPITypeFor`. It seems we could directly `return strconv.ParseBool(value)` in case of `boolTag` and so on...

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen_test.go
```diff
@@ -0,0 +1,220 @@
...
+	return nil
+}
+
+func createFile(filePath string, fileData []byte) error {
+	err := os.WriteFile(filePath, fileData, fs.ModePerm)
+	if err != nil {
+		return fmt.Errorf("unable to create %s file: %s", fileData, err.Error())
+	}
+	return nil
+}
```
praveenrewar: nit: add newline at end of file

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen_test.go
```diff
@@ -0,0 +1,220 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package release
```
praveenrewar: should be under _test suffix.

cli/pkg/kctrl/cmd/package/release/release.go
```diff
@@ -159,6 +153,18 @@ func (o *ReleaseOptions) releaseResources(appSpec kcv1alpha1.AppSpec, pkgBuild c
 	return nil
 }
 
+func generateValuesSchema(pkgBuild cmdpkgbuild.PackageBuild) (*kcdatav1alpha1.ValuesSchema, error) {
+	var yttPaths []string
+	for _, templateStage := range pkgBuild.Spec.Template.Spec.App.Spec.Template {
+		if templateStage.HelmTemplate != nil {
+			return NewHelmValuesSchemaGen(templateStage.HelmTemplate.Path).Schema()
```
praveenrewar, line 169: What should happen if there are more templating step after helm?

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,362 @@
...
+// Yaml tags that differentiate the type of scalar object in the node
+const (
+	nullTag  = "!!null"
+	boolTag  = "!!bool"
+	intTag   = "!!int"
+	floatTag = "!!float"
+)
+
+type MapItem struct {
+	Key   interface{}
```
100mik: Won't the key here always be a string?
Looks like the sort function is dependent on that too.
Why cast there if we can use it as a string? I do not think we could have numeric keys at all

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,362 @@
...
+			OpenAPIv3: runtime.RawExtension{Raw: nil},
+		}, nil
+	}
+
+	var document yaml3.Node
+	err = yaml3.Unmarshal(fileData, &document)
+	if err != nil {
+		return nil, err
+	}
+	if document.Kind == 0 {
```
100mik: I am not sure if the value of `Kind` can be zero post marshalling (loss like valid values are powers of 2 [here](https://pkg.go.dev/gopkg.in/yaml.v3#Kind)

However, since we are doing nothing if that's the case, I would say we should do this if `IsZero()`
```suggestion
	if document.IsZero() {
```

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
100mik: +1

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: They are being used by sort internally

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,362 @@
...
+		// It is possible that helm chart doesn't have values.yml file.
+		if os.IsNotExist(err){
+			return nil, nil
+		}
+		return nil, err
+	}
+	return fileData, nil
+}
+
+func (h HelmValuesSchemaGen) convertToYAML(val interface{}) interface{} {
```
100mik: Are we sure this covers all the cases? 
Are we ensuring in `calculateProperties` that we won't end up with slices as values here? Does this still cover the case?

I think using a function exported by the YAML library would be safer?

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Since third one is on `document.Content[0].Kind`, I didn't do it.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,362 @@
...
+func (o openAPIKeys) Swap(i, j int) {
+	o[i], o[j] = o[j], o[i]
+}
+
+type HelmValuesSchemaGen struct {
+	// dirPath to the helm chart directory
+	dirPath string
+}
+
+func NewHelmValuesSchemaGen(path string) HelmValuesSchemaGen {
```
100mik, line 74: Will it help with test coverage if we make this section accept file data instead of file, and remove the file reading part from it's scope?

This might let us define better unit tests

100mik (commented):
Taking a closer look at how we are calculating properties, putting down thoughts so far

cli/pkg/kctrl/cmd/package/release/release.go
rohitagg2020: Any number of combinations are possible. But since we are saying that tool will handle only one fetch config, I dont think we can support complex use case.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
100mik: ^ Where is `Sort` being used?
Is it used internally by one of the libraries?

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: I will rename it to `openAPIType`. `For` was like openAPITypeFor this tag and value.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: in openapi Spec, for a float value, type is number but format is float. Even ytt does the same.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: What is being passed to `calculateProperties` is different in both the cases. Hence, they are separate

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Even if I move it out, there is still 1 if condition needs to be added. I think this way is cleaner.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen_test.go
rohitagg2020: if it is about filename, filename has test suffix`helm_openapi_schema_gen_test.go`

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: In that case, we have to read the fileData outside and pass here. I wanted to keep all the things at one place.

I have added the test case for empty file and file not exist anyway.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: > Are we ensuring in calculateProperties that we won't end up with slices as values here? Does this still cover the case?

Yes, all the slices values become properties. 

We cant use the library function as it is because that adds `key` and `value` literal while marshalling.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Nice, adopted

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
100mik: I noticed the unit tests later, I was wondering if unit tests should be writing files
I _think_ it would be better if it was all in memory.
Consider it a nit!

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: I looked at values, but forgot to look at key part. I think we can move it to string.

cli/pkg/kctrl/cmd/package/release/release.go
rohitagg2020: What are your thoughts?

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: its used in calculateproperties.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen_test.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
100mik: I think `mapToYAMLBytes` would be a better name

rohitagg2020: IMHO as it takes interface{} as an input, thats why I have named it generic. If you still think `mapToYAMLBytes` sounds better, I can make the change

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,361 @@
...
+	err = yaml3.Unmarshal(fileData, &document)
+	if err != nil {
+		return nil, err
+	}
+	if document.IsZero() {
+		return &kcdatav1alpha1.ValuesSchema{
+			OpenAPIv3: runtime.RawExtension{Raw: nil},
+		}, nil
+	}
+	if document.Kind != yaml3.DocumentNode {
```
praveenrewar, line 100: nit: add new line

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,361 @@
...
+		return nil, fmt.Errorf("invalid node kind supplied: %d", document.Kind)
+	}
+	if document.Content[0].Kind != yaml3.MappingNode {
+		return nil, fmt.Errorf("values file must resolve to a map (was %d)", document.Content[0].Kind)
+	}
+	openAPIProperties, err := h.calculateProperties(nil, document.Content[0])
```
praveenrewar, line 107: nit: new line

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
praveenrewar: > IMHO as it takes interface{} as an input, thats why I have named it generic

+1

How about `toYAML()`?

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,361 @@
...
+}
+
+func (h HelmValuesSchemaGen) getDescriptionFromNode(node *yaml3.Node) (string, bool) {
+	if node == nil || node.HeadComment == "" {
+		return "", false
+	}
+
+	comment := node.HeadComment
+	comment = strings.ReplaceAll(comment, "\n##", "")
+	comment = strings.ReplaceAll(comment, "\n#", "")
```
praveenrewar, line 239: I am a bit unclear about what we are doing over here 🤔

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen_test.go
```diff
@@ -0,0 +1,221 @@
...
+func createDir(dirName string) error {
+	err := os.Mkdir(dirName, fs.ModePerm)
+	if err != nil {
+		return fmt.Errorf("unable to create %s directory: %s", dirName, err.Error())
+	}
+	return nil
+}
+
+func createFile(filePath string, fileData []byte) error {
+	err := os.WriteFile(filePath, fileData, fs.ModePerm)
```
praveenrewar: I think we can directly use this function (Same for `Mkdir`).

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Sometimes in helm values.yaml, comment is divided into multi line, we are converting them to a single line

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,350 @@
...
+	Items []*MapItem
+}
+
+func (m *Map) Iterate(iterFunc func(k string, v interface{})) {
+	for _, item := range m.Items {
+		iterFunc(item.Key, item.Value)
+	}
+}
+
+type byKey []*MapItem
```
100mik: "byKey" does not seem to be very obvious?
Maybe `KeyValues` or `KeyValueSet` or `KeyValueList` will be more apt?
(Or maybe something else that is clear)

cli/pkg/kctrl/cmd/package/release/release.go
```diff
@@ -159,6 +153,18 @@ func (o *ReleaseOptions) releaseResources(appSpec kcv1alpha1.AppSpec, pkgBuild c
 	return nil
 }
 
+func generateValuesSchema(pkgBuild cmdpkgbuild.PackageBuild) (*kcdatav1alpha1.ValuesSchema, error) {
+	var yttPaths []string
+	for _, templateStage := range pkgBuild.Spec.Template.Spec.App.Spec.Template {
+		if templateStage.HelmTemplate != nil {
```
100mik: This logic relies on the user always using HelmTemplate first if they do, which is supposed to be the case

Not sure if we should make it a Bir generic

cli/pkg/kctrl/cmd/package/release/release.go
```diff
@@ -159,6 +153,18 @@ func (o *ReleaseOptions) releaseResources(appSpec kcv1alpha1.AppSpec, pkgBuild c
...
+			return NewHelmValuesSchemaGen(templateStage.HelmTemplate.Path).Schema()
+		} else if templateStage.Ytt != nil {
+			yttPaths = append(yttPaths, templateStage.Ytt.Paths...)
```
100mik: If the values are fed to only the first templating step even if it is YTT
Maybe we should stop doing this

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,350 @@
...
+				if err != nil {
+					return nil, err
+				}
+				for _, item := range calculatedProperties.Items {
+					if item.Key == itemsKey {
+						properties.Items = append(properties.Items, item.Value.(*Map).Items...)
+					}
+				}
+			default:
+				return nil, fmt.Errorf("Unrecognized type %T", v.Kind)
```
100mik: Is there any type other than this that we do not wanna deal with?
Looks like we are doing the same thing for all types other than `ScalarNode`, could be an `if` statement

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Removed it all together.

cli/pkg/kctrl/cmd/package/release/release.go
rohitagg2020: Not able to comprehend. You want to do something like this:

``` 
for index, templateStage := range pkgBuild.Spec.Template.Spec.App.Spec.Template {
		
               if templateStage.HelmTemplate != nil && index == 0 {

                ...
}```

File: cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go (hunk 33 above)
Comment (rohitagg2020) [reply cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go]: Ideally, there is no type. I have added this as we dont know yet how to deal with if new type comes in the future.

File: cli/pkg/kctrl/cmd/package/release/release.go (hunk 32 above)
Comment (rohitagg2020) [reply cli/pkg/kctrl/cmd/package/release/release.go]: Not able to understanfd. But looks like here, you are trying to collate all the folder paths for ytt.

File: cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen_test.go (hunk 29 above)
Comment (rohitagg2020) [reply cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen_test.go]: Resolved

File: cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go (hunk 25 above)
Comment (praveenrewar) [reply cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go:74]: > I was wondering if unit tests should be writing files

+1

File: cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go [hunk 34]
```diff
@@ -0,0 +1,280 @@
...
+type MapItem struct {
+	Key   string
+	Value interface{}
+}
+
+type Map struct {
+	Items []*MapItem
+}
+
+func (m *Map) Iterate(iterFunc func(k string, v interface{})) {
```
100mik: nit: Let's call this `ForEach()`

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,280 @@
...
+func (h HelmValuesSchemaGen) readValuesFile() ([]byte, error) {
+	fileInfo, err := os.Stat(h.dirPath)
+	if err != nil {
+		return nil, err
+	}
+	if !fileInfo.IsDir() {
+		return nil, fmt.Errorf("expected %s to be directory", h.dirPath)
+	}
+
+	fileData, err := os.ReadFile(filepath.Join(h.dirPath, "values.yaml"))
```
100mik, line 134: Should we also account for `values.yml` or is it always `values.yaml`?

100mik: We should probably see whether helm accepts both

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,280 @@
...
+
+// YAML tags that differentiate the type of scalar object in the node
+const (
+	nullTag  = "!!null"
+	boolTag  = "!!bool"
+	intTag   = "!!int"
+	floatTag = "!!float"
+)
+
+type MapItem struct {
```
100mik, line 54: `Key` for us will always be string
Is there any other reason to not use `yaml2.MapItem` directly?

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,280 @@
...
+			calculatedProperties, err := h.calculateProperties(value.Content[i], value.Content[i+1])
+			if err != nil {
+				return nil, err
+			}
+			properties = append(properties, &MapItem{Key: calculatedProperties.Items[0].Key, Value: calculatedProperties.Items[0].Value})
+		}
+		if len(properties) > 0 {
+			apiKeys = append(apiKeys, &MapItem{Key: propertiesKey, Value: &Map{Items: properties}})
+		} else {
+			apiKeys = append(apiKeys, &MapItem{Key: defaultKey, Value: "{}"})
```
100mik, line 183: We could move all default values (like `[]` or `{}`) into constants

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,280 @@
...
+
+		if len(value.Content) > 0 {
+			// For now, we have decided to add only the first item if it is an array.
+			// Based on feedback, we will revisit and enhance the implementation.
+			v := value.Content[0]
+			if len(v.Content) > 0 && v.Content[0].HeadComment == "" {
+				v.Content[0].HeadComment = v.HeadComment
+			}
+			val := v
+			if v.Kind == yaml3.AliasNode {
```
100mik: What does this check signify?

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,280 @@
...
+		apiKeys = append(apiKeys, &MapItem{Key: defaultKey, Value: []interface{}{}})
+
+		if len(value.Content) > 0 {
+			// For now, we have decided to add only the first item if it is an array.
+			// Based on feedback, we will revisit and enhance the implementation.
+			v := value.Content[0]
+			if len(v.Content) > 0 && v.Content[0].HeadComment == "" {
+				v.Content[0].HeadComment = v.HeadComment
+			}
+			val := v
```
100mik: nit: Do we need another copy of `v` here?

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,280 @@
...
+		} else {
+			apiKeys = append(apiKeys, &MapItem{Key: defaultKey, Value: "{}"})
+		}
+	case yaml3.SequenceNode:
+		apiKeys = append(apiKeys, &MapItem{Key: typeKey, Value: arrayVal})
+		apiKeys = append(apiKeys, &MapItem{Key: defaultKey, Value: []interface{}{}})
+
+		if len(value.Content) > 0 {
+			// For now, we have decided to add only the first item if it is an array.
+			// Based on feedback, we will revisit and enhance the implementation.
```
100mik: nit: Lets keep the comment concise,
`TODO: Do we need to consider that elements in a list might have different or more keys?`

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
```diff
@@ -0,0 +1,280 @@
...
+			properties = append(properties, &MapItem{Key: calculatedProperties.Items[0].Key, Value: calculatedProperties.Items[0].Value})
+		}
+		if len(properties) > 0 {
+			apiKeys = append(apiKeys, &MapItem{Key: propertiesKey, Value: &Map{Items: properties}})
+		} else {
+			apiKeys = append(apiKeys, &MapItem{Key: defaultKey, Value: "{}"})
+		}
+	case yaml3.SequenceNode:
+		apiKeys = append(apiKeys, &MapItem{Key: typeKey, Value: arrayVal})
+		apiKeys = append(apiKeys, &MapItem{Key: defaultKey, Value: []interface{}{}})
```
100mik, line 187: All other defaults seem to be string values, what is different here?

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Helm doesnt accept `values.yml`.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Adopted

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Not able to understand. In case of AliasNode, the value is defined in the `Alias` field rather than `Content`.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: I didnt want to reuse `v` as that looked confusing to me.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: It comes with quotes if we use it as string.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: As this is being used here only and normally we move things to constant only when they are being used at > 1 place.

cli/pkg/kctrl/cmd/package/release/helm_openapi_schema_gen.go
rohitagg2020: Apart from this, I have made it a struct so that we can add other information in the future if we want.

praveenrewar (approved):
LGTM!

100mik (approved):
Lookin' good!

PR #909, opened 2022-09-28
Activate caching of images and bundles

### LINKED ISSUE #664: Make kapp-controller more resilient to network failures while talking to the registry

With this feature, we want to try and address some of the following problems
- When an application is being reconciled, the lack of access to a registry, in some cases, should not make the reconciliation fail
- When an application is being reconciled, in some cases, it should not need to retrieve information from the registry in order to complete the reconciliation

kapp-controller tries to get all the running applications to a consistent state every 10 minutes, which means that every 10 minutes, `kapp-controller` would need to retrieve the configuration, and template and apply them to the cluster the configuration. Having to do this when you have 10 clusters with 100 applications in each, means that every 10 minutes, `kapp-controller` would have to reach out 1000 times to a registry.

This behavior can cause the registry to be overwhelmed with the burst of requests and randomly fail to reply, but this also forces the `kapp-controller` to rely on the network to ensure eventual consistency of the resources in the cluster. There are some cases where `kapp-controller` knows that the configuration cannot be changed, ex: when PackageCR is pointing to an OCI image that is referenced by its SHA.

A possible solution would be to implement features in `kapp-controller`, or the tools used by it, to understand if a particular piece of configuration can be reused after the initial retrieval and keep it, preventing reaching out to the registry to get it again.

For an MVP of this feature, we should consider that our target is configuration store in OCI images.
Cases where `kapp-controller` should keep the configuration
- PackageRepository Custom Resource is configured to retrieve the Packages from an OCI image while using the SHA of the image
- Package Custom Resource is configured to retrieve the configuration from an OCI image while using the SHA of the image
- App Custom Resource is configured to retrieve the configuration from an OCI image while using the SHA of the image
- PackageRepository Custom Resource is configured to retrieve the Packages from an imgpkg bundle while using the SHA of the image
   We should only consider fully relocated bundles
- Package Custom Resource is configured to retrieve the configuration from an imgpkg bundle  while using the SHA of the image
   We should only consider fully relocated bundles
- App Custom Resource is configured to retrieve the configuration from an imgpkg bundle while using the SHA of the image
   We should only consider fully relocated bundles

The rationale for only considering fully relocated bundles is the feature from `imgpkg` when it knows that all images are present in a particular repository it will update the ImagesLock, file in disk, to point to the images that are in this repository. This will allow Kubernetes to retrieve the images from a location that is nearer.

This issue is going to be used as the base issue that will be used to track the feature since this work might span multiple tools

Status of the feature:
- [x] https://github.com/vmware-tanzu/carvel-imgpkg/issues/390 
- [x] https://github.com/vmware-tanzu/carvel-vendir/issues/160
- [x] https://github.com/vmware-tanzu/carvel-kapp-controller/issues/688
- [x] https://github.com/vmware-tanzu/carvel-kapp-controller/issues/689

---
Vote on this request

### LINKED ISSUE #688: [vendir cache] Enable caching in vendir
Implementation of https://github.com/vmware-tanzu/carvel-kapp-controller/issues/664

Provide vendir with all the information needed to enable caching of images and bundles

kapp-controller will create a folder that will provide to vendir to store all the cached data as well as the maximum size of the contents that can be cacheable.

Implementation bits:
- kapp-controller deployment can contain an `emptyDir` that will be used to store all the cached content
- kapp-controller needs to create a folder in this directory per AppCR, to ensure other AppCR cannot use images they do not have access to.
- The recommendation for the folder name could be "namespace+secretName". If we do this may be different AppCR could reuse caching because they have the same credentials which means the same access level to the registry. Keep in mind that having the folders per secret if we remove the folder when a particular AppCR is deleted, this will cause other AppCRs to refetch the contents

Open questions:
- What should be the maximum size that kapp-controller should allow to be cached?

---
Vote on this request

### LINKED ISSUE #689: [vendir cache] Clear cache when AppCR is removed
Implementation of https://github.com/vmware-tanzu/carvel-kapp-controller/issues/664

When an AppCR is deleted, kapp-controller does not need to keep the caching information for that particular application

When kapp-controller is informed of the removal of an AppCR it should delete the folder containing the cache contents for that particular application.
More information on how to check which folder is used by which AppCR can be found in the implementation of #688

This is nice to have, but maybe not 100% required for MVP.

---
Vote on this request

We are also happy to receive and review Pull Requests if you want to help work on this issue.

joaopapereira:
In this PR `kapp-controller` will start to cache images and bundles. This feature will allow `kapp-controller` to reconcile pre-existing applications and package repositories even if the registry is no longer available.
This only happens when the images/bundles are referenced by SHA and in the case of bundles if all the images are colocated.

Fixes #664
Fixes #688
Fixes #689

```release-note
Allow `kapp-controller` to keep working even if the image registry is no longer available. This change also lowers the number of requests that `kapp-controller` needs to do to the registry, decreasing the load in these registries. 
```

##### Review Checklist:

  a link to that PR
  change

### DIFF
--- a/pkg/pkgrepository/app_fetch.go
+++ b/pkg/pkgrepository/app_fetch.go
@@ -39,7 +39,7 @@ func (a *App) fetch(dstPath string) (string, exec.CmdRunResult) {
 		return "", result
 	}
 
-	result = vendir.Run(conf, dstPath)
+	result = vendir.Run(conf, dstPath, a.cacheID())
 
 	// retry if error occurs before reporting failure.
 	// This is mainly done to support private registry
@@ -61,7 +61,7 @@ func (a *App) fetch(dstPath string) (string, exec.CmdRunResult) {
 				// no secrets/configmaps have changed, no point in retrying
 				continue
 			}
-			result = vendir.Run(newConf, dstPath)
+			result = vendir.Run(newConf, dstPath, a.cacheID())
 			if result.Error == nil {
 				break
 			}
@@ -77,4 +77,8 @@ func (a *App) fetch(dstPath string) (string, exec.CmdRunResult) {
 	}
 
 	return dstPath, result
+}
+
+func (a *App) cacheID() string {
+	return string(a.pkgRepoUID)
 }
--- a/test/e2e/kappcontroller/imgpkg_bundle_test.go
+++ b/test/e2e/kappcontroller/imgpkg_bundle_test.go
@@ -7,9 +7,11 @@ import (
 	"fmt"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
+	v1alpha12 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
 	"github.com/vmware-tanzu/carvel-kapp-controller/test/e2e"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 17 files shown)


pkg/pkgrepository/app_fetch.go
```diff
@@ -78,3 +78,7 @@ func (a *App) fetch(dstPath string) (string, exec.CmdRunResult) {
 
 	return dstPath, result
 }
+
+func (a *App) cacheID() string {
+	return fmt.Sprintf("%s-%s", a.app.Namespace, a.app.Name)
```
cppforlife: we should use UID here as well (similar to apps) to avoid accidental cache sharing in some failure-to-clean-up cases.

test/e2e/kappcontroller/imgpkg_bundle_test.go
```diff
@@ -123,8 +118,226 @@ spec:
...
+		kapp.RunWithOpts([]string{"deploy", "-a", pkgrName, "-f", "-"}, e2e.RunOpts{
+			StdinReader:  strings.NewReader(pkgrYaml),
+			OnErrKubectl: []string{"get", "pkgr", "-A", "-oyaml"},
+		})
+
+		kubectl.Run([]string{"get", "package", "pkg.test.carvel.dev.1.0.0"})
+
+		time.Sleep(5 * time.Second) // This sleep was added because it looks like there are 2 consecutive fetch
+		output := kubectl.Run([]string{"get", "packagerepository", pkgrName, "-oyaml"})
+		fmt.Println(output)
```
cppforlife: run away fmt.print

test/e2e/kappcontroller/imgpkg_bundle_test.go
```diff
@@ -123,8 +118,226 @@ spec:
...
+`, env.Namespace, pkgrName, registryNamespace) + sas.ForNamespaceYAML()
+
+		kapp.RunWithOpts([]string{"deploy", "-a", pkgrName, "-f", "-"}, e2e.RunOpts{
+			StdinReader:  strings.NewReader(pkgrYaml),
+			OnErrKubectl: []string{"get", "pkgr", "-A", "-oyaml"},
+		})
+
+		kubectl.Run([]string{"get", "package", "pkg.test.carvel.dev.1.0.0"})
+
+		time.Sleep(5 * time.Second) // This sleep was added because it looks like there are 2 consecutive fetch
```
cppforlife: @joe-kimmel-vmw @neil-hickey yall know anything about this?

test/e2e/kappcontroller/imgpkg_bundle_test.go
```diff
@@ -123,8 +118,226 @@ spec:
...
+	pkgrName := "pkgr-to-test-caching"
+	registryNamespace := "registry"
+	registryName := "test-registry"
+
+	cleanUp := func() {
+		// Delete App with kubectl first since kapp
+		// deletes ServiceAccount before App
+		kapp.Run([]string{"delete", "-a", pkgrName})
+		kapp.Run([]string{"delete", "-a", "secret-export"})
+		if isRegistryPresent {
```
cppforlife: `if isRegistryPresent {` is not needed since kapp delete is idempotent. it will succeed if the app does not exist.

test/e2e/kappcontroller/imgpkg_bundle_test.go
```diff
@@ -123,8 +118,226 @@ spec:
...
+	// If this changes, the skip-tls-verify domain must be updated to match
+	pkgrName := "pkgr-to-test-caching"
+	registryNamespace := "registry"
+	registryName := "test-registry"
+
+	cleanUp := func() {
+		// Delete App with kubectl first since kapp
+		// deletes ServiceAccount before App
+		kapp.Run([]string{"delete", "-a", pkgrName})
+		kapp.Run([]string{"delete", "-a", "secret-export"})
```
cppforlife: dead code line?

test/e2e/kappcontroller/imgpkg_bundle_test.go
```diff
@@ -123,8 +118,226 @@ spec:
...
+
+	logger.Section("deploy PackageRepository that auths to registry", func() {
+		pkgrYaml := fmt.Sprintf(`---
+apiVersion: packaging.carvel.dev/v1alpha1
+kind: PackageRepository
+metadata:
+  name: %[2]s
+  namespace: %[1]s
+spec:
+  syncPeriod: 30s
```
cppforlife, line 304: instead of 30s delay here, can we set this to be higher, and do an explicit kick in the test? (would in theory be more reliable and faster)

cppforlife (commented):
impl itself looks pretty good, would love to make e2e a bit better.

pkg/pkgrepository/app_fetch.go
joaopapereira: Unfortunately, there is no UID on the pkgrepository that is the reason why we have gone with namespace+app.
Do you have a different suggestion?

test/e2e/kappcontroller/imgpkg_bundle_test.go
joaopapereira: I tried the pause/unpause trick but I was not getting a reconciliation. I can try to do it again to see if it started working now

cppforlife: lmk if you need second set of eyes.

pkg/pkgrepository/app_fetch.go
cppforlife: all objects that include ObjectMeta should have it. i see it here for PackageRepository:

```
dk@dks-MacBook-Pro ipt % kubectl get pkgr -A -oyaml
apiVersion: v1
items:
- apiVersion: packaging.carvel.dev/v1alpha1
  kind: PackageRepository
  metadata:
    annotations: ...
    creationTimestamp: "2022-09-30T16:25:07Z"
    finalizers:
    - finalizers.packagerepository.packaging.carvel.dev/delete
    generation: 1
    labels:
      kapp.k14s.io/app: "1664555106454196000"
      kapp.k14s.io/association: v1.bdddd67774df3b38143b12b13282a26e
    managedFields: ...
    name: e2e-repo.test.carvel.dev
    namespace: kapp-controller-packaging-global
    resourceVersion: "3677"
    uid: 6b2583f3-dbc4-4e39-92e2-aeeaa2df9342 <-------------
  spec:
    fetch:
      imgpkgBundle:
        image: index.docker.io/k8slt/kc-e2e-test-repo@sha256:ddd93b67b97c1460580ca1afd04326d16900dc716c4357cade85b83deab76f1c
  status: ...
```

joaopapereira: Yes they have it. The problem I saw was that in the precise moment I need it the PackgeRepo does not have it yet

cppforlife (approved):

PR #913, opened 2022-09-29
Refactor app/pkg init logic

### LINKED ISSUE #827: `kctrl`: Post-alpha refactoring tracker for package authoring commands
Tasks at hand:
- [x] Refactor vendor utils to be a scoped struct with associated functions
- [x] Refactor workflows to move away from the step interface and update them according to refactoring described above
    - [x] Local directory
    - [x] Generic vendir step
    - [x] Git release workflow
    - [x] Files from git repo
    - [x] Helm chart from git repo
    - [x] Helm chart from OCI repository
- [x] Dedup and break up `package init` logic into smaller pieces
- [x] Move away from file utils and use existing art in in-built go packages (`ioutil`, `os`) instead
- [x] Refactor existing structs into packages
- [x] Handle how fetch option is passed between stages better
- [x] Verify that there is no regressions through means of existing tests and a bit of manual testing

100mik:
Refactors and cleans up app/package init logic

Fixes #827 

```release-note
NONE
```

##### Review Checklist:

  a link to that PR
  change

```docs

```


100mik:
Gotta work towards wrapping up pending cleanup on this through a bunch of reviews @praveenrewar

cli/pkg/kctrl/cmd/package/init/pkg_build.go
```diff
@@ -73,11 +77,61 @@ func (b *PackageBuild) SetObjectMeta(metaObj *metav1.ObjectMeta) {
...
+		if appTemplate.HelmTemplate != nil {
+			return true
+		}
+	}
+	return false
+}
+
+func (b *PackageBuild) ConfigureExportSection() {
+	fetchSource := b.GetObjectMeta().Annotations[appbuild.FetchContentAnnotationKey]
+	exportSection := *b.GetExport()
```
rohitagg2020:
```suggestion
	exportSection := b.GetExport()
```

cli/pkg/kctrl/cmd/app/init/git.go
```diff
@@ -11,62 +11,55 @@ import (
...
 
-func (g *GitStep) PreInteract() error { return nil }
-
-func (g *GitStep) Interact() error {
-	contents := g.vendirConfig.Directories[0].Contents
+func (g *GitConfiguration) Configure() error {
+	contents := g.vendirConfig.Contents()
 	if contents == nil {
-		err := g.initializeContentWithGit()
+		err := g.initializeContentWithGit(contents)
```
rohitagg2020: As `g` already has the information about `contents`, IMO we can skip passing contents.

cli/pkg/kctrl/cmd/app/init/git.go
```diff
@@ -11,62 +11,55 @@ import (
...
 	}
-	return g.getIncludedPaths()
+	return g.getIncludedPaths(contents)
 }
 
-func (g *GitStep) PostInteract() error { return nil }
-
-func (g *GitStep) initializeContentWithGit() error {
+func (g *GitConfiguration) initializeContentWithGit(contents []vendirconf.DirectoryContents) error {
 	//TODO Rohit need to check this how it should be done. It is giving path as empty.
```
rohitagg2020: This comment can be removed.

cli/pkg/kctrl/local/sources/fetch.go
```diff
@@ -0,0 +1,139 @@
...
+const (
+	FetchFromGithubRelease       string = "Github Release"
+	FetchFromGit                 string = "Git Repository"
+	FetchFromHelmRepo            string = "Helm Chart from Helm Repository"
+	FetchChartFromGit            string = "Helm Chart from Git Repository"
+	FetchFromLocalDirectory      string = "Local Directory"
+	MultipleFetchOptionsSelected string = "MultipleFetchOptionsSelected"
+)
+
+type FetchConfiguration struct {
```
praveenrewar: `FetchConfiguration` -> `Source`? (How does `Source.Configure` sound compared to `FetchConfiguration.Configure`?)

cli/pkg/kctrl/local/sources/fetch.go
```diff
@@ -0,0 +1,139 @@
...
+	vendirFileName      = "vendir.yml"
+	includeAllFiles     = "*"
+)
+
+const (
+	FetchFromGithubRelease       string = "Github Release"
+	FetchFromGit                 string = "Git Repository"
+	FetchFromHelmRepo            string = "Helm Chart from Helm Repository"
+	FetchChartFromGit            string = "Helm Chart from Git Repository"
+	FetchFromLocalDirectory      string = "Local Directory"
```
praveenrewar: I think we can remove `FetchFrom` as the package name itself is source? Also, should we create a type for these?

cli/pkg/kctrl/local/sources/fetch.go
```diff
@@ -0,0 +1,139 @@
...
+	f.ui.PrintInformationalText("Please provide the location from where your Kubernetes manifests or Helm chart can be fetched. This will be bundled as a part of the package.")
+
+	vendirConfig := NewVendirConfig(vendirFileName)
+	err := vendirConfig.Load()
+	if err != nil {
+		return "", nil, err
+	}
+
+	isHelmTemplateExistInPreviousOption := f.build.HasHelmTemplate()
+	previousFetchOptionSelected := vendirConfig.FetchMode(isHelmTemplateExistInPreviousOption)
```
praveenrewar: inline?

cli/pkg/kctrl/local/sources/fetch.go
```diff
@@ -0,0 +1,139 @@
...
+	if previousFetchOptionSelected == MultipleFetchOptionsSelected {
+		// As this is advanced use case, we dont know how to handle it.
+		f.ui.PrintInformationalText("Since vendir is syncing data from multiple resources, we will not reconfigure vendir.yml and run vendir sync.")
```
praveenrewar: Can we add a TODO instead of this comment?
Also, the informational text seems to be bit ambiguous, are we saying we will run `vendir sync` or are we saying we won't run `vendir sync`?

cli/pkg/kctrl/local/sources/fetch.go
```diff
@@ -0,0 +1,139 @@
...
+		return "", nil, nil
+	}
+
+	options := []string{FetchFromLocalDirectory, FetchFromGithubRelease, FetchFromHelmRepo, FetchFromGit, FetchChartFromGit}
+	previousFetchOptionIndex := f.getPreviousFetchOptionIndex(options, previousFetchOptionSelected)
+	defaultFetchOptionIndex := previousFetchOptionIndex
+	choiceOpts := ui.ChoiceOpts{
+		Label:   "Enter source",
+		Default: defaultFetchOptionIndex,
```
praveenrewar:
```suggestion
		Default: f.getPreviousFetchOptionIndex(options, previousFetchOptionSelected),
```

cli/pkg/kctrl/local/sources/fetch.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	currentFetchOptionIndex, err := f.ui.AskForChoice(choiceOpts)
+	if err != nil {
+		return "", nil, err
+	}
+	currentFetchOptionSelected := options[currentFetchOptionIndex]
+
+	if previousFetchOptionSelected != "" && currentFetchOptionSelected != previousFetchOptionSelected {
+		return "", nil, fmt.Errorf("Transitioning from one fetch option to another is not allowed. Earlier option selected: %s, Current Option selected: %s", previousFetchOptionSelected, currentFetchOptionSelected)
+	}
```
praveenrewar: If we don't allow transitioning from one source to another, maybe we shouldn't ask that question and just go ahead with the previous option along with a message?

praveenrewar: If we consider working on this, maybe a separate PR would make sense.

cli/pkg/kctrl/local/sources/fetch.go
```diff
@@ -0,0 +1,139 @@
...
+	vendirDirectories := vendirConfig.Directories()
+	if len(vendirDirectories) > 1 {
+		return currentFetchOptionSelected, nil, fmt.Errorf("More than 1 directory config found in the vendir file. (hint: Run vendir sync manually)")
+	}
+	if len(vendirDirectories) == 0 {
+		err := f.initializeVendirDirectorySection(vendirConfig)
+		if err != nil {
+			return currentFetchOptionSelected, nil, err
+		}
+	} else {
```
praveenrewar: `switch`?

cli/pkg/kctrl/local/sources/fetch.go
```diff
@@ -0,0 +1,139 @@
...
+		previousFetchOptionIndex = 0
+	} else {
+		for i, fetchTypeName := range manifestOptions {
+			if fetchTypeName == previousFetchOption {
+				previousFetchOptionIndex = i
+				break
+			}
+		}
+	}
+	return previousFetchOptionIndex
```
praveenrewar:
```suggestion
	for i, fetchTypeName := range manifestOptions {
		if fetchTypeName == previousFetchOption {
			return i
		}
	}
	return 0
```

cli/pkg/kctrl/local/sources/fetch.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func (f *FetchConfiguration) initializeVendirDirectorySection(vendirConfig *VendirConfig) error {
+	var directory vendirconf.Directory
+	directory = vendirconf.Directory{
```
praveenrewar: inline?

cli/pkg/kctrl/local/sources/fetch.go
```diff
@@ -0,0 +1,139 @@
...
+		Path: VendirSyncDirectory,
+		Contents: []vendirconf.DirectoryContents{
+			{
+				Path: ".",
+			},
+		},
+	}
+	directories := []vendirconf.Directory{directory}
+	vendirConfig.SetDirectories(directories)
+	err := vendirConfig.Save()
```
praveenrewar: It seems like this whole function should be a method on vendirConfig instead?

cli/pkg/kctrl/local/sources/vendir_runner.go
```diff
@@ -0,0 +1,110 @@
...
+func (r VendirRunner) RunSync() error {
+	_, err := os.Stat(vendirFileName)
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+	if os.IsNotExist(err) {
+		return nil
+	}
+
+	r.ui.PrintInformationalText("We will use vendir to fetch the data from the source to the local directory. Vendir allows us to declaratively state what should be in a directory and sync data sources into it. All the information entered above has been persisted into a vendir.yml file.")
```
praveenrewar: Maybe split it across multiple lines?

cli/pkg/kctrl/local/sources/vendir_runner.go
```diff
@@ -0,0 +1,110 @@
...
+
+type VendirRunner struct {
+	ui cmdcore.AuthoringUI
+}
+
+func NewVendirRunner(ui cmdcore.AuthoringUI) VendirRunner {
+	return VendirRunner{ui: ui}
+}
+
+func (r VendirRunner) RunSync() error {
```
praveenrewar: Since it's already called VendirRunner, maybe we should just use `Sync` instead of `RunSync`?

cli/pkg/kctrl/local/sources/vendir_runner.go
```diff
@@ -0,0 +1,110 @@
...
+	_, err := os.Stat(vendirFileName)
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+	if os.IsNotExist(err) {
+		return nil
+	}
```
praveenrewar: I guess sometimes it's okay to have nested `if`s
```suggestion
	if err != nil {
	        if os.IsNotExist(err) {
		       return nil
	        }
		return err
	}
```
BTW, what other kind of errors are we expecting here?

cli/pkg/kctrl/local/sources/vendir_runner.go
```diff
@@ -0,0 +1,110 @@
...
+		return err
+	}
+	err = r.runVendirSync()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r VendirRunner) printVendirFile() error {
```
praveenrewar: I feel that this is not needed. We can directly use `PrintActionableText` and `printFile`

cli/pkg/kctrl/local/sources/vendir_runner.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (r VendirRunner) printFile(filePath string) error {
+	file, err := os.Open(filePath)
+	if err != nil {
+		return fmt.Errorf("Printing file: %w", err)
+	}
+	defer func() {
+		file.Close()
+	}()
```
praveenrewar: `defer file.Close()`?

cli/pkg/kctrl/local/sources/vendir_runner.go
```diff
@@ -0,0 +1,110 @@
...
+	r.ui.PrintActionableText("Running vendir sync")
+	r.ui.PrintCmdExecutionText("vendir sync -f vendir.yml\n")
+	var stdoutBs, stderrBs bytes.Buffer
+
+	localCmdRunner := exec.NewPlainCmdRunner()
+	cmd := goexec.Command("vendir", []string{"sync", "-f", vendirFileName}...)
+	cmd.Stdin = nil
+	cmd.Stdout = &stdoutBs
+	cmd.Stderr = &stderrBs
+	err := localCmdRunner.Run(cmd)
```
praveenrewar: We might have a good reason but I am not sure why we are not using `cmd.Run()`

cli/pkg/kctrl/local/sources/fetch.go
100mik: Definitely better, 
I think the returned values make sense in this context as well

cli/pkg/kctrl/local/sources/fetch.go
100mik: Yep makes sense

cli/pkg/kctrl/local/sources/fetch.go
100mik: I actually did create a type `FetchMode`, with functions `IsLocal()` and `String()`
Since they are directly used as string options while prompting the user directly, it was making things a bit verbose

But I am happy to make the change

cli/pkg/kctrl/local/sources/fetch.go
100mik: I will clean it up, we made a call to not handle this at all, so we can get rid of it

cli/pkg/kctrl/local/sources/vendir_runner.go
100mik: Makes sense!

cli/pkg/kctrl/local/sources/vendir_runner.go
100mik: Most I can think of is lack of permissions

cli/pkg/kctrl/local/sources/vendir_runner.go
100mik: Yep, definitely

cli/pkg/kctrl/local/sources/vendir_runner.go
100mik: I will need to refresh my memory here

cli/pkg/kctrl/cmd/package/init/pkg_build.go
100mik: We won't be able to do length checks on `Export` if we make that change.
But I guess `GetExport` could return a non-pointer value

cli/pkg/kctrl/cmd/app/init/git.go
100mik: Sticking to this because all of the subsequent calls use contents, and we would end up calling `Contents()` multiple times. Can switch it up if you feel stronglyabout this

cli/pkg/kctrl/cmd/package/init/pkg_build.go
100mik: Made this change ^

cli/pkg/kctrl/cmd/app/init/git.go
100mik: Done

cli/pkg/kctrl/local/sources/fetch.go
praveenrewar: Although something to think about here is that the package name is sources, so outside this function it would be `sources.Source` 🤔

100mik: Also renaming GitConfiguration to GitSource, and the others too along similar lines

cli/pkg/kctrl/local/sources/fetch.go
100mik: Removed the prefix for now, leaving this open if we wanna have a type

cli/pkg/kctrl/local/sources/fetch.go
praveenrewar: Yep, was about to add that as a comment 👍🏻

cli/pkg/kctrl/local/sources/fetch.go
100mik: And to clarify we basically won't overwrite the file and run vendor sync, making it a bit clearer

100mik: Done, let me know if it is satisfactory once I push the changes (leaving this open for your to hit resolve)

cli/pkg/kctrl/local/sources/fetch.go
100mik: Yep, let's do a separate one for this
Closing this after I have created an issue

cli/pkg/kctrl/cmd/package/init.go
```diff
@@ -64,54 +64,63 @@ func (o *InitOptions) Run() error {
...
 	}
-	pkgBuild.Save()
+
+	// Source does not need to be conifgured if manifests are in the local directory
+	if sourceConfiguration != nil {
+		err = sourceConfiguration.Configure()
+		if err != nil {
+			return err
+		}
+	}
```
praveenrewar: Should this be directly handled in the main `Configure` function?

cli/pkg/kctrl/local/sources/fetch.go
100mik: Done!

cli/pkg/kctrl/cmd/package/init.go
```diff
@@ -215,6 +224,22 @@ func (o *InitOptions) newOrExistingPackageResources() (*v1alpha1.Package,
 	return pkg, pkgMetadata, pkgInstall, nil
 }
 
+func (o *InitOptions) updateAndDefaultPackageResources(pkgRefName string, pkgMetadata *v1alpha1.PackageMetadata, pkg *v1alpha1.Package) {
+	pkgMetadata.Name = pkgRefName
+	pkg.Spec.RefName = pkgRefName
+	pkgMetadata.Spec.DisplayName = strings.Split(pkgRefName, ".")[0]
+
+	shortDesc := pkgMetadata.Spec.ShortDescription
+	if len(shortDesc) == 0 {
```
praveenrewar: inline? 
What if this was removed by the user?

cli/pkg/kctrl/cmd/package/init.go
```diff
@@ -215,6 +224,22 @@ func (o *InitOptions) newOrExistingPackageResources() (*v1alpha1.Package,
...
+		pkgMetadata.Spec.ShortDescription = pkgRefName
+	}
+
+	longDesc := pkgMetadata.Spec.LongDescription
+	if len(longDesc) == 0 {
```
praveenrewar: Same as above.

cli/pkg/kctrl/local/sources/fetch.go
100mik: Done!

cli/pkg/kctrl/local/sources/fetch.go
100mik: Makes sense!

cli/pkg/kctrl/local/sources/fetch.go
100mik: >Although something to think about here is that the package name is sources, so outside this function it would be sources.Source 🤔

I think that is fine (Example: `resources.Resource`)

cli/pkg/kctrl/local/sources/git.go
```diff
@@ -11,62 +11,61 @@ import (
...
+
+func (g *GitConfiguration) Configure() error {
+	contents := g.vendirConfig.Contents()
 	if contents == nil {
-		err := g.initializeContentWithGit()
+		err := g.initializeContentWithGit(contents)
 		if err != nil {
 			return err
 		}
 	} else if contents[0].Git == nil {
```
praveenrewar: Should there be a check on len of contents?

cli/pkg/kctrl/local/sources/fetch.go
100mik: Done!

cli/pkg/kctrl/local/sources/fetch.go
100mik: Makes sense, making the move

cli/pkg/kctrl/local/sources/git.go
```diff
@@ -11,62 +11,61 @@ import (
...
-	return g.getIncludedPaths()
+	return g.getIncludedPaths(contents)
 }
 
-func (g *GitStep) PostInteract() error { return nil }
-
-func (g *GitStep) initializeContentWithGit() error {
+func (g *GitConfiguration) initializeContentWithGit(contents []vendirconf.DirectoryContents) error {
 	//TODO Rohit need to check this how it should be done. It is giving path as empty.
-	g.vendirConfig.Directories[0].Contents = append(g.vendirConfig.Directories[0].Contents, vendirconf.DirectoryContents{})
```
praveenrewar, line 57: In the configure we have contents as `g.vendirConfig.Contents()` but here it is `g.vendirConfig.Directories[0].Contents`, is this change intentional?

cli/pkg/kctrl/local/sources/git.go
```diff
@@ -11,62 +11,61 @@ import (
...
-	contents := g.vendirConfig.Directories[0].Contents
+func NewGitConfiguration(ui cmdcore.AuthoringUI, vendirConfig *VendirConfig) *GitConfiguration {
+	return &GitConfiguration{ui: ui, vendirConfig: vendirConfig}
+}
+
+func (g *GitConfiguration) Configure() error {
+	contents := g.vendirConfig.Contents()
 	if contents == nil {
-		err := g.initializeContentWithGit()
+		err := g.initializeContentWithGit(contents)
```
praveenrewar: I agree with @rohitagg2020, passing contents seems to be creating a lot of noise here.

cli/pkg/kctrl/local/sources/vendir_runner.go
100mik: Done

cli/pkg/kctrl/local/sources/vendir_runner.go
100mik: Done!

cli/pkg/kctrl/local/sources/vendir_runner.go
100mik: done!

cli/pkg/kctrl/cmd/package/init.go
100mik: We will get a nil pointer exception 😢

100mik: Before it reaches the function

cli/pkg/kctrl/cmd/package/init.go
100mik: Good question, but I cannot imagine how we will figure that out 🤔 
Since we are only setting doing this for values that are displayed by kctrl while listing/getting

I guess nudging users towards giving them values is not the worst idea 😆

100mik: I guess we can discuss this and put up an issue if we wanna give it more thought

cli/pkg/kctrl/cmd/package/init.go
100mik: Needs more discussion as I mentioned above

cli/pkg/kctrl/local/sources/git.go
100mik: We will be replacing `contents[0]` with `vendirConfig.Contents()[0]` in all the functions if we want to do that, true for all of these cases
Happy to do it if y'all think I should

cli/pkg/kctrl/local/sources/git.go
100mik: That would mean the user removed the item added by kctrl
But yep makes sense I guess, will do these split up changes tomorrow

cli/pkg/kctrl/cmd/package/init.go
praveenrewar: I am not sure how that will result in an exception. What I am suggesting is that instead of returning the sourceConfiguration, we directly call the sourceConfiguration.Configure in the main configure function based on the fetch option selected and then return the final result (an error).

cli/pkg/kctrl/local/sources/git.go
praveenrewar: Maybe we can do the `contents := g.vendirConfig...` in the functions where we are using `vendirConfig.Contents()[0]` more frequently?

cli/pkg/kctrl/cmd/package/init.go
100mik: Oh I thought the you meant the function should handle the nil check we are doing

That is what we were doing before, It will also eliminate the need of having an interface, happy to do that 
Will make that change tomorrow

100mik: Done!

cli/pkg/kctrl/local/sources/vendir_runner.go
100mik: Looks like it does the exact same thing 😆 
Making the change

100mik: I do not see the point of using the exec.Result object either, it just adds `(see status.usefulErrormessage for details)` which we do not want

100mik: Done!

cli/pkg/kctrl/cmd/package/init.go
100mik: Made this inline

cli/pkg/kctrl/cmd/package/init.go
100mik: Same resolution as above

cli/pkg/kctrl/local/sources/git.go
100mik: Looks like we are doing he same thing?
That is what `Contents` returns :/

cli/pkg/kctrl/local/sources/fetch.go
100mik: Another case for having a type is that the check `fetchMode == LocalDirectory` is in a bunch of places and we will need to add cases if we introduce more local workflows.
Having something like `IsLocal` which has this condition really helps out , hm.

cli/pkg/kctrl/cmd/package/init.go
```diff
@@ -64,54 +64,55 @@ func (o *InitOptions) Run() error {
...
 	if err != nil {
 		return err
 	}
 
-	pkgBuild.Save()
+	appSpec := pkgBuild.GetAppSpec()
+	if appSpec.Deploy == nil {
+		appSpec.Deploy = []kcv1alpha1.AppDeploy{{Kapp: &kcv1alpha1.AppDeployKapp{}}}
+	}
+	pkgBuild.SetAppSpec(appSpec)
```
praveenrewar: Maybe all of this should be under on function? (It seems like we are just setting deploy value to if it's nil).

cli/pkg/kctrl/cmd/package/repository/release/release.go
```diff
@@ -184,33 +183,33 @@ func (o *ReleaseOptions) Run() error {
...
-			},
-			ObjectMeta: metav1.ObjectMeta{
-				CreationTimestamp: metav1.NewTime(time.Now()),
-			},
-			Spec: build.PackageRepoBuildSpec{
-				Export: &build.PackageRepoBuildExport{
-					ImgpkgBundle: &build.PackageRepoBuildExportImgpkgBundle{},
+	_, err := os.Stat(pkgRepoBuildFilePath)
+	if err != nil {
+		if os.IsNotExist(err) {
```
praveenrewar:
```go
if !os.IsNotExist(err) {
      return &build.PackageRepoBuild{}, err
}
... snip ...
```

cli/pkg/kctrl/local/buildconfigs/pkg_build.go
```diff
@@ -73,11 +86,25 @@ func (b *PackageBuild) SetObjectMeta(metaObj *metav1.ObjectMeta) {
...
 	return
 }
+
+func (b *PackageBuild) HasHelmTemplate() bool {
+	appSpec := b.getAppSpec()
+	if appSpec == nil || appSpec.Template == nil {
+		return false
+	}
+	appTemplates := appSpec.Template
+	for _, appTemplate := range appTemplates {
```
praveenrewar: inline?

cli/pkg/kctrl/local/sources/template.go
```diff
@@ -1,50 +1,40 @@
...
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
 )
 
-const (
-	UpstreamFolderName = "upstream"
-	StdIn              = "-"
-)
-
-type TemplateStep struct {
+type TemplateConfiguration struct {
```
praveenrewar: Should this just be `Template`? (`TemplateConfiguration.Configure` sounds a bit weird)

cli/pkg/kctrl/local/sources/vendir_runner.go
```diff
@@ -0,0 +1,85 @@
...
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		r.ui.PrintCmdExecutionOutput(scanner.Text())
+	}
+	return nil
+}
+
+// TODO vendir sync failure. Reproduce: In case of 429 from github, we dont show errors today.
+func (r VendirRunner) runVendirSync() error {
```
praveenrewar: `runVendirSync` -> `sync`?

cli/pkg/kctrl/cmd/package/init.go
100mik: Done!

cli/pkg/kctrl/cmd/package/repository/release/release.go
100mik: Done!

cli/pkg/kctrl/local/buildconfigs/pkg_build.go
100mik: Done!

cli/pkg/kctrl/local/sources/template.go
100mik: Makes sense! I wonder if there could be a better name though

cli/pkg/kctrl/local/sources/vendir_runner.go
100mik: Done!

cli/pkg/kctrl/local/sources/template.go
100mik: Changed it to template for now

praveenrewar (approved):
LGTM!

PR #936, opened 2022-10-13
Add 'yanked' field to PackageCR

neil-hickey:
This is done as a new PR instead of https://github.com/vmware-tanzu/carvel-kapp-controller/pull/624


pkg/apiserver/apis/datapackaging/types.go
```diff
@@ -103,6 +103,17 @@ type PackageSpec struct {
...
+	// Yanked indicates that this package is no longer recommended for use.
+	// This package likely contains unsupported software, a CVE or a critical bug.
+	// +optional
+	Yanked *Yanked `json:"yanked,omitempty"`
+}
+
+// Yanked contains details of the rationale of why this version was yanked
+type Yanked struct {
+	Reason string      `json:"reason,omitempty" protobuf:"bytes,1,opt,name=reason"`
+	At     metav1.Time `json:"at,omitempty" protobuf:"bytes,2,opt,name=at"`
```
cppforlife, line 116: should we throw in a pointer on the type? im wondering if the value is not set, then round trip through k8s might end up with `null` value where it isnt already set.

pkg/reconciler/status.go
```diff
@@ -129,8 +129,15 @@ func (s *Status) SetDeleteCompleted(err error) {
 	} else {
 		// assume resource will be deleted, hence nothing to update
 	}
+}
 
-	s.UpdateFunc(s.S)
+// SetPackageYanked appends a new condition to indicate a package is yanked
+func (s *Status) SetPackageYanked(meta metav1.ObjectMeta, reason string) {
```
cppforlife, line 135: feels a bit weird to have a "yanked" knowledge within generic pkg/reconciler package. wdyt?

pkg/apis/kappctrl/v1alpha1/status.go
```diff
@@ -32,6 +32,8 @@ const (
 
 	Deleting     ConditionType = "Deleting"
 	DeleteFailed ConditionType = "DeleteFailed"
+
+	PackageYanked ConditionType = "PackageYanked"
```
cppforlife, line 36: it seems that this condition should be defined by a package that defines packaging concepts. apis/kappctrl is all about lower level primitives (ie App).

pkg/reconciler/status.go
```diff
@@ -75,26 +84,21 @@ func (s *Status) SetReconciling(meta metav1.ObjectMeta) {
...
 		Type:   kcv1alpha1.Deleting,
 		Status: corev1.ConditionTrue,
 	})
 
 	s.S.FriendlyDescription = "Deleting"
-
-	s.UpdateFunc(s.S)
 }
 
 func (s *Status) SetReconcileCompleted(err error) {
```
cppforlife, line 100: it seems that we are modifying generic pkg/reconciler. im guessing changes would affect other CRs like App CR. it seems that just removal of reconciling vs everything would cause problems, right?

cppforlife (commented):
might be worth jumping on zoom to talk through comments i left.

PR #939, opened 2022-10-17
Instantiate coreClient while using reconciler only for the dev command

### LINKED ISSUE #917: `kctrl`:Remove dependency on cluster for the authoring commands
Today, some of the Package authoring commands are dependent on a connection to the cluster

The authoring commands should not be dependent on the cluster unless resources are being deployed onto the cluster

**Tasks:**
- [ ] Ensure that core client is instantiated only when required in the `local` package

---
Vote on this request

100mik:
It ensures that package authoring commands do not need an active cluster connection

Fixes #917 

```release-note
Package authoring commands no longer require a cluster connection
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/cli/pkg/kctrl/cmd/dev/cmd.go
+++ b/cli/pkg/kctrl/cmd/dev/cmd.go
@@ -72,10 +72,11 @@ func (o *DevOptions) Run() error {
 	reconciler := cmdlocal.NewReconciler(o.depsFactory, cmdRunner, o.logger)
 
 	reconcileErr := reconciler.Reconcile(configs, cmdlocal.ReconcileOpts{
-		Local:     o.Local,
-		KbldBuild: o.KbldBuild,
-		Delete:    o.Delete,
-		Debug:     o.Debug,
+		Local:           o.Local,
+		KbldBuild:       o.KbldBuild,
+		Delete:          o.Delete,
+		Debug:           o.Debug,
+		DeployResources: true,
 
 		BeforeAppReconcile: o.beforeAppReconcile,
 		AfterAppReconcile:  o.afterAppReconcile,
--- a/cli/pkg/kctrl/local/reconciler.go
+++ b/cli/pkg/kctrl/local/reconciler.go
@@ -44,10 +44,11 @@ func NewReconciler(depsFactory cmdcore.DepsFactory,
 }
 
 type ReconcileOpts struct {
-	Local     bool
-	KbldBuild bool
-	Delete    bool
-	Debug     bool
+	Local           bool
+	KbldBuild       bool
+	Delete          bool
+	Debug           bool
+	DeployResources bool
 
 	BeforeAppReconcile func(kcv1alpha1.App, *fakekc.Clientset) error
 	AfterAppReconcile  func(kcv1alpha1.App, *fakekc.Clientset) error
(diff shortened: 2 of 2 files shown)


cli/pkg/kctrl/local/reconciler.go
```diff
@@ -44,10 +44,11 @@ func NewReconciler(depsFactory cmdcore.DepsFactory,
...
 type ReconcileOpts struct {
-	Local     bool
-	KbldBuild bool
-	Delete    bool
-	Debug     bool
+	Local           bool
+	KbldBuild       bool
+	Delete          bool
+	Debug           bool
+	DeployResources bool
```
praveenrewar, line 51: There are get/list commands as well where we need connection but we don't deploy resources. Maybe we can simply use `coreClient` or `clientSet` or something similar?

cli/pkg/kctrl/local/reconciler.go
```diff
@@ -81,17 +82,24 @@ func (o *Reconciler) Reconcile(configs Configs, opts ReconcileOpts) error {
...
-	coreClient, err := o.depsFactory.CoreClient()
-	if err != nil {
-		return fmt.Errorf("Getting core client: %s", err)
-	}
-
-	err = o.hackyConfigureKubernetesDst(coreClient)
-	if err != nil {
-		return err
+	var coreClient kubernetes.Interface
+	fmt.Println(opts.DeployResources)
```
praveenrewar: nit: remove Print statement

cli/pkg/kctrl/cmd/dev/cmd.go
```diff
@@ -72,10 +72,11 @@ func (o *DevOptions) Run() error {
...
 	reconcileErr := reconciler.Reconcile(configs, cmdlocal.ReconcileOpts{
-		Local:     o.Local,
-		KbldBuild: o.KbldBuild,
-		Delete:    o.Delete,
-		Debug:     o.Debug,
+		Local:           o.Local,
+		KbldBuild:       o.KbldBuild,
+		Delete:          o.Delete,
+		Debug:           o.Debug,
+		DeployResources: true,
```
praveenrewar, line 79: Shouldn't this be false for dev command or am I missing something?

cli/pkg/kctrl/local/reconciler.go
100mik: oops took care of that

cli/pkg/kctrl/local/reconciler.go
100mik: > There are get/list commands as well where we need connection but we don't deploy resources.

Elaborate on what you are thinking about?
From my perspective, `Deploy` is true only when `kapp` (or the deploy section) is going to be used.

Are you trying to say that we also use it to grab service accounts? That is how we "configure" kapp from my perspective

cli/pkg/kctrl/cmd/dev/cmd.go
100mik: The `dev` command will be the only one deploying resources.
Thus, this value is true.

Side-note: This also ensures that we do not have to make changes to any other file affected by the ongoing refactoring 😄 as the value defaults to false for all other commands

praveenrewar: Never mind, I mixed up things. Just re read the issue and I think this is fine.

cli/pkg/kctrl/local/reconciler.go
praveenrewar: Never mind, I mixed up things. Just re read the issue and I think this is fine.

praveenrewar (approved):
LGTM!

cli/pkg/kctrl/cmd/dev/cmd.go
rohitagg2020: For `kctrl dev`, we deploy the resources to the cluster. I think thats why, for dev it is set to true.

rohitagg2020:
LGTM

PR #962, opened 2022-10-28
Adding test case for local dir test case.

rohitagg2020:
Adding test case for local dir test case.

##### Review Checklist:

  change

```docs

```

### DIFF
--- a/cli/test/e2e/package_authoring_e2e_test.go
+++ b/cli/test/e2e/package_authoring_e2e_test.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


cli/test/e2e/package_authoring_e2e_test.go
```diff
@@ -876,6 +998,12 @@ spec:
 		const pkgDir = "./carvel-artifacts/packages/testpackage.corp.dev/"
 		promptOutput := newPromptOutput(t)
 
+		switch testcase.Name {
+		case "Local dir Flow":
+			err := createContentForLocalDir(workingDir)
```
praveenrewar: I think we can do this outside the loop.

rohitagg2020: To do it outside the loop, I will not be able to cleaup `workingDir` once the test case is finished. And I think it make sense to cleanup everything once testcase is finished. What do you think?

praveenrewar: > I will not be able to cleaup workingDir once the test case is finished

I see. Maybe we can always create it then? (Would that cause a conflict in other test cases)
I feel that the increase in number of conditionals for each test case is an indication that we shouldn’t use table tests in this case 🤔

rohitagg2020: I think its better to move it out from table test case. Dont want to create something which wont be used by other TC.

PR #963, opened 2022-10-28
`kctrl`: Garbage collect .imgpkg directory only if we create it

100mik:
We do not want to delete user supplied `.imgpkg` folders.

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
+++ b/cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
@@ -71,15 +71,14 @@ func (b *AppSpecBuilder) Build() (kcv1alpha1.AppSpec, error) {
 	}
 
 	// Make lock output directory if it does not exist
-	tmpImgpkgFolder := LockOutputFolder
-	_, err := os.Stat(tmpImgpkgFolder)
+	_, err := os.Stat(LockOutputFolder)
 	if err != nil && os.IsNotExist(err) {
-		err := os.Mkdir(tmpImgpkgFolder, os.ModePerm)
+		err := os.Mkdir(LockOutputFolder, os.ModePerm)
 		if err != nil {
 			return kcv1alpha1.AppSpec{}, err
 		}
+		defer os.RemoveAll(LockOutputFolder)
 	}
-	defer os.RemoveAll(LockOutputFolder)
 
 	// Build images and resolve references using reconciler
 	tempImgpkgLockPath := filepath.Join(LockOutputFolder, LockOutputFile)
--- a/cli/pkg/kctrl/cmd/package/repository/release/release.go
+++ b/cli/pkg/kctrl/cmd/package/repository/release/release.go
@@ -118,30 +118,23 @@ func (o *ReleaseOptions) Run() error {
 		return err
 	}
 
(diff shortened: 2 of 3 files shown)


cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
```diff
@@ -78,8 +78,8 @@ func (b *AppSpecBuilder) Build() (kcv1alpha1.AppSpec, error) {
 		if err != nil {
 			return kcv1alpha1.AppSpec{}, err
 		}
+		defer os.RemoveAll(LockOutputFolder)
```
praveenrewar, line 80: Let's also add a test for this (not deleting user supplied folders)?

cli/pkg/kctrl/cmd/package/repository/release/release.go
```diff
@@ -140,8 +140,8 @@ func (o *ReleaseOptions) Run() error {
 		if err != nil {
 			return err
 		}
+		defer os.RemoveAll(filepath.Join(wd, LockOutputFolder))
```
rohitagg2020:
```suggestion
		defer os.RemoveAll(tmpImgpkgFolder)
```

cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
rohitagg2020:
```suggestion
		defer os.RemoveAll(tmpImgpkgFolder)
```
I think it is more readable as we are creating `tmpImgpkgFolder` in `os.Mkdir(tmpImgpkgFolder, os.ModePerm)`

100mik: Adding the test, meanwhile, I switched to using `LockOutputFolder` directly everywhere as it might or might not be temporary

cli/pkg/kctrl/cmd/package/repository/release/release.go
100mik: Removed usage of wd as it is not necessary and started using `LockOutputFolder` directly

100mik:
Related to the cleanup that Rohit suggested, I am removing the logic where we pass down working directory while releasing repositories like we do for package artifacts. Does not seem necessary as we do not have the concept of a `copy-to` flag while releasing repositories. 

An addition of a flag like that can go in as a separate PR.

100mik:
cc: @rohitagg2020 @praveenrewar can we take another swing at getting this merged?

praveenrewar:
@100mik Is this still relevant? If so can we have a test for this?

xlf12:
Well, typically we try to build images based on a lot of generated things, especially references to other bundles.

In doing so, we copy things into a build folder before generating further input for the bundle. The carvel tool kctrl behaves a bit strangely in this regard, as it uses the current source code folder as the build folder and even has its own opinion on a build cycle. 

A switch that disables garbage collection would be a good approach IMHO.

Ok, that's a small point, at least until I can get these https://github.com/carvel-dev/kapp-controller/issues/943 problems to work. Which allows us to really do this.

PR #989, opened 2022-11-27
Do not run/stop workflows that are not required

praveenrewar:
- Do not run kapp-controller tests for cli changes
- Cancel previous in progress runs of workflows on same branch

### DIFF
--- a/.github/workflows/kind-action.yml
+++ b/.github/workflows/kind-action.yml
@@ -5,6 +5,10 @@ on:
     types: [opened, reopened, synchronize]
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   run-tests:
     runs-on: ubuntu-latest
--- a/.github/workflows/kind-action.yml
+++ b/.github/workflows/kind-action.yml
@@ -3,6 +3,10 @@ name: Kind Cluster E2E tests
 on:
   pull_request:
     types: [opened, reopened, synchronize]
+    paths-ignore:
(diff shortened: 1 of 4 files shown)


.github/workflows/kind-action.yml
```diff
@@ -3,8 +3,16 @@ name: Kind Cluster E2E tests
 on:
   pull_request:
     types: [opened, reopened, synchronize]
+    paths-ignore:
+    - 'docs/**'
+    - '*.md'
+    - 'cli/**'
   workflow_dispatch:
 
+concurrency:
```
neil-hickey, line 12: oh this is cool, wasn't aware of it. Nice!

.github/workflows/upgrade-testing.yml
```diff
@@ -12,6 +12,10 @@ name: upgrade-testing
       - 'docs/**'
       - '*.md'
 
+concurrency:
```
neil-hickey, line 17: Do we intend for the upgrade tests to run on cli changes? @praveenrewar

neil-hickey (approved):
One question about whether we need to run the upgrade tests on the cli, other than that, looks good! This will help our PRs go faster, nice!

.github/workflows/upgrade-testing.yml
praveenrewar: Ah, good catch! Let me make the changes here too.

praveenrewar: Done!

praveenrewar:
> One question about whether we need to run the upgrade tests on the cli

Good catch! We don't need to, I have made the changes accordingly :)

PR #992, opened 2022-11-29
Remove version constraint on kind tests

### LINKED ISSUE #971: Should `kapp-controller` be running tests on latest version of k8s available?
As of today, it looks like the `test-gh` workflow runs exclusively on Kubernetes version 1.20.4. 
Where as "Kind cluster E2E tests" is locked onto using `kind v0.14.0` which defaults to using Kubernetes 1.24.0.

Are we intentionally maintaining a minimum supported version and maximum supported version here?
If that is not the case, I believe removing the version constraint on `helm/kind-action` will run the tests on the latest available image and this might give us a heads up if there are any breaking changes we need to handle.

100mik:
A version of kind locks onto a version of Kubernetes image. To ensure that we are testing against the latest image available we should use the latest version of kind available via the action

Fixes #971 

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/.github/workflows/kind-action.yml
+++ b/.github/workflows/kind-action.yml
@@ -12,7 +12,6 @@ jobs:
       - name: Create k8s Kind Cluster
         uses: helm/kind-action@v1.4.0
         with:
-          version: v0.14.0
           kubectl_version: v1.25.0
           cluster_name: kinder
       - uses: actions/checkout@v3.1.0


100mik:
The new [run](https://github.com/vmware-tanzu/carvel-kapp-controller/actions/runs/3571639676/jobs/6003700006#step:2:20) seems to have the intended effect as opposed to this old [one](https://github.com/vmware-tanzu/carvel-kapp-controller/actions/runs/3568957558/jobs/5998398144#step:2:20).

cppforlife (approved):

PR #1021, opened 2022-12-12
Set seccompProfile type for both containers

praveenrewar:
Set seccompProfile type for kapp-controller and kapp-controller-sidecarexec containers

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```


neil-hickey:
I don't think this will be needed @praveenrewar , we will opt for a new minor to be cut for CE 1.4. So as long as this is on develop, this should be good :)

PR #1047, opened 2023-01-10
restrict exposing values in KC package

### LINKED ISSUE #1046: Restrict exposing some values we manipulate as part of KC Package
**What did you expect:**
Today, We expose every field as part of KC package that is there in [config/values.yml](https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/config/values.yml) we should restrict exposing values we manipulate during development of KC Package 

---
Vote on this request

sethiyash:
Fixes #1046 

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```


config-release/values-schema.yaml
```diff
@@ -0,0 +1,17 @@
+#@data/values-schema
+---
+#@overlay/match missing_ok=True
+#@overlay/match-child-defaults
+#@schema/desc "Configuration explicitly for developing secretgen-controller"
+dev:
+  #@schema/desc "Location of kapp-controller image"
+  image_repo: "docker.io/ysethiya/carvel-kapp-controller"
```
neil-hickey:
```suggestion
  image_repo: "docker.io/k14stest/kapp-controller-test"
```

config-release/values-schema.yaml
```diff
@@ -0,0 +1,17 @@
+#@data/values-schema
+---
+#@overlay/match missing_ok=True
+#@overlay/match-child-defaults
+#@schema/desc "Configuration explicitly for developing secretgen-controller"
```
neil-hickey:
```suggestion
#@schema/desc "Configuration explicitly for developing kapp-controller"
```

config/values-schema.yml
```diff
@@ -0,0 +1,18 @@
+#@data/values-schema
```
neil-hickey, line 1: We should align our values as closely as we can with TCE:

https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/kapp-controller/0.44.1/bundle/config/values.yaml

This can be done in this PR or a follow-up. But the downstream consumers of TCE will expect the values of their kc package to align, or else they will have to add an overlay. For secretgen-controller we chose to align as much as we felt was valuable to consumers to expose.

config/values-schema.yml
```diff
@@ -0,0 +1,18 @@
...
+---
+#@schema/desc "The namespace in which to deploy kapp-controller"
+namespace: kapp-controller
+#@schema/desc "Whether to create namespace specified for kapp-controller"
+create_namespace: true
+#@schema/desc "packaging global namespace value"
```
neil-hickey: We should be consistent about starting the description with a Capital letter. And some of these don't really describe the values so much as a long version of the key, e.g.:

```
#@schema/desc "dangerous enable pprof flag"
dangerous_enable_pprof: false
```

I still have to go read the documentation to find out why or why not to enable this flag. It would be helpful if the values were documenting as best we could such that you don't need to go too far to know what each key is for

hack/build-release.sh
```diff
@@ -8,8 +8,11 @@ mkdir -p .imgpkg/
 # makes the get_kappctrl_ver function available (scrapes version from git tag)
 source $(dirname "$0")/version-util.sh
 
-ytt -f config/ -f config-release -v kapp_controller_version="$(get_kappctrl_ver)" --data-values-env=KCTRL | kbld --imgpkg-lock-output .imgpkg/images.yml -f- > ./tmp/release.yml
+export version="$(get_kappctrl_ver)"
 
+yq eval '.metadata.annotations."kapp-controller.carvel.dev/version" = env(version)' -i "config/deployment.yml"
+
+ytt -f config/ -f config-release -v dev.kapp_controller_version="$(get_kappctrl_ver)" --data-values-env=KCTRL | kbld --imgpkg-lock-output .imgpkg/images.yml -f- > ./tmp/release.yml
```
neil-hickey: I'm still trying to figure out why secretgen-controller needed the `v` prefix to be removed but KC doesn't:

https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/develop/hack/build-and-publish-release.sh#L15

config-release/values-schema.yaml
```diff
@@ -0,0 +1,17 @@
...
+dev:
+  #@schema/desc "Location of kapp-controller image"
+  image_repo: "docker.io/ysethiya/carvel-kapp-controller"
+  #@schema/desc "Development version"
+  kapp_controller_version: develop
+  #@schema/desc "Comma separated list of supported architectures"
+  platform: "linux/amd64,linux/arm64"
+  #@schema/desc "image push flag"
```
neil-hickey: Let's copy the descriptions from SGC: 

https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/develop/config/dev/values-schema.yml#L7

neil-hickey (commented):
Could you attach a copy of the release.yml that get's created / links to the package-bundle you created on your fork? Looks like it's not a public release with the tested changes, (or maybe its a draft) - so I can't see the actual results of this PR.

config/values-schema.yml
sethiyash: Sure. Got the description from `dev.md` it is used to enable profiling by setting it to true.

config/values-schema.yml
sethiyash: So you mean we should have all these new values as part of our values-schema.yml or the value fields we already have should have exact same value as per TCE values.yaml?

sethiyash:
> Could you attach a copy of the release.yml that get's created / links to the package-bundle you created on your fork? Looks like it's not a public release with the tested changes, (or maybe its a draft) - so I can't see the actual results of this PR.

https://github.com/sethiyash/carvel-kapp-controller/releases/tag/v3.2.0 This is the release I have made on my forked repo. BTW `release.yml` that is getting generated has some args coming null in deployment resource. Looking into it.

hack/build-release.sh
sethiyash: Not sure about this part. Should I do the same here as well?

config/values-schema.yml
neil-hickey: > So you mean we should have all these new values as part of our values-schema.yml 

Right, not that we *should* have them all, so much as we have a direct consumer that needs these things configurable via values. So we should consider them in our official values schema. Otherwise downstream consumers will have to add an overlay to expose these values

sethiyash:
> https://github.com/sethiyash/carvel-kapp-controller/releases/tag/v3.2.0 This is the release I have made on my forked repo. BTW release.yml that is getting generated has some args coming null in deployment resource. Looking into it.

Resolved this now was happening earlier as `yq` command messing up with the indentation of ytt part considering it as comment. Now, achieving it with `ytt`

config/values-schema.yml
sethiyash: Done added the values aligning with TCE

neil-hickey:
looks like this is not in a good state right now @sethiyash , lemme know if you want another set of eyes to get it green

sethiyash:
> looks like this is not in a good state right now @sethiyash , lemme know if you want another set of eyes to get it green

@neil-hickey It's good to review now please take a look when you get a chance.

config/values-schema.yml
```diff
@@ -0,0 +1,64 @@
+#@data/values-schema
+---
```
neil-hickey, line 2: hey @sethiyash , I guess maybe there was a misunderstanding here. But this is our public values file. So when I go to install KC via a package install I can configure all these values.

But they do nothing right now, so they actually have less value by including them.

Take SGC for example: when I choose an updateStrategy in my values file, the deployment is changed based on this:

https://github.com/carvel-dev/secretgen-controller/blob/develop/config/package-bundle/config/deployment.yml#L12

I think we want something similar here, either via overlays or via branching (i.e ifs/ switches whatever) directly in the deployment.yml

neil-hickey (changes requested):

neil-hickey:
done as part of https://github.com/carvel-dev/kapp-controller/pull/1111

PR #1052, opened 2023-01-17
Do not set seccompProfile in kc 0.44.x

praveenrewar:
Do not set seccompProfile
Setting seccompProfile violates the policies set by default SCCs in older versions of Openshift

#### Tests
- Tested on Openshift 4.10
- Tested on Openshift 4.11 (There's a warning to set seccompProfile but the default SCCs set it to RuntimeDefault for all pods)

### DIFF
--- a/config/deployment.yml
+++ b/config/deployment.yml
@@ -59,8 +59,6 @@ spec:
           capabilities:
             drop:
             - ALL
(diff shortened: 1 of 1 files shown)


neil-hickey (approved):
LGTM

PR #1053, opened 2023-01-17
Do not set seccompProfile

praveenrewar:
Do not set seccompProfile
Setting seccompProfile violates the policies set by default SCCs in older versions of Openshift

#### Tests
- Tested on Openshift 4.10
- Tested on Openshift 4.11 (There's a warning to set seccompProfile but the default SCCs set it to RuntimeDefault for all pods)

### DIFF
--- a/config/deployment.yml
+++ b/config/deployment.yml
@@ -59,8 +59,6 @@ spec:
           capabilities:
             drop:
             - ALL
-          seccompProfile:
-            type: RuntimeDefault
       - name: kapp-controller-sidecarexec
         image: kapp-controller
         args: ["--sidecarexec"]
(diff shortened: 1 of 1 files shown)


cppforlife:
@praveenrewar lets add to the PR which versions of openshift you've tested on.

neil-hickey:
Maybe you can help me understand why we are backing out of this change:

> Setting seccompProfile violates the policies set by default SCCs in older versions of Openshift

So in `Openshift 4.10` we can't set the policy? But in `Openshift 4.11+` it is recommended to set and gives a warning if you don't?

praveenrewar:
> So in Openshift 4.10 we can't set the policy? But in Openshift 4.11+ it is recommended to set and gives a warning if you don't?

Yep, that's correct. 
In 4.11+, along with the warning, it also sets the seccompProfile to RuntimeDefault, so the warning shouldn't become a failure in future.

neil-hickey (approved):
LGTM

PR #1077, opened 2023-01-30
Disable ytt validations while building packages

### LINKED ISSUE #1028: `kctrl package release` validation issue
**What steps did you take:**

```sh
git clone https://github.com/matteo-magni/carvel-lab.git
cd carvel-lab/kctrl/authoring-sample
kctrl package release --repo-output ./repository --tag 0.0.1 --version 0.0.1 --yes
```

**What happened:**

If some package values are required to be provided (i.e. via schema validation) the package won't build, and the error returned is

```text
kctrl: Error: Reconciling: ytt:
  Error:
    Validating final data values:
  name
    from: schema.yml:4
    - must be: length >= 1 (by: schema.yml:3)
      found: length = 0
```

**What did you expect:**

I expect some variables to be required at install time but not at build time.
For example, if I provide a package to deploy infrastructure on AWS I want to force the user to choose the destination region, so no default can be set.

The same problem was raised by me and @joostvdg in another GH issue two months ago (#927).

**Environment:**

- kctrl version: v0.44.1

---
Vote on this request

100mik:
`kctrl` runs the values in the packaged config through `ytt` and `kbld` so that it can build any dependent images. Since, we do not expect validated values to be passed at this stage. We should be skipping the validation checks.

Fixes #1028 

```release-note
The `package release` command now supports values schemas defining values without valid defaults
```

- I think given how validations are put together and how `ytt` is, it would not produce invalid YAML if a value is missing. So `kbld` can definitely parse `ytt` output.
- Looks like a test for this would require piping for testing packages built from local (we do not have this today)

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
+++ b/cli/pkg/kctrl/cmd/app/release/app_spec_builder.go
@@ -34,6 +34,8 @@ type AppSpecBuilderOpts struct {
 	BundleImage   string
 	Debug         bool
 	BundleTag     string
+
+	BuildYttValidations bool
 }
 
 func NewAppSpecBuilder(depsFactory cmdcore.DepsFactory, logger logger.Logger, ui cmdcore.AuthoringUI, opts AppSpecBuilderOpts) *AppSpecBuilder {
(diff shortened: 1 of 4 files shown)


100mik:
@praveenrewar do you think this PR should introduce testing for the local directory workflow?
Or should adding a test for this scenario along with a generic local directory scenario be a separate PR all together?

praveenrewar:
I think adding adding the test as part of the same PR makes sense, as the tests would be relevant to current changes.

100mik:
Did not add generic piping for testing local directory cases with the interactive flow.
But did add a separate test testing for this use case. Let me know what you think!

100mik:
Tests are failing because setup action got rate limited 😢 
I wonder if this is caused by the migration?

renuy:
check if github token is set, have been seeing a warning for that.

100mik:
I do not have access to do that :(

praveenrewar (approved):

PR #1092, opened 2023-02-03
check that downward api supports dots in field paths

cppforlife:
just a little unit test that shows how to use downward api field path with dots.


neil-hickey (approved):
lgtm, looks like you have to rebase and sign off your commits @cppforlife. Then we can merge 👍

neil-hickey:
@cppforlife just putting this back on your radar so we can get it merged in

PR #1100, opened 2023-02-20
Introduce dry run flag for kctrl package install and package repo add

### LINKED ISSUE #1029: dry-run option for "install" and "add" commands
When teaching new folks about what kctrl is doing behind the scenes to get them more familiar with the gitops workflows it enables, it's a two step process currently.

1. Run install package / add repository commands
2. kubectl get those items and explain what they are

Ideally we'd have a dry run option that simply outputs the yaml that it's attempting to install/add.

We could eventually consider adding this to update commands as well, but I think initially it'd be most useful and simplest to dry-run on the create commands.

---
Vote on this request

100mik:
It introduces a flag that let's users observe the resource YAML that `kctrl` would apply, without applying them to the cluster

Fixes #1029 

```release-note
Introduced dry-run flag for `package install` and `package repo add`
```

To ensure that we do not introduce a lot of branching, this implementation implements logic for dry-run completely separately from the existing workflow.

##### Review Checklist:

  a link to that PR
  change

```docs
kctrl command reference will have to be updated
```

### DIFF
--- a/cli/pkg/kctrl/cmd/package/installed/create_or_update.go
+++ b/cli/pkg/kctrl/cmd/package/installed/create_or_update.go
@@ -54,6 +54,8 @@ type CreateOrUpdateOptions struct {
 
 	install bool
 
+	DryRun bool
+
 	Name                 string
 	NamespaceFlags       cmdcore.NamespaceFlags
 	SecureNamespaceFlags cmdcore.SecureNamespaceFlags
@@ -100,6 +102,7 @@ func NewCreateCmd(o *CreateOrUpdateOptions, flagsFactory cmdcore.FlagsFactory) *
 	cmd.Flags().StringVar(&o.serviceAccountName, "service-account-name", "", "Name of an existing service account used to install underlying package contents, optional")
 	cmd.Flags().StringVar(&o.valuesFile, "values-file", "", "The path to the configuration values file, optional")
 	cmd.Flags().BoolVar(&o.values, "values", true, "Add or keep values supplied to package install, optional")
+	cmd.Flags().BoolVar(&o.DryRun, "dry-run", false, "Print YAML for resources being applied to the cluster without applying them, optional")
 
 	o.WaitFlags.Set(cmd, flagsFactory, &cmdcore.WaitFlagsOpts{
 		AllowDisableWait: true,
@@ -145,6 +148,7 @@ func NewInstallCmd(o *CreateOrUpdateOptions, flagsFactory cmdcore.FlagsFactory) 
 	cmd.Flags().StringVar(&o.serviceAccountName, "service-account-name", "", "Name of an existing service account used to install underlying package contents, optional")
 	cmd.Flags().StringVar(&o.valuesFile, "values-file", "", "The path to the configuration values file, optional")
 	cmd.Flags().BoolVar(&o.values, "values", true, "Add or keep values supplied to package install, optional")
+	cmd.Flags().BoolVar(&o.DryRun, "dry-run", false, "Print YAML for resources being applied to the cluster without applying them, optional")
 
 	o.WaitFlags.Set(cmd, flagsFactory, &cmdcore.WaitFlagsOpts{
 		AllowDisableWait: true,
@@ -200,6 +204,8 @@ func NewUpdateCmd(o *CreateOrUpdateOptions, flagsFactory cmdcore.FlagsFactory) *
 }
 
 func (o *CreateOrUpdateOptions) RunCreate(args []string) error {
+	o.createdAnnotations = NewCreatedResourceAnnotations(o.Name, o.NamespaceFlags.Name)
+
 	if o.pkgCmdTreeOpts.PositionalArgs {
 		o.Name = args[0]
 	}
@@ -215,6 +221,14 @@ func (o *CreateOrUpdateOptions) RunCreate(args []string) error {
 	err := o.SecureNamespaceFlags.CheckForDisallowedSharedNamespaces(o.NamespaceFlags.Name)
 	if err != nil {
 		return err
+	}
+
+	if o.DryRun {
+		err := PackageInstalledDryRun{o}.PrintResources()
+		if err != nil {
+			return (fmt.Errorf("Generating resource YAML: %s", err))
+		}
+		return nil
 	}
 
 	if len(o.version) == 0 {
@@ -247,8 +261,6 @@ func (o *CreateOrUpdateOptions) RunCreate(args []string) error {
 			return err
 		}
 	}
-
-	o.createdAnnotations = NewCreatedResourceAnnotations(o.Name, o.NamespaceFlags.Name)
 
 	// Fallback to update if resource exists
 	if pkgInstall != nil && err == nil {
--- a/cli/pkg/kctrl/cmd/package/installed/package_install_dry_run.go
+++ b/cli/pkg/kctrl/cmd/package/installed/package_install_dry_run.go
@@ -1,0 +1,147 @@
+package installed
+
+import (
+	"fmt"
+
+	cmdcore "github.com/vmware-tanzu/carvel-kapp-controller/cli/pkg/kctrl/cmd/core"
+	kcpkgv1alpha1 "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
+	versions "github.com/vmware-tanzu/carvel-vendir/pkg/vendir/versions/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
+)
+
+type PackageInstalledDryRun struct {
+	*CreateOrUpdateOptions
+}
+
+func (d PackageInstalledDryRun) PrintResources() error {
+	const yamlSeperator = "---\n"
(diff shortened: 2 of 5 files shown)


cli/pkg/kctrl/cmd/package/installed/create_or_update.go
```diff
@@ -145,6 +148,7 @@ func NewInstallCmd(o *CreateOrUpdateOptions, flagsFactory cmdcore.FlagsFactory)
 	cmd.Flags().StringVar(&o.serviceAccountName, "service-account-name", "", "Name of an existing service account used to install underlying package contents, optional")
 	cmd.Flags().StringVar(&o.valuesFile, "values-file", "", "The path to the configuration values file, optional")
 	cmd.Flags().BoolVar(&o.values, "values", true, "Add or keep values supplied to package install, optional")
+	cmd.Flags().BoolVar(&o.DryRun, "dry-run", false, "Print YAML for resources being applied to the cluster without applying them, optional")
```
praveenrewar, line 151: I think the `dry-run` concept is followed from kubectl, but in kubectl, there are 2 types of dry run, server or client, in `client`, the yamls are printed and in `server` the requests are submitted but resources are not persisted. So if we introduce dry-run with only one type (`client`), it might become a bit confusing 🤔 
I think we used `--diff-changes-yaml` in kapp due to the same reason instead of using dry-run, should we do something similar over here as well?

cli/pkg/kctrl/cmd/package/installed/package_install_dry_run.go
```diff
@@ -0,0 +1,147 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+type PackageInstalledDryRun struct {
+	*CreateOrUpdateOptions
+}
+
+func (d PackageInstalledDryRun) PrintResources() error {
+	const yamlSeperator = "---\n"
+	packageInstall := &kcpkgv1alpha1.PackageInstall{
```
praveenrewar, line 21: We seem to be duplicating a lot of code from create_or_update, maybe those sections should be taken out as separate functions to be reused here as well?

cli/pkg/kctrl/cmd/package/installed/package_install_dry_run.go
```diff
@@ -0,0 +1,147 @@
...
+				VersionSelection: &versions.VersionSelectionSemver{
+					Constraints: d.version,
+					Prereleases: &versions.VersionSelectionSemverPrereleases{},
+				},
+			},
+		},
+	}
+
+	rbacResourcesYAML := ""
+	createRBAC := d.serviceAccountName != ""
```
rohitagg2020: nit: Rename the variable. As of now, it seems like variable name is holding the opp. value what it suggests. 
```
createRBAC := d.serviceAccountName == ""
```
and then change the code accordingly below.

Or change to

May be `isServiceAccountNameProvided`.
and then
```
if isServiceAccountNameProvided {
  packageInstall.Spec.ServiceAccountName = d.serviceAccountName
} else {
  // create RBAC
  packageInstall.Spec.ServiceAccountName = d.createdAnnotations.ServiceAccountAnnValue()

  serviceAccount := &corev1.ServiceAccount{
  ...
}
```

cli/test/e2e/package_install_dry_run_test.go
```diff
@@ -0,0 +1,164 @@
...
+  creationTimestamp: null
+  name: test
+  namespace: kctrl-test
+spec:
+  packageRef:
+    refName: test.carvel.dev
+    versionSelection:
+      constraints: 1.0.0
+      prereleases: {}
+  serviceAccountName: test-svc-acc
```
rohitagg2020, line 103: Service Account here is different from line 90 and 93. I am not familiar, why so. Can you please explain.

cli/test/e2e/package_install_dry_run_test.go
```diff
@@ -0,0 +1,164 @@
+package e2e
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestPackageInstallDryRun(t *testing.T) {
```
rohitagg2020, line 10: nit: Should we make it table driven test case?

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
rohitagg2020: maybe, `dry-run=client`??

100mik: I do not see us using `--dry-run=server` anytime soon 🤔 
I did copy over the flag name as suggested in the issue. How do y'all feel about `--resource-yamls`?

cli/pkg/kctrl/cmd/package/installed/package_install_dry_run.go
100mik: That was the original intention, however all the existing code relies on a client. The logic was becoming quite branched for one feature. (Conditionally check if dry run is being used every time we use a client and every time we print something).
The names are derived from the shared `createdAnnotations` resource and I have reused the function managing the annotations added to the pkgi.

Furthermore, since these APIs are stable I see the resource creation logic not evolving a lot over time 🤔

cli/pkg/kctrl/cmd/package/installed/package_install_dry_run.go
100mik: Gotcha will make this change

cli/test/e2e/package_install_dry_run_test.go
100mik: The annotations should not be added at all, gotta take a closer look at the part where we reuse the function 🤔

cli/test/e2e/package_install_dry_run_test.go
100mik: Looks like this is due to the boolean not being appropriately named, fixing that would fix this!

100mik: There are 3 cases with different flags and this list won't grow, so I think it is fine as is. Why do you think it should be table driven?

cli/pkg/kctrl/cmd/package/installed/package_install_dry_run.go
praveenrewar: > The logic was becoming quite branched for one feature

Hmm, yeah I could see a lot of branching required for that. 

> Furthermore, since these APIs are stable I see the resource creation logic not evolving a lot over time

Cool, let's go with it.

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
praveenrewar: Sounds good to me, only concern is that the name doesn't indicate that the resources won't be created, unlike kapp, there's no other way for users to say "no".

100mik: The help section does call it out in the description, would that not suffice?
I am hesitant to use diff run here as we have considered building an actual `diff-run` multiple times and might still do so in the future 🤔

praveenrewar: > The help section does call it out in the description, would that not suffice?

I mean eventually users would know that it doesn't create resources when they run it, but it sounds a bit weird.

> I am hesitant to use diff run here as we have considered building an actual diff-run multiple times and might still do so in the future

This flag would make more sense with something like a diff-run, or maybe we should just go ahead with the original and call it dry-run because that's what it is doing if we only print the resource yamls and not create the resources 😅

cli/test/e2e/package_install_dry_run_test.go
rohitagg2020: I missed the different flags. 

Since we were having yaml as input and the output was compared against yaml, table-driven would have been better if we had the same flags...

cli/pkg/kctrl/cmd/package/installed/create_or_update.go
rohitagg2020: Do you think it will be good to print a line in the output stating that `dry-run` will only print the resource yaml and it won't submit server-side requests without persisting the resource? Basically, print your help description when somebody runs a `dry-run`?

100mik: This would affect how well the output can be piped I think ^
But it is already called out in the command help section and I believe folks expect resources not to be persisted when they are using the flag

100mik: @praveenrewar This can totally evolve over time. I do not see users using this as a part of their automated workflows right away. I think this is something we can think about and be a bit more lenient when it comes to backwards compatibility while we iterate over it.
One step at a time!

cli/pkg/kctrl/cmd/package/installed/package_install_dry_run.go
100mik: Fixed this up

cli/test/e2e/package_install_dry_run_test.go
100mik: Got fixed with the resolution, please do verify the test cases now!

cli/test/e2e/package_install_dry_run_test.go
100mik: Yep agreed! Input however is flag based here

praveenrewar (approved):
LGTM!

PR #1111, opened 2023-02-22
Do not expose development values to kc package bundle

### LINKED ISSUE #1038: Fix kapp-controller package bundle to exclude development values
**What steps did you take:**
Installed the latest kapp-controller package (v0.44.1)

**What happened:**
The values.yml contains development values, such as `version: develop` instead of production values only.

**What did you expect:**
The values.yml should only contain configurable values we expect users to need.

The structure of secretgen-controller folders is something we may want to replicate here as part of this, so that we have clear separation of values based on dev and prod.

https://github.com/vmware-tanzu/carvel-secretgen-controller/tree/develop/config

ghcr.io/vmware-tanzu/carvel-kapp-controller-package-bundle@sha256:9874d2c959d1b8600da37b5798580dd3f6f797512e8da531e31780765db2e59c

Current values.yml can be found here: https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/config/values.yml

We should not expose the following values, SGC has an example set of values we chose to expose, these could help inform the decision in kapp-controller (https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/develop/config/package-bundle/config/values-schema.yml) : 

```
push_images: false
image_cache: true
image_repo: docker.io/k14stest/kapp-controller-test
platform: ""
kapp_controller_version: "develop"
```

**Environment:**

---
Vote on this request

praveenrewar:
- Do not expose development values to kc package bundle
- Structure the config to have a clear separation of values

Fixes #1038 

- [Link](https://github.com/praveenrewar/kapp-controller/releases/tag/v0.60.0) to a release published on my fork using these changes.
- [Diff](https://www.diffchecker.com/ERRvxO1s/) between the latest kc release.yml and the release.yml from the test release respectively. The differences are the expected ones only.


config/release/build.yml
```diff
@@ -8,17 +9,18 @@ sources:
...
-      - #@ "--build-arg=KCTRL_VER="+data.values.kapp_controller_version
-      #@ if/end data.values.platform != "":
-      - #@ "--platform="+data.values.platform
+      - #@ "--build-arg=KCTRL_VER="+data.values.dev.version
+      #@ if/end data.values.dev.platform != "":
+      - #@ "--platform="+data.values.dev.platform
+      #@ end
 
-#@ if/end data.values.push_images:
+#@ if/end data.values.dev.push_images:
```
sethiyash: We can remove this check. Now, as we have separate build.yml for dev and release, and for release we always want to push the image.

package-build.yml
```diff
@@ -15,8 +15,7 @@ spec:
           template:
           - ytt:
               paths:
-              - config
-              - config-release
+              - config/package-bundle/config
```
neil-hickey: Are we able to keep this as `config`?

Why? There are third party consumers of our OSS packages that have other overlays and things in the config folder, and changing this structure would affect them. Ideally we don't need to subdir down to this package-bundle folder.

config/package-bundle/config/values-schema.yml
```diff
@@ -0,0 +1,19 @@
+#@data/values-schema
+---
```
neil-hickey: Hey @praveenrewar so I had a similar conversation with yash about these values.

It would be ideal if we could align them to what TCE were using:

https://github.com/carvel-dev/kapp-controller/pull/1047/files#r1066411323

package-build.yml
praveenrewar: That makes sense. I tried to keep the structure same as secretgen-controller, but it would make more sense to keep it compatible with the existing structure for kc package.

config/package-bundle/config/values-schema.yml
praveenrewar: Sorry, missed that earlier conversation, I will add the other values as well.

package-build.yml
praveenrewar: Done

config/package-bundle/config/values-schema.yml
praveenrewar: I have started working on adding the overlays for these additional values, but I feel this should be part of another follow up PR. What do you think?

hack/dev-deploy.sh
```diff
@@ -18,34 +16,14 @@ fi
 
 echo "got kc latest image: $kc_latest_image"
 
-cat << EOF > tmp/build/Dockerfile
+cat << EOF > Dockerfile.dev
 FROM ${kc_latest_image} AS build
 FROM scratch
 COPY --from=build / /
```
praveenrewar, line 22: For some reason, I get 
`
ERROR: failed to solve: kbld:kapp-controller-sha256-37dd1f0e39e3fc238ca0f80dae5f9d0b7f1765035b0348731423560c1459cddb: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
` at this step. The same is happening on develop branch so not sure if it's due to the changes I introduced.
I tried removing this step and the deployment worked fine, but we need to copy stuff from `build` so we do need this step.

praveenrewar: Seems to be working after restarting my laptop.

neil-hickey (approved):
LGTM!

PR #1113, opened 2023-02-26
`kctrl`: Flag to create namespace when adding new repo

### LINKED ISSUE #1001: `kctrl`: When adding a package repo, include option to create the namespace if not existing
When I add a Carvel package repository to a cluster, I first need to create a dedicated namespace explicitly.

1. `kubectl create namespace kadras-packages`
2. `kctrl package repository add -r kadras-repo --url ghcr.io/arktonix/kadras-packages:0.5.1 -n kadras-packages`

I'd like to skip the explicit creation of the namespace and have it created as part of the `kctrl package` command, similar to how the `helm` CLI works. For example, the functionality could be enabled via a `--create-namespace` flag.

`kctrl package repository add -r kadras-repo --url ghcr.io/arktonix/kadras-packages:0.5.1 -n kadras-packages --create-namespace`

One of the reasons for this feature request is to make the onboarding experience even simpler when introducing people to Carvel packaging or when using a Carvel metapackage as the foundation of projects/tutorials relying on a Carvel-based platform.

With this feature, that would result in three operations rather than four:

1. deploy kapp-controller
2. add the package repository
3. install the package

---
Vote on this request

ThomasVitale:

When adding a new package repository to a cluster, it's now possible to create the installation namespace automatically by specifying the "--create-namespace" flag.

Fixes gh-1001

```release-note
Added a new flag `create-namespace` for creating the installation namespace when adding a new package repository to a cluster.
```

After a first review of this pull request to validate the behaviour of the new flag, I will submit a PR to update the documentation on the project website.

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/cli/pkg/kctrl/cmd/package/repository/add_or_update.go
+++ b/cli/pkg/kctrl/cmd/package/repository/add_or_update.go
@@ -161,11 +161,11 @@ func (o *AddOrUpdateOptions) Run(args []string) error {
 
 		_, err = coreClient.CoreV1().Namespaces().Create(context.Background(), namespace, metav1.CreateOptions{})
 		if err != nil {
-			if errors.IsAlreadyExists(err) {
-				o.statusUI.PrintMessagef("The namespace '%s' already exists", o.NamespaceFlags.Name)
-			} else {
+			if !errors.IsAlreadyExists(err) {
 				return err
 			}
+		} else {
+			o.statusUI.PrintMessagef("Created namespace '%s'", o.NamespaceFlags.Name)
 		}
 	}
 
--- a/cli/pkg/kctrl/cmd/package/repository/add_or_update.go
+++ b/cli/pkg/kctrl/cmd/package/repository/add_or_update.go
@@ -19,6 +19,7 @@ import (
 	kcpkg "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
 	kcclient "github.com/vmware-tanzu/carvel-kapp-controller/pkg/client/clientset/versioned"
 	versions "github.com/vmware-tanzu/carvel-vendir/pkg/vendir/versions/v1alpha1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -33,6 +34,7 @@ type AddOrUpdateOptions struct {
 	SecureNamespaceFlags cmdcore.SecureNamespaceFlags
 	Name                 string
 	URL                  string
+	CreateNamespace      bool
 
 	CreateRepository bool
 
@@ -71,6 +73,8 @@ func NewAddCmd(o *AddOrUpdateOptions, flagsFactory cmdcore.FlagsFactory) *cobra.
 
 	// TODO consider how to support other repository types
 	cmd.Flags().StringVar(&o.URL, "url", "", "OCI registry url for package repository bundle (required)")
+
+	cmd.Flags().BoolVar(&o.CreateNamespace, "create-namespace", false, "Create the package repository namespace if not present (default false)")
 
 	o.WaitFlags.Set(cmd, flagsFactory, &cmdcore.WaitFlagsOpts{
 		AllowDisableWait: true,
@@ -139,6 +143,24 @@ func (o *AddOrUpdateOptions) Run(args []string) error {
 	client, err := o.depsFactory.KappCtrlClient()
 	if err != nil {
 		return err
+	}
+
+	if o.CreateNamespace {
+		coreClient, err := o.depsFactory.CoreClient()
+		if err != nil {
+			return err
+		}
+
+		namespace := &v1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: o.NamespaceFlags.Name,
+			},
+		}
+
+		_, err = coreClient.CoreV1().Namespaces().Create(context.Background(), namespace, metav1.CreateOptions{})
+		if err != nil {
+			o.ui.PrintLinef("The namespace %s already exists", o.NamespaceFlags.Name)
+		}
 	}
 
 	existingRepository, err := client.PackagingV1alpha1().PackageRepositories(o.NamespaceFlags.Name).Get(
--- a/cli/pkg/kctrl/cmd/package/repository/add_or_update.go
+++ b/cli/pkg/kctrl/cmd/package/repository/add_or_update.go
@@ -19,7 +19,7 @@ import (
 	kcpkg "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
 	kcclient "github.com/vmware-tanzu/carvel-kapp-controller/pkg/client/clientset/versioned"
 	versions "github.com/vmware-tanzu/carvel-vendir/pkg/vendir/versions/v1alpha1"
-	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -146,12 +146,14 @@ func (o *AddOrUpdateOptions) Run(args []string) error {
 	}
 
 	if o.CreateNamespace {
+		o.statusUI.PrintMessagef("Creating namespace '%s'", o.NamespaceFlags.Name)
+
 		coreClient, err := o.depsFactory.CoreClient()
 		if err != nil {
 			return err
 		}
 
-		namespace := &v1.Namespace{
+		namespace := &corev1.Namespace{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: o.NamespaceFlags.Name,
 			},
@@ -159,7 +161,11 @@ func (o *AddOrUpdateOptions) Run(args []string) error {
 
 		_, err = coreClient.CoreV1().Namespaces().Create(context.Background(), namespace, metav1.CreateOptions{})
 		if err != nil {
-			o.ui.PrintLinef("The namespace %s already exists", o.NamespaceFlags.Name)
+			if errors.IsAlreadyExists(err) {
+				o.statusUI.PrintMessagef("The namespace '%s' already exists", o.NamespaceFlags.Name)
+			} else {
+				return err
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


cli/pkg/kctrl/cmd/package/repository/add_or_update.go
```diff
@@ -141,6 +145,24 @@ func (o *AddOrUpdateOptions) Run(args []string) error {
...
+
+		namespace := &v1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: o.NamespaceFlags.Name,
+			},
+		}
+
+		_, err = coreClient.CoreV1().Namespaces().Create(context.Background(), namespace, metav1.CreateOptions{})
+		if err != nil {
+			o.ui.PrintLinef("The namespace %s already exists", o.NamespaceFlags.Name)
```
praveenrewar: We should probably return the error if it's *not* of type [IsAlreadyExists](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsAlreadyExists) and ignore if it already exists.

cli/pkg/kctrl/cmd/package/repository/add_or_update.go
```diff
@@ -19,6 +19,7 @@ import (
 	kcpkg "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
 	kcclient "github.com/vmware-tanzu/carvel-kapp-controller/pkg/client/clientset/versioned"
 	versions "github.com/vmware-tanzu/carvel-vendir/pkg/vendir/versions/v1alpha1"
+	v1 "k8s.io/api/core/v1"
```
100mik: nit: lets have a more explicit name
```suggestion
	corev1 "k8s.io/api/core/v1"
```

cli/pkg/kctrl/cmd/package/repository/add_or_update.go
100mik: +1 to Praveens comment. 
We do something similar over [here](https://github.com/carvel-dev/kapp-controller/blob/develop/cli/pkg/kctrl/cmd/package/installed/create_or_update.go#L585-L595)

However, in the example we attempt an `update` if the service already exists. Where as here, we return the error if the error is `!IsAlreadyExists()`

cli/pkg/kctrl/cmd/package/repository/add_or_update.go
```diff
@@ -141,6 +145,24 @@ func (o *AddOrUpdateOptions) Run(args []string) error {
...
+		if err != nil {
+			return err
+		}
+
+		namespace := &v1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: o.NamespaceFlags.Name,
+			},
+		}
+
```
100mik, line 159: Let's add a log here saying that we are going to create a namespace. Something like `Creating namespace 'asdf'`.
Let's use `o.statusUI.PrintMessagef()` for this. (Adds a nice little timestamp in front of the message)

I think the message we are printing when namespace already exists should use `statusUI` as well 🤔

100mik (commented):
Some thoughts!

In addition to these looks like there is a redundant log line being printed on line 187
(More specifically: `o.ui.PrintLinef("Waiting for package repository to be added")`)
That can be removed as a part of this PR but does not block the PR itself!

cli/pkg/kctrl/cmd/package/repository/add_or_update.go
100mik: "here" being right before we make the call to create it

cli/pkg/kctrl/cmd/package/repository/add_or_update.go
ThomasVitale: Good point! I changed it as suggested.

cli/pkg/kctrl/cmd/package/repository/add_or_update.go
ThomasVitale: Updated

cli/pkg/kctrl/cmd/package/repository/add_or_update.go
ThomasVitale: Done. I have used `statusUI` for both messages.

praveenrewar: @100mik @ThomasVitale Instead of saying `Creating namespace xyz` and `Namespace xyz already exists` every time (thinking about updated scenarios), should we say `Created namespace xyz` once when we create the namespace?

(No strong opinions)

cli/pkg/kctrl/cmd/package/repository/add_or_update.go
```diff
@@ -141,6 +145,30 @@ func (o *AddOrUpdateOptions) Run(args []string) error {
...
+		_, err = coreClient.CoreV1().Namespaces().Create(context.Background(), namespace, metav1.CreateOptions{})
+		if err != nil {
+			if errors.IsAlreadyExists(err) {
+				o.statusUI.PrintMessagef("The namespace '%s' already exists", o.NamespaceFlags.Name)
+			} else {
+				return err
+			}
```
praveenrewar:
```suggestion
			if !errors.IsAlreadyExists(err) {
			         return err
			} 
			o.statusUI.PrintMessagef("The namespace '%s' already exists", o.NamespaceFlags.Name)
```

cli/pkg/kctrl/cmd/package/repository/add_or_update.go
ThomasVitale: I like that! I have updated the code to only print a message if the namespace gets created.

cli/pkg/kctrl/cmd/package/repository/add_or_update.go
ThomasVitale: I have refactored that part to print only a message if the namespace is created as suggested in the previous comment.

cli/pkg/kctrl/cmd/package/repository/add_or_update.go
```diff
@@ -141,6 +145,30 @@ func (o *AddOrUpdateOptions) Run(args []string) error {
 		return err
 	}
 
+	if o.CreateNamespace {
+		o.statusUI.PrintMessagef("Creating namespace '%s'", o.NamespaceFlags.Name)
```
praveenrewar: Should we remove this line as well? Or you already did and forgot to push the change?

100mik: I think I am good with the changes, once this is resolved ^

cli/test/e2e/package_repository_test.go
```diff
@@ -127,4 +127,21 @@ func TestPackageRepository(t *testing.T) {
 		require.Exactly(t, expectedOutputRows, output.Tables[0].Rows)
 	})
 
+	logger.Section("creating a repository in a new namespace that doesn't exist", func() {
```
100mik, line 133: Just one change, can we add deletion of these two namespaces to the [cleanup function](https://github.com/carvel-dev/kapp-controller/pull/1113/files#diff-09f0c0ae3712103375872b02f78bf3bc30f715f8386e471d6235b5cf54d3e349R25)?
Maybe move the variables storing the names as well so that we can do that, something like `existingRepoNamesapce` and `newRepoNamespace` maybe.

cli/pkg/kctrl/cmd/package/repository/add_or_update.go
ThomasVitale: Thanks for catching this. You'd think after several commits I'd learn the lesson about signing off commits, but I keep missing that and had to rebase (my Git is configured with the `-S` parameter for signing, but the DCA sign-off is with `-s`). I guess while doing that, this message popped up again.

cli/test/e2e/package_repository_test.go
ThomasVitale: I'm trying doing this, but the tests are failing. I couldn't find a way to run the tests locally, so the troubleshooting is a bit slow.

ThomasVitale: What do you think about this solution? https://github.com/carvel-dev/kapp-controller/pull/1113/commits/82dac38b2d06fbb4fadca77bb169dfa60a992440

100mik: Sorry! Were we already deleting the namespaces in the Cleanup function? It looks like we are removing those bits in this commit. If that is the case we do not need this change

ThomasVitale: From what I understood, the cleanup function is removing the package and related namespace in all test cases except the 2 new tests added because I use different namespaces. So I added the explicit cleanup for those 2 new test cases to delete both package and namespace since they are not caught by the cleanup function.

100mik: Ah I see, I was wondering if we could be deleting the new namespaces in the cleanup function itself
Mainly, because a namespace left over from an interrupted test run could affect these tests.

The commit you linked seemed to be removing the deletion of those namespaces from the cleanup function which had me a bit confused.

Ideally : We would have the commands deleting the namespaces in the cleanup function itself, so that we make sure they do not exist before and after running the tests!
What do you think about that?

100mik: (To put it concisely, I think the RemoveClusterResource call for these two namespaces should be within the cleanup function)

100mik (approved):
Changes look good to me! Thank you so much for contributing

@praveenrewar feel free to merge this away if it looks good to you 🚀

cli/test/e2e/package_repository_test.go
```diff
@@ -127,4 +132,18 @@ func TestPackageRepository(t *testing.T) {
...
+		kappCtrl.Run([]string{"package", "repository", "add", "-r", pkgrName, "--url", pkgrURL, "-n", newRepoNamespace, "--create-namespace"})
+
+		kubectl.Run([]string{"get", kind, pkgrName, "-n", newRepoNamespace})
+	})
+
+	logger.Section("creating a repository in a namespace that already exists", func() {
+		kubectl.Run([]string{"create", "namespace", existingRepoNamespace})
```
praveenrewar: We could probably just use env.Namespace instead of this. It would remove the overhead of creating/deleting one namespace.

praveenrewar: Sorry didn't take a look at the tests earlier 😅

ThomasVitale: I considered doing that, but then opted for creating it explicitly to make the test case resemble the current 2-steps way of adding a repo as opposed to the changes I made in the PR (1. create the namespace, 2. add a repo). But I can change it to use the general one, if you want :)

ThomasVitale: Updated the tests here to re-use env.Namespace: https://github.com/carvel-dev/kapp-controller/pull/1113/commits/893f9e7beb0e8faa1892be52c9a07e706e7e8002

praveenrewar (approved):
LGTM! Thank you so much for the contribution 🚀

ThomasVitale:
@100mik @praveenrewar thanks a lot for your guidance and review

PR #1114, opened 2023-02-26
Fix flaky tests

praveenrewar:
* Use token with setup-action in kctrl tests as we were experiencing rate limiting.
* Use token in `dependencies_test.go` to avoid rate limiting. Currently we only use it to download the dependencies, but not while testing dependency download.
* Remove Job from the resources used in `TrustConfig_TrustCACerts`, we have readinessProbe defined for the deployment which should make sure that the service is ready to accept request.  
  - For some reason the Job which is supposed to make sure the same doesn't work. For now, removing the Job seems to have resolved the flakiness (tried running the tests ~10 times with these changes), but even if we face the same issue again we should be able to look at the error from the App CR and try to have a fix for that.

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/.github/workflows/release-process.yml
+++ b/.github/workflows/release-process.yml
@@ -23,6 +23,7 @@ jobs:
         uses: carvel-dev/setup-action@v1
         with:
           only: imgpkg
+          token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v2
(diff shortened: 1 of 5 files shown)


neil-hickey:
Nice! This is great, ty @praveenrewar ! Don't forget to remove the WIP from the title, but when that's done, happy to merge away 🥳

neil-hickey (approved):

praveenrewar:
> Don't forget to remove the WIP from the title

Done. I will create more PR(s) for other tests.

PR #1115, opened 2023-02-27
Show target cluster in the list of apps

wolfedale:
This is for the issue: https://github.com/carvel-dev/kapp-controller/issues/1103

```release-note
"NONE"
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/config/crds.yml
+++ b/config/crds.yml
@@ -734,6 +734,10 @@ spec:
       jsonPath: .status.deploy.startedAt
       name: Since-Deploy
       type: date
+    - description: Target Cluster name taken from the kubeconfig.
+      jsonPath: .spec.cluster.kubeconfigSecretRef.name
+      name: TargetCluster
+      type: string
     - description: Time since creation
       jsonPath: .metadata.creationTimestamp
       name: Age
--- a/pkg/apis/kappctrl/v1alpha1/types.go
+++ b/pkg/apis/kappctrl/v1alpha1/types.go
@@ -13,6 +13,7 @@ import (
 // +kubebuilder:resource:categories={carvel}
 // +kubebuilder:printcolumn:name=Description,JSONPath=.status.friendlyDescription,description=Friendly description,type=string
 // +kubebuilder:printcolumn:name=Since-Deploy,JSONPath=.status.deploy.startedAt,description=Last time app started being deployed. Does not mean anything was changed.,type=date
+// +kubebuilder:printcolumn:name=TargetCluster,JSONPath=.spec.cluster.kubeconfigSecretRef.name,description=Target Cluster name taken from the kubeconfig.,type=string
 // +kubebuilder:printcolumn:name=Age,JSONPath=.metadata.creationTimestamp,description=Time since creation,type=date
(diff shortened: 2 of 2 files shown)


config/crds.yml
```diff
@@ -734,6 +734,10 @@ spec:
       jsonPath: .status.deploy.startedAt
       name: Since-Deploy
       type: date
+    - description: Target Cluster name taken from the kubeconfig.
+      jsonPath: .spec.cluster.kubeconfigSecretRef.name
+      name: TargetCluster
```
praveenrewar, line 739:
```suggestion
      name: Target-Cluster
```

praveenrewar: Does it make sense to have a default value, something like `<none>` (if possible)? Or is it okay to have an empty column when target cluster is not used?
cc @neil-hickey @100mik What are your thoughts?

neil-hickey: Default ideally is the name of "this" cluster. i.e the one it's running on.

praveenrewar: ^ Makes sense 💯 , it would be good to have the target-cluster populated with the current cluster by default, but, is `.spec.cluster.kubeconfigSecretRef.name` populated by default?

wolfedale: This would require some code changes or it's about adding something like `// +kubebuilder:default:="this"` ?

100mik: I imagine it would involve adding a `cluster` field to the `status` which is populated on reconcile. I do not imagine the controller populates any of the `spec` fields 🤔

100mik: I definitely agree with Neil in regards to what should the default value be though!

neil-hickey:
Just following up here, we had a chat about this at our community meeting this week. Generally everyone was onboard with the idea of `current` / `this` as a default rather than blank as that might indicate to the operator that something is still going on.

The current approach is to use the kubeconfig name for the external cluster. That might be a bit weird if you kubeconfig name doesn't actually mention the cluster or has other stuff in it. Ideally this is actually the name of the cluster from within the kubeconfig (kubeconfig's can also have multiple cluster info within them)

vrabbi:
I think another option instead of empty that makes sense is in-cluster.
Regarding the name, i agree the secret name is not optimal. Best would be to get the clustername of the current context set in the kubeconfig eithin the secret. This adds logic that would be needed but would make the UX much better.
Another option that could be looked at is similar to argo where the target field is actually the clusters api endpoint. So for local you would have kubernetes.default.svc.cluster.local and for others you would parse the api endpoint from the kubeconfig. Names may be more user friendly if naming is done well, but api endpoint is definitive and is the actual target.

wolfedale:
How about creating another field for `ClusterTarget` and check if `.spec.cluster.` contain data. If something is there, we can read and parse `.spec.cluster.kubeconfigSecretRef.`, find selected cluster name and add it to the `ClusterTarget` field. If not, add `in-cluster` or `this` to it. Does it make sense?

wolfedale:
Any comment? :-)

PR #1120, opened 2023-03-03
Expose values in kapp-controller package

praveenrewar:
Expose values in kapp-controller package. [Release ](https://github.com/praveenrewar/kapp-controller/releases/tag/v0.76.2)on fork.

Fixes #

```release-note
- Expose more values in kc package.
- Breaking change: Use lowerCamelCase for existing values.
```

##### Review Checklist:

  a link to that PR
  change

```docs

```


config/controller-config-overlay.yml
```diff
@@ -0,0 +1,37 @@
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:data", "data")
+
+#@ def is_config_empty(config):
```
praveenrewar: I wanted to make sure that we don't make any changes to our releases. Is there a better way of checking if all the values are empty?

config/values-schema.yml
```diff
@@ -17,3 +17,57 @@ dangerous_enable_pprof: false
...
 #@schema/desc "API port"
 api_port: 10350
+#@schema/desc "The coreDNSIP will be injected into /etc/resolv.conf of kapp-controller pod"
+coreDNSIP: ""
+#@schema/desc "HostNetwork of kapp-controller deployment."
+hostNetwork: false
+#@schema/desc "PriorityClassName of kapp-controller deployment."
+priorityClassName: ""
+#@schema/desc "Concurrency of kapp-controller deployment"
+concurrency: 0
```
praveenrewar: I used 0 to not make any change to our current release artefact. But this would also result in a default value of 0 for concurrency, which would be weird.

config/update-strategy-overlay.yml
```diff
@@ -0,0 +1,63 @@
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:data", "data")
+
+#@ def matcher():
```
praveenrewar: I tried to cover all the values from [here](https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/kapp-controller/0.44.1/bundle/config/values.yaml), but should we include these or expect users to add overlays for these?

neil-hickey (approved):
Nice nice, thanks for this @praveenrewar !

For future us... we opted for renaming the existing values to use the lowerCamelCase so as not to have two different styles of keys. Given on March 8th, we had one download of the kc bundle over the last month. We agreed that this was better to break now while noone is using it :D

praveenrewar:
While testing the kc package, I noticed that I had used 
```
- #@ "-metricsBindAddress={}".format(data.values.metricsBindAddress)
``` 
instead of 
```
- #@ "-metrics-bind-address={}".format(data.values.metricsBindAddress)
``` 
I have updated it now. 
@neil-hickey Do you want to take a look at it again or I can go ahead and merge it?

rohitagg2020:
LGTM

PR #1145, opened 2023-03-21
Include required properties in `--default-values-file-output`

praveenrewar:
Include required properties in --default-values-file-output

Fixes #

```release-note
`--default-values-file-output` will also include required properties.
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/cli/hack/test-e2e.sh
+++ b/cli/hack/test-e2e.sh
@@ -1,10 +1,17 @@
 #!/bin/bash
 
-set -e -x -u
+set -e -x -o pipefail
 
 go clean -testcache
 
 export KCTRL_BINARY_PATH="${KCTRL_BINARY_PATH:-$PWD/kctrl}"
+
+if [ -z "$KCTRL_E2E_NAMESPACE" ]; then
+    echo "setting e2e namespace to: kctrl-test";
+    export KCTRL_E2E_NAMESPACE="kctrl-test"
+fi
+# create ns if not exists because the `apply -f -` won't complain on a no-op if the ns already exists.
+kubectl create ns $KCTRL_E2E_NAMESPACE --dry-run=client -o yaml | kubectl apply -f -
 
 go test ./test/e2e/ -timeout 60m -test.v $@
 
--- a/cli/pkg/kctrl/cmd/package/available/default_values.go
+++ b/cli/pkg/kctrl/cmd/package/available/default_values.go
@@ -49,7 +49,7 @@ func (s PackageSchema) DefaultValues() ([]byte, error) {
 		return nil, err
 	}
 
-	return s.commentDefaultValues(b), nil
+	return []byte(s.commentDefaultValues(s.removeEmptyMaps(b.String()))), nil
 }
 
 // schemaDefault does defaulting of x depending on default values in s.
@@ -61,6 +61,11 @@ func (s PackageSchema) DefaultValues() ([]byte, error) {
 func (s PackageSchema) schemaDefault(x interface{}, ss *structuralschema.Structural) {
 	if ss == nil {
 		return
+	}
+
+	requiredProperties := map[string]bool{}
+	for _, prop := range ss.ValueValidation.Required {
+		requiredProperties[prop] = true
 	}
 
 	switch x := x.(type) {
@@ -79,7 +84,7 @@ func (s PackageSchema) schemaDefault(x interface{}, ss *structuralschema.Structu
 					}
 				}
 
-				if shouldCreateDefault {
+				if shouldCreateDefault || requiredProperties[k] {
 					prop.Default.Object = make(map[string]interface{})
 				} else {
 					continue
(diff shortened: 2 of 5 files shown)


cli/test/e2e/env.go
```diff
@@ -44,6 +44,14 @@ func (e Env) Validate(t *testing.T) {
 		errStrs = append(errStrs, "Expected Namespace to be non-empty (hint: kubectl create namespace kctrl-test; export KCTRL_E2E_NAMESPACE=kctrl-test)")
 	}
 
+	if len(errStrs) > 0 {
+		t.Fatalf("%s", strings.Join(errStrs, "\n"))
+	}
+}
+
+func (e Env) ValidateImageURL(t *testing.T) {
+	errStrs := []string{}
```
100mik: I think we can do without errStrs here because it won't be checked with the other parameters anyway.

cli/pkg/kctrl/cmd/package/available/default_values.go
```diff
@@ -138,28 +137,36 @@ func (s PackageSchema) isKindInt(src interface{}) bool {
...
+	if structural.Type != "object" {
+		return append(b, true)
+	}
 	for _, v := range structural.Properties {
-		// return true if there is a non-nested(not object) with a default value
-		if v.Type != "object" && v.Default.Object != nil {
-			b = append(b, true)
-			return b
+		// return true if there is a non-nested(not object)
+		if v.Type != "object" {
```
100mik: Are we not able to check if the non-object field is required at this point?

cli/test/e2e/release_with_changed_spec_test.go
```diff
@@ -21,6 +21,7 @@ func TestPackageReleaseWithChangedSpec(t *testing.T) {
 	env := BuildEnv(t)
 	logger := Logger{}
 	kappCtrl := Kctrl{t, env.Namespace, env.KctrlBinaryPath, logger}
+	env.ValidateImageURL(t)
```
100mik: not sure about running validations on image env like this, mainly because the user will only hit this when this tests is run. Should we move this to a separate PR? Setting an image that is not used does not seem that much of an overhead 🤔

cli/test/e2e/env.go
praveenrewar: Reverted this change.

cli/test/e2e/release_with_changed_spec_test.go
praveenrewar: Reverted this change.

cli/pkg/kctrl/cmd/package/available/default_values.go
praveenrewar: There's no required field on the properties. The required field information is available separately on the values schema.

100mik: The OpenAPI schema does reflect which fields are required right? 🤔 
(Or is it always false)

praveenrewar: This is how it's defined:
```
valuesSchema:
    openAPIv3:
      properties:
        testField:
          description: The test field
          type: string
        requiredField1:
          description: The required field
          type: string
        requiredField2:
          description: Another required field
          type: string
      required:
      - requiredField1
      - requiredField2
```

100mik: Correct, so should we not emit only the ones that are in the required list? 
(I thought that was the intention)

praveenrewar: That wasn't the intention of the PR, but do you think we should take that path?
I felt that it would further complicate the code.

100mik: I wonder how it affects larger packages with a decent number of fields that are generally unused

cli/hack/test-e2e.sh
```diff
@@ -1,11 +1,18 @@
 #!/bin/bash
 
-set -e -x -u
+set -e -x -o pipefail
```
100mik, line 3: Just to be sure, we are removing `-u` because we do not want to error out if ns env variable is not set correct?

cli/pkg/kctrl/cmd/package/available/default_values.go
```diff
@@ -32,6 +36,10 @@ func (s PackageSchema) DefaultValues() ([]byte, error) {
 		return nil, err
 	}
 
+	if len(ss.Properties) == 0 {
```
100mik, line 39: nit: refactor: `ss` => `structuralSchema`

100mik (commented):
Some thoughts.

Since this PR has a few refactors/improvements separated by commits I was wondering if it would be worth cleaning up some of the nomenclature in `default_values.go`

The nit mentioned aside I think we should rename the slice of booleans (`b`) and even `x` (this seems to have different meanings in different scopes?)
`k` = `key` is the conclusion I drew , but maybe this could have a better name too?

My main concern here is legibility, took me a bit to figure out what b is in context of `createDefault`

cli/hack/test-e2e.sh
praveenrewar: Yep, that's correct.

praveenrewar:
> Since this PR has a few refactors/improvements separated by commits I was wondering if it would be worth cleaning up some of the nomenclature in default_values.go

I have done the refactoring in the bash script only and tried to keep the changes in the other files to a minimal. I wanted to do more refactoring in `default-values.go` but I feel that a separate PR would make more sense?

100mik:
I think I am fine with a separate PR for this stuff
Track this somewhere maybe?

100mik (approved):
LGTM!

praveenrewar:
> Track this somewhere maybe?

I will be creating a new PR soon.

PR #1208, opened 2023-05-19
Unblock app deletion when ns is terminating and resources are in the same ns only

### LINKED ISSUE #1132: Enable kapp-controller to delete App when namespace is deleted
**Description**
Deleting namespace gets stuck in `Terminating` state as kapp-controller won't delete an application without valid service account

**What steps did you take:**

* Created a minikube cluster on local machine (I used k8s version `1.26.1`)
* Installed kapp-controller using:
```
kubectl apply -f https://github.com/carvel-dev/kapp-controller/releases/download/v0.44.6/release.yml
```
* Create a `dev` namespace
```
kubectl create ns dev
```
* Create a sample config map to use later while creating an app, and a serviceaccount and rolebindings
```
kubectl apply -f https://raw.githubusercontent.com/atmandhol/tap-nsp-gitops/main/random/app/app-permissions.yaml
kubectl apply -f https://raw.githubusercontent.com/atmandhol/tap-nsp-gitops/main/random/app/app-config.yaml
```
* Create an `App`
```
kubectl apply -f https://raw.githubusercontent.com/atmandhol/tap-nsp-gitops/main/random/app/app.yaml
```
I now have an `App` that is successfully deployed.

Next, I attempt deletion of the namespace as I no longer need it
```
kubectl delete ns dev
```

**What happened:**
When I ran the delete command
* ServiceAccount `default-ns-sa` that I was using for the app is deleted as part of the namespace deletion
* Namespace `dev` gets stuck in the `Terminating` state until the `App` is deleted.
* `App` gets stuck and fails to delete with the following error
```
Preparing kapp: Getting service account: serviceaccounts "default-ns-sa" not found
```

**What did you expect:**
If the namespace in which the `App` resides is in `Terminating` state, kapp-controller should delete the `App` and should not need the `ServiceAccount` that the `App` was using as the user/some process is clearly signaling that it is not needed (because user/some process deleted the namespace).

In large scale kubernetes deployments where there are developer portals or some automated processes that creates and deletes namespaces for users, It's not going to be feasible for operators to manually inspect namespaces and delete all existing `Apps` before deleting the namespace itself.

**Environment:**

- kapp Controller version `0.44.6`
- Kubernetes version `1.26.1`

---
Vote on this request

praveenrewar:
Unblock app deletion when ns is terminating and resources are in the same ns only. This partially fixes #1132.

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -15,6 +15,7 @@ import (
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/metrics"
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/reftracker"
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/template"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/yaml"
 )
@@ -24,6 +25,7 @@ type ComponentInfo interface {
 	KappControllerVersion() (semver.Version, error)
 	KubernetesVersion(serviceAccountName string, specCluster *v1alpha1.AppCluster, objMeta *metav1.ObjectMeta) (semver.Version, error)
 	KubernetesAPIs() ([]string, error)
+	NamespaceStatus(name string) (v1.NamespaceStatus, error)
 }
 
 type Hooks struct {
@@ -289,4 +291,12 @@ func (a App) HasImageOrImgpkgBundle() bool {
 		}
 	}
 	return false
+}
+
+func (a App) isNamespaceTerminating() bool {
+	status, err := a.compInfo.NamespaceStatus(a.app.Namespace)
+	if err != nil {
+		a.log.Error(err, "Error getting app namespace status", "app", a.app.Name, "namespace", a.app.Namespace)
+	}
+	return status.Phase == v1.NamespaceTerminating
 }
--- a/pkg/app/app_reconcile_test.go
+++ b/pkg/app/app_reconcile_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/metrics"
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/template"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	k8sfake "k8s.io/client-go/kubernetes/fake"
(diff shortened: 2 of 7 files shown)


pkg/app/app.go
```diff
@@ -59,15 +63,19 @@ type App struct {
 	pendingStatusUpdate   bool
 	flushAllStatusUpdates bool
 	metadata              *deploy.Meta
+
+	coreClient kubernetes.Interface
```
neil-hickey: We shouldn't need the coreClient in our App, either one of the factories ought to handle interactions with the coreclient, or our compInfo struct. But we want to avoid passing the coreClient around to every newApp instance.

pkg/app/app.go
```diff
@@ -290,3 +298,9 @@ func (a App) HasImageOrImgpkgBundle() bool {
 	}
 	return false
 }
+
+func (a App) isNamespaceTerminating() bool {
```
neil-hickey, line 296: The simplest thing is probably to move this to ComponentInfo. The name doesn't fully reflect it's intention. It's intention is to abstract away calls to the cluster for things outside the scope of the factories. It was introduced so that we don't pass the coreClient around in this way.

pkg/app/app_reconcile_test.go
```diff
@@ -51,7 +51,7 @@ func Test_NoInspectReconcile_IfNoDeployAttempted(t *testing.T) {
 	tmpFac := template.NewFactory(k8scs, fetchFac, false, exec.NewPlainCmdRunner())
 	deployFac := deploy.NewFactory(k8scs, kubeconfig.NewKubeconfig(k8scs, log), nil, exec.NewPlainCmdRunner(), log)
 
-	crdApp := NewCRDApp(&app, log, appMetrics, kappcs, fetchFac, tmpFac, deployFac, FakeComponentInfo{}, Opts{MinimumSyncPeriod: 30 * time.Second})
+	crdApp := NewCRDApp(&app, log, appMetrics, kappcs, fetchFac, tmpFac, deployFac, FakeComponentInfo{}, k8scs, Opts{MinimumSyncPeriod: 30 * time.Second})
```
neil-hickey: all these changes in the test file go away if don't need to pass the coreClient around

pkg/app/app.go
praveenrewar: This makes so much sense! I was trying to find a way of not having to pass the coreClient. Thank you!!

pkg/app/app_reconcile_test.go
praveenrewar: 💯

pkg/app/app.go
praveenrewar: Done

pkg/componentinfo/component_info.go
```diff
@@ -83,6 +85,10 @@ func (ci *ComponentInfo) KubernetesAPIs() ([]string, error) {
 	return metav1.ExtractGroupVersions(groups), nil
 }
 
+func (ci *ComponentInfo) Namespace(name string) (*v1.Namespace, error) {
```
praveenrewar: Does getting the complete namespace make sense?

Also, should I add a comment to get rid of the golint error or is there another way of getting rid of it because the comment would be just adding some noise and not helping much.

neil-hickey: I think the noise is ok, go ahead and add the comment.

> Does getting the complete namespace make sense?

Probably not for this PR? I think we just want the status. We can always refactor later when we have the usecase to need the whole NS object.

neil-hickey (approved):
LGTM, once the linter is happy and the small namespace change is made.

pkg/componentinfo/component_info.go
praveenrewar: Done.

PR #1220, opened 2023-05-29
Don't fallback to automatic noopDelete if cluster is set

praveenrewar:
When namespace is terminating
- Don't fallback to automatic noopDelete if cluster is set
- Check if the number of namespaces from kapp is 1 and is also the same where the App resides

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```


neil-hickey (approved):
lgtm

PR #1295, opened 2023-08-16
Check for exactly one package and one metadata resource

### LINKED ISSUE #1294: [kctrl] Declaring `package-resources.yaml` with missing "Package" resource causes index out of range error
**What steps did you take:**
* Create a `package-resources.yaml` file with `PackageMetadata` and `PackageInstall` resource
* Run `kctrl package release`

**What happened:**
Fails with unhelpful stacktrace

**What did you expect:**
Proper error message indicating an issue with my `package-resources.yaml`

The issue appears to be a bad conditional at https://github.com/carvel-dev/kapp-controller/blob/develop/cli/pkg/kctrl/cmd/package/release/release.go#L93. The left and right of conditional is both checking for exactly one metadata resource, rather than exactly one metadata and one package resource.

**Environment:**
0.46.1 of `kctrl`

---
Vote on this request

andrew-su:
Having the correct conditional here prevents an index out of range error further down the [code](https://github.com/carvel-dev/kapp-controller/blob/665a9d0d514a912c490492614c686c84cb29cd05/cli/pkg/kctrl/cmd/package/release/release.go#L134)

Fixes #1294

### DIFF
--- a/cli/pkg/kctrl/cmd/package/release/release.go
+++ b/cli/pkg/kctrl/cmd/package/release/release.go
@@ -90,7 +90,7 @@ func (o *ReleaseOptions) Run() error {
 	if err != nil {
 		return err
 	}
-	if len(pkgConfigs.PkgMetadatas) != 1 || len(pkgConfigs.PkgMetadatas) != 1 {
+	if len(pkgConfigs.Pkgs) != 1 || len(pkgConfigs.PkgMetadatas) != 1 {
 		return fmt.Errorf("Reading package-resource.yml: file malformed. (hint: delete the file and run `package init` again)")
 	}
 


cli/pkg/kctrl/cmd/package/release/release.go
```diff
@@ -90,7 +90,7 @@ func (o *ReleaseOptions) Run() error {
 	if err != nil {
 		return err
 	}
-	if len(pkgConfigs.PkgMetadatas) != 1 || len(pkgConfigs.PkgMetadatas) != 1 {
+	if len(pkgConfigs.Pkgs) != 1 || len(pkgConfigs.PkgMetadatas) != 1 {
```
praveenrewar, line 93: Nice catch 🚀 
What do you think about making the error more descriptive, i.e telling exactly what is missing. So, if `len(pkgConfigs.Pkgs) != 1` then saying that "Reading package-resource.yml: expected exactly 1 Package, but got %v. (hint: delete the file and run `package init` again)"? (And same for PackageMetadata)

praveenrewar (commented):
Thank you so much for the fix @andrew-su 🚀 
Would you be open to adding a test for this? (A separate PR later on is fine as well)

cli/pkg/kctrl/cmd/package/release/release.go
andrew-su: That sounds like it could be work to update messaging for errors around the app. The error as is, does offer a suggestion on how it can be fixed. Perhaps an issue someone looking to start contributing can pick up as a good first issue?

As for the tests, I don't think I'll have time to get to them soon.

praveenrewar: Sure, no problem, we can make the change as part of a separate PR.

praveenrewar (approved):
LGTM!

PR #1317, opened 2023-09-04
Add defaultNamespace in pkgi/app crs

### LINKED ISSUE #1318: Ability to create app resources in a different namespace
I want to install an app in a particular namespace, for this I have to create my ServiceAccount, Role and RoleBinding in the same namespace. I don't my ServiceAccount to be present in this namespace to ensure that it cannot be used by other users.

I would like to have the ability to set a default namespace for the resources create by my App. This will enable me to create the App (along with the ServiceAccount) in a different namespace while setting a default namespace for the resources. Something like this:
```
--- my-namespace
   |--- Resource A (created by my App)
   |--- Resource B (created by my App)
   |--- Role
   |--- RoleBinding
--- app-namespace
   |--- App
   |--- ServiceAccount
   |--- Role
   |--- RoleBinding
```

We can achieve this by adding a field called `installationNamespace` in App spec which can be set as the preferred namespace in cluster access object.

---
Vote on this request

praveenrewar:
Add defaultNamespace field in App and PackageInstall spec. This can be used to set the default namespace where the resources will be created. Currently this is the same as the App namespace.

Fixes #1318 

```release-note
Introduce defaultNamespace in Apps and PackageInstalls. This can be used to set the default namespace where the resources will be installed.
```

##### Review Checklist:

  a link to that PR
  change

```docs

```


pkg/kubeconfig/kubeconfig.go
```diff
@@ -47,13 +47,13 @@ func NewKubeconfig(coreClient kubernetes.Interface, log logr.Logger) *Kubeconfig
...
 // if the saName is empty then you'll connect to a cluster via the clusterOpts inside the genericOpts, otherwise you'll use the specified SA.
-func (k Kubeconfig) ClusterAccess(saName string, clusterOpts *v1alpha1.AppCluster, accessLocation AccessLocation) (AccessInfo, error) {
+func (k Kubeconfig) ClusterAccess(saName string, clusterOpts *v1alpha1.AppCluster, accessLocation AccessLocation, preferredNamespace string) (AccessInfo, error) {
 	var err error
 	var clusterAccessInfo AccessInfo
 
 	switch {
 	case len(saName) > 0:
-		clusterAccessInfo, err = k.serviceAccounts.Find(accessLocation, saName)
+		clusterAccessInfo, err = k.serviceAccounts.Find(accessLocation, saName, preferredNamespace)
```
praveenrewar: I am thinking that it makes sense to add the preferredNamespace to clusterAccessInfo for both service accounts and clusterOpts.

pkg/kubeconfig/service_accounts.go
```diff
@@ -52,7 +52,7 @@ func (s *ServiceAccounts) Find(accessLocation AccessLocation, saName string) (Ac
 
 	pgoForSA := AccessInfo{
 		Name:       accessLocation.Name,
-		Namespace:  "", // Assume kubeconfig contains preferred namespace from SA
```
neil-hickey: 👋 It's been a while since I looked through this code. And at first glance I am wondering if accessLocation.Namespace is what we are wanting here? Or does that refer to a different namespace again?

praveenrewar: We have [this](https://github.com/carvel-dev/kapp-controller/blob/786f84b0cc75b135423cf7f7010fb7fbe681f628/pkg/deploy/kapp.go#L257-L259) check, so `-n installationNamespace` would be added if user provides spec.installationNamesapce. We could also add the namespace directly to the kubeconfig, but using AccessInfo seemed inline to the kapp behaviour, that by default ns from kubeconfig is used, and we can override that via `-n` flag.

config/config/crds.yml
```diff
@@ -1502,6 +1508,9 @@ spec:
                     description: Specifies namespace in destination cluster (optional)
                     type: string
                 type: object
+              installationNamespace:
+                description: Specifies the default namespace to install the Package resources, by default this is same as the PackageInstall namespace
```
neil-hickey:
```suggestion
                description: Specifies the default namespace to install the Package resources, by default this is same as the PackageInstall namespace (optional; v0.48.0+)
```

hack/dependencies.yml
```diff
@@ -22,18 +22,19 @@
...
-      amd64: b253ea9cf6add07f9497955147dc12e8612c24c36dc9929c9a4fecdc76752bd3
-      arm64: 25491298f6783a8b337d2ebdecf749f7750cf10260fe37086315a9c7da0b558f
+      amd64: f70119163d84b188b818ea7cdf6a88b4482e4e51248d30593a0b47399a1657e6
+      arm64: b1efb38cbc93630e377e74203e64f90b8cdffab2ff2e24c2365baec179a05e3b
   dev: true
   name: kapp
-  repo: vmware-tanzu/carvel-kapp
-  urlTemplate: https://github.com/vmware-tanzu/carvel-{{.Name}}/releases/download/{{.Version}}/{{.Name}}-{{.OS}}-{{.Arch}}
-  version: v0.58.0
+  repo: praveenrewar/kapp
```
neil-hickey: let's just not forget to change this back before we merge

pkg/app/app_deploy.go
```diff
@@ -146,7 +146,9 @@ func (a *App) trySaveMetadata(kapp *ctldep.Kapp) {
 func (a *App) newKapp(kapp v1alpha1.AppDeployKapp, cancelCh chan struct{}) (*ctldep.Kapp, error) {
 
 	return a.deployFactory.NewKapp(kapp, a.app.Spec.ServiceAccountName,
```
neil-hickey, line 148: this factory is starting to get pretty big for its constructor. I think its fine for this PR, but if we have to add much more - we might want to think about how to make this smaller / a structured input

pkg/packageinstall/packageinstall.go
```diff
@@ -374,6 +374,10 @@ func (pi *PackageInstallCR) reconcileDelete(modelStatus *reconciler.Status) (rec
 		existingApp.Spec.Canceled = pi.model.Spec.Canceled
 	}
 
+	if existingApp.Spec.InstallationNamespace != pi.model.Spec.InstallationNamespace {
```
neil-hickey: I'm not saying to change this here in this PR. Just observing and wondering why we even have these checks? Sort of feels redundant, they could just be assignments because if the value is different, we change it to `pi.model.Spec` and it's not then it must be equal to `pi.model.Spec`. So why even bother doing checks....

test/e2e/kappcontroller/installation_namespace_test.go
```diff
@@ -0,0 +1,288 @@
...
+          metadata:
+            name: %s
+          ---
+          apiVersion: v1
+          kind: Namespace
+          metadata:
+            name: %s
+`, installationNamespace, pkgiNamespace)
+
+	rbacYAML := fmt.Sprintf(`
```
neil-hickey: Any change we could move this to `service_accounts.go` into a new function? I feel like if we ever have to change this SA setup for the tests that this test is likely to get forgotten about.

test/e2e/kappcontroller/installation_namespace_test.go
```diff
@@ -0,0 +1,288 @@
...
+	sas := e2e.ServiceAccounts{env.Namespace}
+
+	// deploy app with workspace and pkgi
+	_, err := kapp.RunWithOpts([]string{"deploy", "-a", name, "-f", "-"}, e2e.RunOpts{StdinReader: strings.NewReader(appWithInlineResources(appName, namespacesYAML+rbacYAML+installPkgYAML) + sas.ForClusterYAML())})
+	if err != nil {
+		t.Fatalf("Expected app deploy to succeed, it did not: %v", err)
+	}
+
+	// assert that pkgi resources are present in installation namespace
+	_, err = kubectl.RunWithOpts([]string{"get", "configmap", "my-cm", "-n", installationNamespace}, e2e.RunOpts{AllowError: true, NoNamespace: true})
```
neil-hickey:
```suggestion
	_, err = kubectl.RunWithOpts([]string{"get", "configmap", "my-cm", "-n", installationNamespace}, e2e.RunOpts{AllowError: false, NoNamespace: true})
```

Should this be false on the error allowed? Same as one below

test/e2e/kappcontroller/installation_namespace_test.go
```diff
@@ -0,0 +1,288 @@
...
+          data:
+            key: value
+  template:
+  - ytt: {}
+  deploy:
+  - kapp: {}`, name, env.Namespace, installationNamespace) + sas.ForClusterYAML()
+
+	kapp.RunWithOpts([]string{"deploy", "-a", installationNamespaceApp, "-f", "-"}, e2e.RunOpts{StdinReader: strings.NewReader(namespaceYAML)})
+
+	_, err := kapp.RunWithOpts([]string{"deploy", "-a", name, "-f", "-"}, e2e.RunOpts{StdinReader: strings.NewReader(appYAML)})
```
neil-hickey: For new tests we have been trying to use `"github.com/stretchr/testify/assert"`

So this would be `assert.NoError(t, err)`

neil-hickey (commented):
Looks good, just a couple of smaller things about formatting / libs used. The overall code and whatnot looks good

test/e2e/kappcontroller/installation_namespace_test.go
praveenrewar: I typically prefer not using AllowError in these scenarios so that the kubectl command fails, but I copied this style from other tests where we allow the error to occur and then wrap the error in a message.
So, I am totally fine with setting allow error to false.

cppforlife:
not sure if yall already settled on the name, if so ignore, but if not, id like to throw in `defaultNamespace` as an alternative name. _imho_ reasons why it's better:
- installation feels much closer to PackageInstall but this field is also available on App
- this namespace is used as a _default_ namespace for resources that do not specify a namespace, but doesnt actually override it afaik

pkg/app/app_deploy.go
praveenrewar: Yep, makes sense.

test/e2e/kappcontroller/installation_namespace_test.go
praveenrewar: Done.

test/e2e/kappcontroller/installation_namespace_test.go
praveenrewar: Updated.

pkg/packageinstall/packageinstall.go
praveenrewar: Yeah, the check seems weird. For now, I have removed the check for the defaultNamespace. If we want we can update the others in a separate PR.

hack/dependencies.yml
praveenrewar: Done

neil-hickey (approved):
LGTM

PR #1322, opened 2023-09-13
Return user friendly error when package doesn't exist

### LINKED ISSUE #594: `kubectl get package` or `kctrl package available get` results in a non user friendly error when the package is not available
**What steps did you take:**
Run
```
kubectl get package foo.1.0.0
```
OR
```
kctrl package available get -p foo/1.0.0
```
where the package foo doesn't exist

**What happened:**
Errored out with the message in both the scenarios:
```
internalpackages.internal.packaging.carvel.dev "cpnmubhh5oo2sc0" not found
```

**What did you expect:**
The error message should have been something like 
```
internalpackages.internal.packaging.carvel.dev "foo.1.0.0" not found
```
to make it more user friendly.

**Environment:**

- kapp Controller version - v0.30.2
- Kubernetes version - v1.22.3

---
Vote on this request

rcmadhankumar:

Return user friendly error when package doesn't exist

```
$ kubectl get package foo.1.0.0
Error from server (NotFound): package.data.packaging.carvel.dev "foo.1.0.0" not found

$kctrl package available get -p foo/1.0.0
kctrl: Error: package.data.packaging.carvel.dev "foo.1.0.0" not found
```

Fixes #594 

```release-note
NONE
```

##### Review Checklist:

  a link to that PR
  change

```docs
NONE
```

### DIFF
--- a/pkg/apiserver/registry/datapackaging/package_crd_rest_test.go
+++ b/pkg/apiserver/registry/datapackaging/package_crd_rest_test.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/require"
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/internalpackaging/v1alpha1"
 	"github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/apis/datapackaging"
 	datapkgreg "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/registry/datapackaging"
@@ -20,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	k8sfake "k8s.io/client-go/kubernetes/fake"
 	cgtesting "k8s.io/client-go/testing"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 const globalNamespace = "global.packaging.kapp-controller.carvel.dev"
@@ -31,7 +33,7 @@ func TestPackageVersionListIncludesGlobalAndNamespaced(t *testing.T) {
 	internalClient := fake.NewSimpleClientset(globalIntPackageVersion(), namespacedIntPackageVersion(), excludedNonGlobalIntPackageVersion())
 	fakeCoreClient := k8sfake.NewSimpleClientset(namespace())
 
-	pkgvCRDREST := datapkgreg.NewPackageCRDREST(internalClient, fakeCoreClient, globalNamespace)
+	pkgvCRDREST := datapkgreg.NewPackageCRDREST(internalClient, fakeCoreClient, globalNamespace, logf.Log)
 
 	pkgvList, err := pkgvCRDREST.List(namespacedCtx(nonGlobalNamespace), &internalversion.ListOptions{})
 	if err != nil {
@@ -66,7 +68,7 @@ func TestPackageVersionListPrefersNamespacedOverGlobal(t *testing.T) {
 	internalClient := fake.NewSimpleClientset(globalIntPackageVersion(), overrideIntPackageVersion())
 	fakeCoreClient := k8sfake.NewSimpleClientset(namespace())
 
-	pkgvCRDREST := datapkgreg.NewPackageCRDREST(internalClient, fakeCoreClient, globalNamespace)
+	pkgvCRDREST := datapkgreg.NewPackageCRDREST(internalClient, fakeCoreClient, globalNamespace, logf.Log)
 
 	// list package versions and verify all of them are there
 	pkgvList, err := pkgvCRDREST.List(namespacedCtx(nonGlobalNamespace), &internalversion.ListOptions{})
@@ -101,7 +103,7 @@ func TestPackageVersionGetNotPresentInNS(t *testing.T) {
 	internalClient := fake.NewSimpleClientset(globalPackageVersion)
 	fakeCoreClient := k8sfake.NewSimpleClientset(namespace())
 
-	pkgvCRDREST := datapkgreg.NewPackageCRDREST(internalClient, fakeCoreClient, globalNamespace)
+	pkgvCRDREST := datapkgreg.NewPackageCRDREST(internalClient, fakeCoreClient, globalNamespace, logf.Log)
 
 	obj, err := pkgvCRDREST.Get(namespacedCtx(nonGlobalNamespace), name, &metav1.GetOptions{})
 	if err != nil {
@@ -126,7 +128,7 @@ func TestPackageVersionGetPresentInOnlyNS(t *testing.T) {
 	internalClient := fake.NewSimpleClientset(namespacedPackageVersion)
 	fakeCoreClient := k8sfake.NewSimpleClientset(namespace())
 
-	pkgvCRDREST := datapkgreg.NewPackageCRDREST(internalClient, fakeCoreClient, globalNamespace)
+	pkgvCRDREST := datapkgreg.NewPackageCRDREST(internalClient, fakeCoreClient, globalNamespace, logf.Log)
 
 	obj, err := pkgvCRDREST.Get(namespacedCtx(nonGlobalNamespace), name, &metav1.GetOptions{})
 	if err != nil {
@@ -146,21 +148,20 @@ func TestPackageVersionGetPresentInOnlyNS(t *testing.T) {
 func TestPackageVersionGetNotFound(t *testing.T) {
 	namespacedPackageVersion := excludedNonGlobalIntPackageVersion()
 	name := namespacedPackageName
+	expectedError := "package.data.packaging.carvel.dev \"" + name + "\" not found"
 
 	internalClient := fake.NewSimpleClientset(namespacedPackageVersion)
 	fakeCoreClient := k8sfake.NewSimpleClientset(namespace())
 
-	pkgvCRDREST := datapkgreg.NewPackageCRDREST(internalClient, fakeCoreClient, globalNamespace)
+	pkgvCRDREST := datapkgreg.NewPackageCRDREST(internalClient, fakeCoreClient, globalNamespace, logf.Log)
 
 	_, err := pkgvCRDREST.Get(namespacedCtx(nonGlobalNamespace), name, &metav1.GetOptions{})
 	if err == nil {
 		t.Fatalf("Expected get operation to fail, but it didn't")
 	}
 
-	if !errors.IsNotFound(err) {
-		t.Fatalf("Expected a not found error, got: %v", err)
-	}
-
+	require.True(t, errors.IsNotFound(err))
+	require.ErrorContains(t, err, expectedError)
 }
 
 func TestPackageVersionGetPreferNS(t *testing.T) {
@@ -171,7 +172,7 @@ func TestPackageVersionGetPreferNS(t *testing.T) {
 	internalClient := fake.NewSimpleClientset(overridePackageVersion, globalIntPackageVersion())
 	fakeCoreClient := k8sfake.NewSimpleClientset(namespace())
 
-	pkgvCRDREST := datapkgreg.NewPackageCRDREST(internalClient, fakeCoreClient, globalNamespace)
+	pkgvCRDREST := datapkgreg.NewPackageCRDREST(internalClient, fakeCoreClient, globalNamespace, logf.Log)
 
 	obj, err := pkgvCRDREST.Get(namespacedCtx(nonGlobalNamespace), name, &metav1.GetOptions{})
 	if err != nil {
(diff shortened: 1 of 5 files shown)


rcmadhankumar:
@100mik @praveenrewar can you review this?

pkg/apiserver/registry/datapackaging/package_storage_client.go
```diff
@@ -152,7 +152,28 @@ func (t PackageTranslator) ToExternalWatcher(intObjWatcher watch.Interface, fiel
 }
 
 func (t PackageTranslator) ToExternalError(err error) error {
-	// TODO: implement
+	statusErr, isStatusError := err.(*errors.StatusError)
```
praveenrewar: I think we can use the APIStatus interface, like [this].(https://github.com/kubernetes/apimachinery/blob/12dc3f82eb47fc972b5be0b222f4422cb1b51612/pkg/api/errors/errors.go#L99)
```
status, ok := err.(APIStatus)
if !ok {
  return err
}
```

praveenrewar: nit: Typically, we just use an `ok` (instead of `isStatusError`) as it's a very short lived variable.

pkg/apiserver/registry/datapackaging/package_storage_client.go
```diff
@@ -152,7 +152,28 @@ func (t PackageTranslator) ToExternalWatcher(intObjWatcher watch.Interface, fiel
...
+	if !isStatusError {
+		return err
+	}
+
+	details := statusErr.ErrStatus.Details
+	if details.Kind == "internalpackages" && details.Group == internalpkgingv1alpha1.SchemeGroupVersion.Group {
```
praveenrewar: Let's also do a nil check for details. Also, we can just return if it one of these conditions is not true.

pkg/apiserver/registry/datapackaging/package_storage_client.go
```diff
@@ -152,7 +152,28 @@ func (t PackageTranslator) ToExternalWatcher(intObjWatcher watch.Interface, fiel
...
+		packageName, err := t.ToExternalName(details.Name)
+		if err != nil {
+			return err
```
praveenrewar, line 169: Let's return the original error if there is any error while translating to external error.

pkg/apiserver/registry/datapackaging/package_storage_client.go
```diff
@@ -152,7 +152,28 @@ func (t PackageTranslator) ToExternalWatcher(intObjWatcher watch.Interface, fiel
...
+		}
+
+		switch statusErr.ErrStatus.Reason {
```
praveenrewar: Let's have a nil check for reason as well.

praveenrewar: Maybe add a TODO to implement other type of errors?

test/e2e/kappcontroller/package_test.go
```diff
@@ -409,3 +409,27 @@ spec:
 		}
 	})
 }
+
+func TestPackageNotExists(t *testing.T) {
```
praveenrewar: Maybe `Test_PackageNotFound`? I am really bad at this so not sure.

test/e2e/kappcontroller/package_test.go
```diff
@@ -409,3 +409,27 @@ spec:
...
+	env := e2e.BuildEnv(t)
+	logger := e2e.Logger{}
+	k := e2e.Kubectl{t, env.Namespace, logger}
+	packageName := "foo.1.0.0"
+	expectedError := "stderr: 'Error from server (NotFound): package.data.packaging.carvel.dev \"foo.1.0.0\" not found"
+
+	logger.Section("Get Package", func() {
+		_, err := k.RunWithOpts([]string{"get", "package", packageName}, e2e.RunOpts{AllowError: true})
+		if !strings.Contains(err.Error(), expectedError) {
```
praveenrewar: Let's check if err is non nil first.
We are trying to switch to `assert`, so let's use that.

praveenrewar (approved):
LGTM!

pkg/apiserver/registry/datapackaging/package_crd_rest_test.go
```diff
@@ -157,10 +159,8 @@ func TestPackageVersionGetNotFound(t *testing.T) {
 		t.Fatalf("Expected get operation to fail, but it didn't")
 	}
 
-	if !errors.IsNotFound(err) {
-		t.Fatalf("Expected a not found error, got: %v", err)
-	}
-
+	assert.True(t, errors.IsNotFound(err))
```
joaopapereira: I think we should use require in this case. The main difference is that assert does not stop the execution if the assertion is true, while the require does that.

pkg/apiserver/registry/datapackaging/package_storage_client.go
```diff
@@ -152,7 +153,30 @@ func (t PackageTranslator) ToExternalWatcher(intObjWatcher watch.Interface, fiel
 }
 
 func (t PackageTranslator) ToExternalError(err error) error {
-	// TODO: implement
+
```
joaopapereira:
```suggestion
```

pkg/apiserver/registry/datapackaging/package_storage_client.go
```diff
@@ -152,7 +153,30 @@ func (t PackageTranslator) ToExternalWatcher(intObjWatcher watch.Interface, fiel
...
+	status, ok := err.(apierrors.APIStatus)
+	if !(ok || errors.As(err, &status)) {
+		return err
+	}
```
joaopapereira: Should this code be something like this?
```suggestion
	var status *apierrors.APIStatus
	if !(errors.As(err, &status)) {
		return err
	}
```

pkg/apiserver/registry/datapackaging/package_storage_client.go
```diff
@@ -152,7 +153,30 @@ func (t PackageTranslator) ToExternalWatcher(intObjWatcher watch.Interface, fiel
...
+
+	details := status.Status().Details
+	if details != nil && details.Kind == "internalpackages" && details.Group == internalpkgingv1alpha1.SchemeGroupVersion.Group {
+		packageName, translateErr := t.ToExternalName(details.Name)
+		if translateErr != nil {
+			return err
```
joaopapereira, line 169: We are hiding here a possible translation error. In what situation does this error happen?
Maybe it would make sense to save it somehow, or by wrapping the current error and extending with more information or to log this error out somewhere.

pkg/apiserver/registry/datapackaging/package_storage_client.go
```diff
@@ -152,7 +153,30 @@ func (t PackageTranslator) ToExternalWatcher(intObjWatcher watch.Interface, fiel
...
+		}
+
+		switch status.Status().Reason {
+		case metav1.StatusReasonNotFound:
+			return apierrors.NewNotFound(datapkgingv1alpha1.Resource("package"), packageName)
+		case metav1.StatusReasonAlreadyExists:
+			return apierrors.NewAlreadyExists(datapkgingv1alpha1.Resource("package"), packageName)
+		// TODO Handle other types of errors
```
joaopapereira: Do we plan on doing this for other errors? if not we can remove the TODO

test/e2e/kappcontroller/package_test.go
```diff
@@ -409,3 +410,23 @@ spec:
...
+	env := e2e.BuildEnv(t)
+	logger := e2e.Logger{}
+	k := e2e.Kubectl{t, env.Namespace, logger}
+	packageName := "foo.1.0.0"
+	expectedError := "stderr: 'Error from server (NotFound): package.data.packaging.carvel.dev \"foo.1.0.0\" not found"
+
+	logger.Section("Get Package", func() {
+		_, err := k.RunWithOpts([]string{"get", "package", packageName}, e2e.RunOpts{AllowError: true})
+		assert.NotNil(t, err)
+		assert.ErrorContains(t, err, expectedError)
```
joaopapereira:
```suggestion
		require.NotNil(t, err)
		require.ErrorContains(t, err, expectedError)
```
Also ErrorContains should already check if the error is not or not.

test/e2e/kappcontroller/package_test.go
```diff
@@ -409,3 +410,23 @@ spec:
...
+	})
+
+	logger.Section("delete Package", func() {
+		_, err := k.RunWithOpts([]string{"delete", "package", packageName}, e2e.RunOpts{AllowError: true})
+		assert.NotNil(t, err)
+		assert.ErrorContains(t, err, expectedError)
```
joaopapereira:
```suggestion
		require.NotNil(t, err)
		require.ErrorContains(t, err, expectedError)
```
Also ErrorContains should already check if the error is not or not.

joaopapereira (changes requested):
Made some comments with some suggested changes for the PR. In general looks good.
Since we are starting to use testify in the tests, it would be good if we did update, at least the test file to use `require` instead of the `t.Fatalf`.

pkg/apiserver/registry/datapackaging/package_storage_client.go
rcmadhankumar: We need to plan on reproducing errors with these[ status reasons](https://github.com/kubernetes/apimachinery/blob/60eaa653342bd0cc48626fa2604bb6d3e7171706/pkg/api/errors/errors.go#L47) and handle the possible errors.

pkg/apiserver/registry/datapackaging/package_storage_client.go
rcmadhankumar: ToExternalName decodes the base 64 encoded package name.
```
func (t PackageTranslator) ToExternalName(name string) (string, error) {
	decodedBytes, err := base32.HexEncoding.WithPadding(base32.NoPadding).DecodeString(strings.ToUpper(name))
	return string(decodedBytes), err
}
```

joaopapereira: Ok, but we are hiding an error, is that ok?

pkg/apiserver/registry/datapackaging/package_storage_client.go
joaopapereira: If we are not planning to surface them separately in kapp-controller we should just remove the TODO since we are not planning on doing it

pkg/apiserver/registry/datapackaging/package_storage_client.go
praveenrewar: Good catch, I somehow overlooked this after my last comment on returning the original error.
I think we should log the error, wrapping the existing error might confuse the users (as the original itself is confusing because of the hashed name)

pkg/apiserver/registry/datapackaging/package_storage_client.go
praveenrewar: Makes sense, I think we have covered the most common errors, we can think of handling other errors if there is a demand for it.

pkg/apiserver/registry/datapackaging/package_storage_client.go
praveenrewar: This was a copy from https://github.com/kubernetes/apimachinery/blob/12dc3f82eb47fc972b5be0b222f4422cb1b51612/pkg/api/errors/errors.go#L99. I am okay with both the approaches.

pkg/apiserver/registry/datapackaging/package_storage_client.go
rcmadhankumar: which means we should return the translate error?

praveenrewar: We should log the translation error.

pkg/apiserver/registry/datapackaging/package_crd_rest_test.go
rcmadhankumar: require is used.

pkg/apiserver/registry/datapackaging/package_storage_client.go
rcmadhankumar: Logger is added to print the translate error.

pkg/apiserver/registry/datapackaging/package_crd_rest_test.go
```diff
@@ -31,7 +35,10 @@ func TestPackageVersionListIncludesGlobalAndNamespaced(t *testing.T) {
 	internalClient := fake.NewSimpleClientset(globalIntPackageVersion(), namespacedIntPackageVersion(), excludedNonGlobalIntPackageVersion())
 	fakeCoreClient := k8sfake.NewSimpleClientset(namespace())
 
-	pkgvCRDREST := datapkgreg.NewPackageCRDREST(internalClient, fakeCoreClient, globalNamespace)
+	log := zap.New(zap.UseDevMode(false)).WithName("package_crd_rest_test")
+	logf.SetLogger(log)
+	klog.SetLogger(log)
+	pkgvCRDREST := datapkgreg.NewPackageCRDREST(internalClient, fakeCoreClient, globalNamespace, log)
```
praveenrewar: I guess we can directly use logf.Log here as we are not testing the logging part.
Same for other places.

rcmadhankumar: logger changed for the tests.

joaopapereira (approved):
LGTM

praveenrewar (approved):
LGTM!

PR #1382, opened 2023-10-27
[Add] CEL rule for validating App and PackageInstall Spec

### LINKED ISSUE #598: App CR should require one of serviceAccountName or cluster.

Both `serviceAccountName` and `cluster` properties of the `App` CR are marked as optional. Anyone not understanding that one or the other is required may not even add either. When neither is added, then the `App` CR is still accepted and you end up with an error which you have to look in the `App` CR status of:

```
  status:
    conditions:
    - message: 'Preparing kapp: Expected service account or cluster specified'
      status: "True"
      type: ReconcileFailed
```

Further, when trying to delete the `App` in this situation it fails with it eventually exiting with the error:

```
kapp: Error: waiting on reconcile app/educates-training-platform (kappctrl.k14s.io/v1alpha1) namespace: default:
  Finished unsuccessfully (Reconcile failed:  (message: Preparing kapp: Expected service account or cluster specified))
```

resulting in you needing to manually delete the finalizer from the `App` resource.

The latter problem relates to issue https://github.com/vmware-tanzu/carvel-kapp-controller/issues/416.

The CR definition should enforce that one or the other of `serviceAccountName` and `cluster` is required so that attempting to create an `App` CR with neither is rejected immediately.

My understanding is that this can be achieved in the CR schema by using:

```
anyOf:
- required:
  - serviceAccountName
- required:
  - cluster
```

Having to look into the status of the `App` CR is non obvious so anything that avoids getting into that situation in the first place is a bonus. So failing early at the point of creating the `App` is better.

---
Vote on this request

### LINKED ISSUE #1380: Introduce CEL based validation in App spec for SA and Cluster

Tried creating an App CR with no service account or cluster specified. The resource got created, but reconcile errored stating:
`Reconcile failed: Preparing kapp: Expected service account or cluster specified`. The [documentation](https://carvel.dev/kapp-controller/docs/v0.48.x/app-spec/) states that both of them are optional fields, hence as a user I'm unaware that one of them needs to be specified. 

Based on the conversation [here](https://kubernetes.slack.com/archives/CH8KCCKA5/p1698351828408469?thread_ts=1698350082.438849&cid=CH8KCCKA5), it does make sense to either have the user specify the target cluster where App CR needs to be installed, or the sa which needs to be used to install it in the same cluster. However, running this validation after the resource is created is not necessary (refer: https://github.com/carvel-dev/kapp-controller/blob/5b1294bdf60107c9ff20b79e519abaefd71c4682/pkg/kubeconfig/kubeconfig.go#L75). This can also be done on the admission control level before an App CR is created. 

Solution: Add CEL based validation to ensure that one of them are specified. 

None 

---
Vote on this request

varshaprasad96:
This PR:
- Adds KB marker to ensure that either spec.ServiceAccount or spec.Cluster is present in App CR.
- Bumps controller-tools to 0.10.0 to support CEL based validation marker. That is the latest version compatible with the k8s release the project is currently at.

Fixes #1380 

No. The validation already exists, this change performs the same action before the resource is created. 

##### Review Checklist:

  a link to that PR
  change

```docs

```


prembhaskal:
one comment based on my first glance.
currently if both serviceAccount and cluster are specified in a PackageInstall CRD, the serviceAccount is read and used.
if none are specified, we get an error.

so validation in PR  might break some existing configuration.
@praveenrewar  @100mik Is current behaviour to choose serviceAccount over cluster config by design?

config/config/crds.yml
```diff
@@ -1306,6 +1309,9 @@ spec:
                   type: object
                 type: array
             type: object
+            x-kubernetes-validations:
+            - message: Expected service account or cluster, both cannot be specified
+              rule: 'has(self.serviceAccountName) ? !has(self.cluster) : has(self.cluster)'
```
prembhaskal: validation will be needed in "kind: PackageInstall" too, the next crd in this file.

praveenrewar: +1

praveenrewar (commented):
@varshaprasad96 Could you run `'./hack/build.sh', './hack/gen.sh', and './hack/gen-apiserver.sh'` and add the changes to the generated files as well.
With these changes, 
- we will error out when serviceAccount and cluster both are not provided (✅ )
- we will error out when both serviceAccount and cluster are provided (Even though this is how it should be, like @prembhaskal mentioned this might be a breaking change for some folks, for example if they have an App CR where they have mentioned both serviceAccount and cluster, then they will start seeing this error when they upgrade to the newer version of kc.)

> Is current behaviour to choose serviceAccount over cluster config by design?

I am not sure if this was intended for long term, but yes, this is the current expected behaviour (I think it's mentioned somewhere in the docs as well) 

cc @neil-hickey @joaopapereira

varshaprasad96:
@prembhaskal @praveenrewar I've modified the validation to just check the existence of atleast one of it. This should retain the current behaviour and fix the failing tests.

praveenrewar (commented):
Thank you so much for the PR @varshaprasad96. The changes look good to me, could you also add a quick e2e test [here](https://github.com/carvel-dev/kapp-controller/tree/develop/test/e2e/kappcontroller).

prembhaskal (commented):
Changes look good.

prembhaskal (dismissed):

prembhaskal:
@praveenrewar  @100mik please review.

test/e2e/kappcontroller/packageinstall_test.go
```diff
@@ -950,3 +950,64 @@ spec:
 		assert.Equal(t, actualSyncPeriod, expectedSyncPeriod)
 	})
 }
+
+
+func TestCELValidation(t *testing.T) {
```
praveenrewar: Should we call it `TestPackageInstallWithoutSAOrCluster`?

test/e2e/kappcontroller/packageinstall_test.go
```diff
@@ -950,3 +950,64 @@ spec:
...
+	logger := e2e.Logger{}
+	kapp := e2e.Kapp{t, env.Namespace, logger}
+	kubectl := e2e.Kubectl{t, env.Namespace, logger}
+
+	name := "incorrect-spec-without-sa-cluster"
+
+	appYAML := fmt.Sprintf(`
+---
+apiVersion: kappctrl.k14s.io/v1alpha1
+kind: App
```
praveenrewar: The name of the test file indicates that it's for package installs, should we add a new file for cel validations?

test/e2e/kappcontroller/packageinstall_test.go
```diff
@@ -950,3 +950,64 @@ spec:
...
+		require.Error(t, err)
+		require.ErrorContains(t, err, "Expected service account or cluster.")
+	})
+
+	logger.Section("Create PackageInstall with kapp", func() {
+		_, err := kapp.RunWithOpts([]string{"deploy", "-a", name, "-f", "-"}, e2e.RunOpts{StdinReader: strings.NewReader(pkginstallYAML), AllowError: true})
+		require.Error(t, err)
+		require.ErrorContains(t, err, "Expected service account or cluster.")
+	})
+}
```
praveenrewar: nit: Add a new line at the end of the file.

praveenrewar (commented):
(A few minor changes)

test/e2e/kappcontroller/packageinstall_test.go
varshaprasad96: I've added this test to a new file, so that in future in any CEL related validation tests to any of the APIs can go in there.

test/e2e/kappcontroller/cel_validation.go
prembhaskal, line 1: file name missing "_test",  so test is still not running in e2e .

prembhaskal (changes requested):

prembhaskal:
~~all seems fine.~~
Well test is failing actually.

prembhaskal:
I think test is failing because k8s version of minikube is set to 1.20 which does not support CEL validation.
https://github.com/carvel-dev/kapp-controller/blob/b4886dfaff8050fc7221e1d7bce9e69e2a4264e3/.github/workflows/test-gh.yml#L49

praveenrewar:
We could bump k8s version in the tests, but I don't want us to indicate that we don't support any version before 1.26. Since the kind tests with the latest version are passing, I think we can add a check in this test to skip if k8s version is <1.26.x (as the validation won't work on older versions but there shouldn't be any other issue because of it).

varshaprasad96:
Sorry for not being able to follow up due to other commitments. The recent changes should skip the test if the server version is less than 1.26.

praveenrewar:
@varshaprasad96 The test seems to be failing while getting the server version (I haven't taken a closer look so not sure exactly which part).

prembhaskal:
> cel_validation_test.go:26: Error getting Kubernetes version: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined

failing because the test code is not running on the kind cluster.

PR #1394, opened 2023-11-08
Enhance fallback to noop logic for apps to account for multiple namespaces

### LINKED ISSUE #1392: Unblock app deletion by falling back to noop when safe
#1208 and #1220 introduced a fallback scenario, where Apps perform a loop delete to unblock themselves when they can do so safely. This however does not take into account multiple namespaces that might be terminating without owning any cluster resources.

The same fallback mechanism should be used while dealing with Apps affecting multiple namespaces and not managing cluster resources.

---
Vote on this request

100mik:
It ensures that Apps and PackageInstalls do not block namespace deletion when the app can safely fallback to `noopDelete`

Fixes #1392 

##### Review Checklist:

  a link to that PR
  change

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -276,6 +276,35 @@ func (a *App) ConfigMapRefs() map[reftracker.RefKey]struct{} {
 	return configMaps
 }
 
+func (a *App) noopDeleteDueToTerminatingNamespaces() bool {
+	if a.app.Status.Deploy == nil || a.app.Status.Deploy.KappDeployStatus == nil || a.app.Spec.ServiceAccountName == "" {
+		return false
+	}
+	if !a.isNamespaceTerminating(a.app.Namespace) {
+		return false
+	}
+	// Ensure that no cluster scoped resources are created by the app
+	// and all affected namespaces are terminating
+	for _, ns := range a.app.Status.Deploy.KappDeployStatus.AssociatedResources.Namespaces {
+		if ns == "(cluster)" {
+			return false
+		}
+		if !a.isNamespaceTerminating(ns) {
+			return false
+		}
+	}
+	a.log.Info("Safely performing noop delete to avoid blocking namespace deletion")
+	return true
+}
+
+func (a *App) isNamespaceTerminating(namespace string) bool {
+	status, err := a.compInfo.NamespaceStatus(namespace)
+	if err != nil {
+		a.log.Error(err, "Error getting app namespace status", "app", a.app.Name, "namespace", a.app.Namespace)
+	}
+	return status.Phase == v1.NamespaceTerminating
+}
+
 // HasImageOrImgpkgBundle is used to determine if the
 // App's spec contains a fetch stage for an image or
 // imgpkgbundle. It is mainly used to determine whether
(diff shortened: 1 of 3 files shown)


pkg/app/app.go
```diff
@@ -276,6 +276,38 @@ func (a *App) ConfigMapRefs() map[reftracker.RefKey]struct{} {
 	return configMaps
 }
 
+func (a *App) isSafeToFallbackToNoopDelete() bool {
```
praveenrewar: `isSafeToFallbackToNoopDelete` the name indicates that we are first trying to delete and then falling back to noopDelete in case of an error. Maybe we can have something like `noopDeleteDueToTerminatingNamespaces`?

pkg/app/app.go
```diff
@@ -276,6 +276,38 @@ func (a *App) ConfigMapRefs() map[reftracker.RefKey]struct{} {
...
+	if a.app.Status.Deploy == nil || a.app.Status.Deploy.KappDeployStatus == nil || a.app.Spec.Cluster != nil {
+		return false
+	}
+
+	// Ensure that app config map will be garbage collected due to namespace termination
+	status, err := a.compInfo.NamespaceStatus(a.app.Namespace)
```
praveenrewar: I think this namespace will be part of the list in the app status as kapp keeps the namespace where the cm is created as well. If we want to keep this to be on the safer side, I guess we can keep the `isNamespaceTerminating` method and pass the namespace as a parameter to avoid some duplicate code.

pkg/app/app.go
```diff
@@ -276,6 +276,38 @@ func (a *App) ConfigMapRefs() map[reftracker.RefKey]struct{} {
 	return configMaps
 }
 
+func (a *App) isSafeToFallbackToNoopDelete() bool {
+	if a.app.Status.Deploy == nil || a.app.Status.Deploy.KappDeployStatus == nil || a.app.Spec.Cluster != nil {
```
praveenrewar: One thing that I forgot last time was that the cluster can be non nil and we still might be relying on the SA if it is non nil as well, so we should be checking spec.SA instead.

praveenrewar (commented):
🚀

pkg/app/app.go
100mik: > I think this namespace will be part of the list in the app status as kapp keeps the namespace

That does not seem to be the case, I just verified with a quick install. I think kapp only stores resources where it has to do list calls while grabbing resources. 

>  I guess we can keep the isNamespaceTerminating method and pass the namespace as a parameter to avoid some duplicate code.

Yep that sounds good!

pkg/app/app.go
100mik: That makes sense to me.

pkg/app/app.go
100mik: Ah I see, you mean we give preference to SA? So checking `a.app.ServiceAccountName == nil` is safer here?

100mik: Made this change

praveenrewar (approved):
LGTM!

joaopapereira (approved):
LGTM

Issue #1286, opened 2023-08-08
Feature request: add option for skipping SSL verification when using Git


joaopapereira:
**Describe the problem/challenge you have**
Similar to carvel-dev/vendir#260, we need to allow the users of kapp-controller to use this new behavior in vendir.

**Describe the solution you'd like**
kapp-controller today as the flag `dangerousSkipTLSVerify`, but currently, we are not providing this flag to vendir when fetching from a git repository. We should enable this.

**Anything else you would like to add:**
[Additional information that will assist in solving the issue.]

---
Vote on this request.

This invites the community to vote on issues to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible."
👎 "There are other more important things to focus on right now."

We are also happy to receive and review Pull Requests if you want to help work on this issue.

joaopapereira:
Closing this issues since PR #1419 was merged in December

PR #1419, opened 2023-12-13
adding option for skipping SSL verification when using Git 

sethiyash:
#1286 

#1286 

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```


praveenrewar (commented):
The changes in the PR looks good to me. However, I think the idea was to use the existing `dangerousSkipTLSVerify` flag available in [kapp config](https://carvel.dev/kapp-controller/docs/latest/controller-config/#controller-configuration-spec), so we need to extend the functionality [here](https://github.com/carvel-dev/kapp-controller/blob/0594982604cb1a553bb12545cfa0d5a285617e5a/pkg/fetch/vendir.go#L382)

pkg/fetch/vendir.go
```diff
@@ -421,3 +421,20 @@ func ExtractImageRegistry(name string) string {
...
+func extractGitHostname(input string) string {
+	u, err := url.Parse(input)
+	if err != nil {
+		return "github.com"
+	}
+	return u.Hostname()
+}
+
+// ExtractHost return registry for Docker Image and Host for git url
+func ExtractHost(input string, isGitURL bool) string {
```
praveenrewar: `isGirURL` doesn't sound very extensible, let's have a _type_ defined for the source type (like imgpkg or git) which we can pass down, and have switch case to extract the host, this way we can extend the functionality for other sources in future if needed.

praveenrewar (commented):
Is it possible to add a test for this change?

pkg/fetch/vendir.go
```diff
@@ -376,11 +386,9 @@ func (v *Vendir) configMapBytes(configMapRef vendirconf.DirectoryContentsLocalRe
 	return kyaml.Marshal(configMap)
 }
 
-// This function only works on image refs. If in the future we decide to
-// expand this option to other fetch options, we will need to add hostname
-// extraction for those
-func (v *Vendir) shouldSkipTLSVerify(url string) bool {
-	return v.opts.SkipTLSConfig.ShouldSkipTLSForAuthority(ExtractImageRegistry(url))
+// This function works on image refs and hostname extraction using isGitURL flag
```
praveenrewar: nit: Update comment

pkg/fetch/vendir_test.go
```diff
@@ -58,32 +58,52 @@ func Test_AddDir_skipsTLS(t *testing.T) {
...
+		},
+		{
+			name:       "https://github.com/bitnami/charts/",
+			sourceType: fetch.GitURL,
+			want:       "github.com",
+		},
+		{
+			name:       "http://github.com/bitnami/charts/",
+			sourceType: fetch.GitURL,
+			want:       "github.com",
```
praveenrewar: Both test cases seem to be the same. Can we add a gitlab example as well?

pkg/fetch/vendir_test.go
```diff
@@ -58,32 +58,52 @@ func Test_AddDir_skipsTLS(t *testing.T) {
 	}
 }
 
-func TestExtractImageRegistry(t *testing.T) {
```
praveenrewar, line 61: Let's also add a test similar to the one above (Test_AddDir_skipTLS) for git urls.

pkg/fetch/vendir.go
```diff
@@ -21,7 +22,15 @@ import (
 	kyaml "sigs.k8s.io/yaml"
 )
 
+// SourceType to extract host
+type SourceType string
+
 const (
+	// GitURL source type to extract host
+	GitURL SourceType = "gitURL"
```
praveenrewar: I don't think we are going to use these string values, so we could probably just switch to int (Iota).

praveenrewar (commented):
Adding an e2e test with a local gitlab server might be a bit much, but it would be great if we can specify what sort of manual testing has been done. (We follow [these](https://github.com/carvel-dev/vendir/pull/265#issuecomment-1746041098) steps and create an App which uses local gitlab server)

pkg/fetch/vendir.go
```diff
@@ -22,6 +23,11 @@ import (
 )
 
 const (
+	// GitURL source type to extract host
+	GitURL = iota
+	// ImageRegistry source type to extract host
+	ImageRegistry
+
 	vendirEntireDirPath = "."
 )
```
praveenrewar: Should we keep these separate?

```suggestion
	ImageRegistry
)

const (
	vendirEntireDirPath = "."
)
```

pkg/fetch/vendir.go
```diff
@@ -421,3 +428,24 @@ func ExtractImageRegistry(name string) string {
...
+func extractGitHostname(input string) string {
+	u, err := url.Parse(input)
+	if err != nil {
+		return "github.com"
+	}
+	return u.Hostname()
+}
+
+// ExtractHost return registry for Docker Image and Host for git url
+func ExtractHost(input string, sourceType int) string {
```
rohitagg2020, line 442: If this is internal function, we can rename it to start with smallcase.

pkg/fetch/vendir.go
```diff
@@ -421,3 +428,24 @@ func ExtractImageRegistry(name string) string {
 	}
 	return registry
 }
+
+// extractGitHostname extracts the hostname from the git URL.
+func extractGitHostname(input string) string {
+	u, err := url.Parse(input)
+	if err != nil {
+		return "github.com"
```
rohitagg2020: If there is an error, why are we returning `github.com`?

pkg/fetch/vendir.go
praveenrewar: It will be used in the test (which is a different package)

pkg/fetch/vendir.go
sethiyash: I was thinking to default to github.com but we can return empty as well

pkg/fetch/vendir_test.go
```diff
@@ -58,32 +58,95 @@ func Test_AddDir_skipsTLS(t *testing.T) {
...
+	}
+	for i, tc := range testCases {
+		err = vendir.AddDir(v1alpha1.AppFetch{
+			Git: &v1alpha1.AppFetchGit{URL: tc.URL},
+		},
+			"dirpath/0")
+		assert.NoError(t, err)
+
+		vConf := vendir.Config()
+		t.Logf("%+v", vConf)
```
praveenrewar: Remove log statements

praveenrewar (approved):
LGTM! (Apart from the minor comment)

PR #1426, opened 2023-12-15
Make bits of kctrl more configurable

### LINKED ISSUE #1425: Make bits of kctrl more configurable
I am consuming bits of kctrl and want it to be more configurable. Allowing:
- Disabling waiting behaviour
- Allowing supplying a default value for a service account
- Allowing creation of resources in shared namespaces 
- Allowing a default cube context override

I should be allowed to configure these bits while attaching the package command tree

This allows kctrl pieces to interact with Kubernetes-like api servers better

---
Vote on this request

100mik:

Fixes #1425 

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/cli/pkg/kctrl/cmd/app/delete.go
+++ b/cli/pkg/kctrl/cmd/app/delete.go
@@ -57,6 +57,7 @@ func NewDeleteCmd(o *DeleteOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Co
 		AllowDisableWait: true,
 		DefaultInterval:  1 * time.Second,
 		DefaultTimeout:   5 * time.Minute,
+		WaitByDefault:    true,
 	})
 
 	return cmd
--- a/cli/pkg/kctrl/cmd/app/kick.go
+++ b/cli/pkg/kctrl/cmd/app/kick.go
@@ -50,6 +50,7 @@ func NewKickCmd(o *KickOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Comman
 		AllowDisableWait: true,
 		DefaultInterval:  1 * time.Second,
 		DefaultTimeout:   5 * time.Minute,
+		WaitByDefault:    true,
 	})
 
 	return cmd
--- a/cli/pkg/kctrl/cmd/core/config_factory.go
+++ b/cli/pkg/kctrl/cmd/core/config_factory.go
@@ -29,6 +29,9 @@ type ConfigFactoryImpl struct {
 
 	qps   float32
 	burst int
+
+	defaultKubeconfigOverridePath    string
+	defaultKubeconfigOverrideContext string
 }
 
 var _ ConfigFactory = &ConfigFactoryImpl{}
@@ -47,6 +50,11 @@ func (f *ConfigFactoryImpl) ConfigureContextResolver(resolverFunc func() (string
 
 func (f *ConfigFactoryImpl) ConfigureYAMLResolver(resolverFunc func() (string, error)) {
 	f.yamlResolverFunc = resolverFunc
+}
+
+func (f *ConfigFactoryImpl) ConfigureKubeconfigOverrides(defaultKubeconfigOverridePath string, defaultKubeconfigOverrideContext string) {
+	f.defaultKubeconfigOverridePath = defaultKubeconfigOverridePath
+	f.defaultKubeconfigOverrideContext = defaultKubeconfigOverrideContext
 }
 
 func (f *ConfigFactoryImpl) ConfigureClient(qps float32, burst int) {
@@ -122,6 +130,11 @@ func (f *ConfigFactoryImpl) clientConfig() (bool, clientcmd.ClientConfig, error)
 	// Based on https://github.com/kubernetes/kubernetes/blob/30c7df5cd822067016640aa267714204ac089172/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/config_flags.go#L124
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
 	overrides := &clientcmd.ConfigOverrides{}
+
+	if len(path) == 0 && len(context) == 0 && f.defaultKubeconfigOverrideContext != "" && f.defaultKubeconfigOverridePath != "" {
+		path = f.defaultKubeconfigOverridePath
+		context = f.defaultKubeconfigOverrideContext
+	}
 
 	if len(path) > 0 {
 		loadingRules.ExplicitPath = path
--- a/cli/pkg/kctrl/cmd/core/examples.go
+++ b/cli/pkg/kctrl/cmd/core/examples.go
@@ -14,6 +14,13 @@ type PackageCommandTreeOpts struct {
 
 	Color bool
 	JSON  bool
+
+	DefaultKubeconfigOverridePath    string
+	DefaultKubeconfigOverrideContext string
+
+	DefaultServiceAcccountName string
+	WaitByDefault              bool
+	AllowSharedNamespaces      bool
 }
 
 type Example struct {
--- a/cli/pkg/kctrl/cmd/core/secure_namespace_flags.go
+++ b/cli/pkg/kctrl/cmd/core/secure_namespace_flags.go
@@ -18,8 +18,8 @@ type SecureNamespaceFlags struct {
 	AllowedSharedNamespaces bool
 }
 
-func (s *SecureNamespaceFlags) Set(cmd *cobra.Command) {
-	cmd.Flags().BoolVar(&s.AllowedSharedNamespaces, "dangerous-allow-use-of-shared-namespace", false, "Allow use of shared namespaces")
+func (s *SecureNamespaceFlags) Set(cmd *cobra.Command, defaultVal bool) {
+	cmd.Flags().BoolVar(&s.AllowedSharedNamespaces, "dangerous-allow-use-of-shared-namespace", defaultVal, "Allow use of shared namespaces")
 }
 
 func (s *SecureNamespaceFlags) CheckForDisallowedSharedNamespaces(namespace string) error {
--- a/cli/pkg/kctrl/cmd/core/wait_flags.go
+++ b/cli/pkg/kctrl/cmd/core/wait_flags.go
@@ -19,13 +19,14 @@ type WaitFlagsOpts struct {
 	AllowDisableWait bool
 	DefaultInterval  time.Duration
 	DefaultTimeout   time.Duration
+	WaitByDefault    bool
 }
 
 func (f *WaitFlags) Set(cmd *cobra.Command, flagsFactory FlagsFactory, opts *WaitFlagsOpts) {
-	if opts.AllowDisableWait {
-		cmd.Flags().BoolVar(&f.Enabled, "wait", true, "Wait for reconciliation to complete")
+	if opts.AllowDisableWait || !opts.WaitByDefault {
+		cmd.Flags().BoolVar(&f.Enabled, "wait", opts.WaitByDefault, "Wait for reconciliation to complete")
 	}
-	f.Enabled = true
+	f.Enabled = opts.WaitByDefault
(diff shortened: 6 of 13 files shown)


cli/pkg/kctrl/cmd/core/examples.go
```diff
@@ -14,6 +14,12 @@ type PackageCommandTreeOpts struct {
 
 	Color bool
 	JSON  bool
+
+	DefaultKubeconfigOverridePath    string
+	DefaultKubeconfigOverrideContext string
+	DefaultServiceAcccountName       string
```
praveenrewar: nit: Some new lines to separate different types of flags.

cli/pkg/kctrl/cmd/core/kubeconfig_flags.go
```diff
@@ -31,6 +31,11 @@ func (f *KubeconfigFlags) Set(cmd *cobra.Command, flagsFactory FlagsFactory, opt
 	kubeconfigYamlEnvVariableKey := fmt.Sprintf("%s_KUBECONFIG_YAML", strings.ToUpper(opts.BinaryName))
 	f.YAML = NewKubeconfigYAMLFlag(kubeconfigYamlEnvVariableKey)
 	cmd.PersistentFlags().Var(f.YAML, "kubeconfig-yaml", fmt.Sprintf("Kubeconfig contents as YAML ($%s)", kubeconfigYamlEnvVariableKey))
+
+	if opts.DefaultKubeconfigOverrideContext != "" && opts.DefaultKubeconfigOverridePath != "" {
```
praveenrewar: Is it possible that someone might set one of these and not both? (Based on the individual flag name, they shouldn't depend on each other)

cli/pkg/kctrl/cmd/core/secure_namespace_flags.go
```diff
@@ -18,7 +18,7 @@ type SecureNamespaceFlags struct {
 	AllowedSharedNamespaces bool
 }
 
-func (s *SecureNamespaceFlags) Set(cmd *cobra.Command) {
+func (s *SecureNamespaceFlags) Set(cmd *cobra.Command, defaultVal bool) {
 	cmd.Flags().BoolVar(&s.AllowedSharedNamespaces, "dangerous-allow-use-of-shared-namespace", false, "Allow use of shared namespaces")
```
praveenrewar: defaultVal should be used, right?
```suggestion
	cmd.Flags().BoolVar(&s.AllowedSharedNamespaces, "dangerous-allow-use-of-shared-namespace", defaultVal, "Allow use of shared namespaces")
```

cli/pkg/kctrl/cmd/kctrl.go
```diff
@@ -76,7 +76,8 @@ func NewKctrlCmd(o *KctrlOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Comm
 		cmdcore.RestOfCommandsHelpGroup,
 	}))
 
-	pkgOpts := cmdcore.PackageCommandTreeOpts{BinaryName: "kctrl", PositionalArgs: false, Color: true, JSON: true}
+	pkgOpts := cmdcore.PackageCommandTreeOpts{BinaryName: "kctrl", PositionalArgs: false, Color: true, JSON: true,
```
praveenrewar: nit: I think we can remove `PositionalArgs` and `AllowSharedNamespaces` as the default value should be false anyway.

cli/pkg/kctrl/cmd/package/installed/delete.go
```diff
@@ -75,6 +75,7 @@ func NewDeleteCmd(o *DeleteOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Co
 		AllowDisableWait: false,
 		DefaultInterval:  1 * time.Second,
 		DefaultTimeout:   5 * time.Minute,
+		DisableByDefault: o.pkgCmdTreeOpts.DisableWaitByDefault,
```
praveenrewar: This does conflict with `AllowDisableWait` in case of package installed delete 🤔

praveenrewar: Also, I think we need to add it to package repository add_or_update, kick and delete?

cli/pkg/kctrl/cmd/core/kubeconfig_flags.go
100mik: The expectation is that if you are consuming these they are both submitted. For the use case these are supposed to address, it is reasonable to expect both at all times.

100mik: It might make sense to separate these none the less, even we do not have a case for it now

cli/pkg/kctrl/cmd/core/secure_namespace_flags.go
100mik: Yep, will make this change

cli/pkg/kctrl/cmd/kctrl.go
100mik: Not related to the current changes, but yeah that makes sense, I will make the change

cli/pkg/kctrl/cmd/package/installed/delete.go
100mik: Yeah I think we need it on kick and delete as well. Let me think about cases that it might run into with `AllowDisableWait`. We need to fallback to false and not allow a change. 
Should be noted that we will ONLY use this value in cases where we are not creating RBAC resources and a default SA is provided. I will try and take note of that someplace

cli/pkg/kctrl/cmd/core/kubeconfig_flags.go
prembhaskal: I think if we want to handle precedence order of 1st --kubeconfig, then KUBECONFIG and then any other (eg. tanzu cli context), then we need to handle the flags differently here.  we will need to check if nothing else is set, then we set the defaults.

100mik: Made this change, do take a peek! I will test out the precedence as well. We should account for this in our tests

cli/pkg/kctrl/cmd/package/installed/delete.go
100mik: Changed it to `UnsetByDefault`.
Behaviour being: The older bool decides whether or not the flag is attached at all and the newer one decides the default value.
However, the older flag takes effect only if the newer one is unset.

cli/pkg/kctrl/cmd/core/config_factory.go
```diff
@@ -123,6 +131,11 @@ func (f *ConfigFactoryImpl) clientConfig() (bool, clientcmd.ClientConfig, error)
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
 	overrides := &clientcmd.ConfigOverrides{}
 
+	if len(path) == 0 && len(context) == 0 && f.defaultKubeconfigOverrideContext != "" && f.defaultKubeconfigOverridePath != "" {
```
praveenrewar, line 134: No strong opinions but do we want separate these conditions?

cli/pkg/kctrl/cmd/core/examples.go
```diff
@@ -14,6 +14,13 @@ type PackageCommandTreeOpts struct {
 
 	Color bool
 	JSON  bool
+
+	DefaultKubeconfigOverridePath    string
+	DefaultKubeconfigOverrideContext string
+
+	DefaultServiceAcccountName string
+	UnsetWaitByDefault         bool
```
praveenrewar: I actually liked the previous name for this flag more, then `Unset` part in `UnsetWaitByDefault` sounds a little bit redundant.

cli/pkg/kctrl/cmd/core/config_factory.go
100mik: Maybe just throw in a new line?

cli/pkg/kctrl/cmd/core/examples.go
100mik: Trying to keep the language right 🤔 
I could flip the logic and make it `WaitByDefault`?

cli/pkg/kctrl/cmd/core/config_factory.go
praveenrewar: I think it's fine this way, I was just thinking if we want to keep those conditions separate.

cli/pkg/kctrl/cmd/core/examples.go
praveenrewar: Yeah, that sounds more intuitive.

100mik: My intention was to keep the changes in the other files to a minimum. Flipping the logic would mean we touch more files. Do you have a better idea for the name? I could just make it `DeactivateWaitByDefault`

cli/pkg/kctrl/cmd/core/wait_flags.go
```diff
@@ -19,13 +19,14 @@ type WaitFlagsOpts struct {
...
 }
 
 func (f *WaitFlags) Set(cmd *cobra.Command, flagsFactory FlagsFactory, opts *WaitFlagsOpts) {
-	if opts.AllowDisableWait {
-		cmd.Flags().BoolVar(&f.Enabled, "wait", true, "Wait for reconciliation to complete")
+	if opts.AllowDisableWait || !opts.WaitByDefault {
+		cmd.Flags().BoolVar(&f.Enabled, "wait", !opts.WaitByDefault, "Wait for reconciliation to complete")
 	}
-	f.Enabled = true
+	f.Enabled = !opts.WaitByDefault
```
praveenrewar: It should be `opts.WaitByDefault` instead of `!opts.WaitByDefault`, right?

cli/pkg/kctrl/cmd/app/delete.go
```diff
@@ -57,6 +57,7 @@ func NewDeleteCmd(o *DeleteOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Co
 		AllowDisableWait: true,
 		DefaultInterval:  1 * time.Second,
 		DefaultTimeout:   5 * time.Minute,
+		WaitByDefault:    true,
```
praveenrewar, line 60: We should be using `o.pkgCmdTreeOpts.WaitByDefault`, right?

cli/pkg/kctrl/cmd/app/kick.go
```diff
@@ -50,6 +50,7 @@ func NewKickCmd(o *KickOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Comman
 		AllowDisableWait: true,
 		DefaultInterval:  1 * time.Second,
 		DefaultTimeout:   5 * time.Minute,
+		WaitByDefault:    true,
```
praveenrewar, line 53: Same as above.

cli/pkg/kctrl/cmd/core/wait_flags.go
100mik: Ah yes, missed that

cli/pkg/kctrl/cmd/app/delete.go
100mik: For this one and the next we will always tail, these are the app commands

praveenrewar (approved):

cli/pkg/kctrl/cmd/package/repository/add_or_update.go
```diff
@@ -84,6 +84,7 @@ func NewAddCmd(o *AddOrUpdateOptions, flagsFactory cmdcore.FlagsFactory) *cobra.
 		AllowDisableWait: true,
 		DefaultInterval:  1 * time.Second,
 		DefaultTimeout:   5 * time.Minute,
+		WaitByDefault:    o.pkgCmdTreeOpts.WaitByDefault,
```
praveenrewar, line 87: This needs to be added for the update command as well, this is the reason for the failing test.

PR #1435, opened 2024-01-06
updating default value for apply-timeout option for kapp to 5mins

prembhaskal:
Existing default value for kapp apply-timeout is 15minutes.
Reducing it to a reasonable time to reduce issues where controller queue is hogged by a very slow or stuck apps (for eg. with incorrect conditions which cannot be met).  
Reduced timeout will ensure queue is freed up faster.
In rare cases, if packages need higher apply-timeout, then it can be fed in as part of Package manifests.

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -164,11 +164,17 @@ func (gc *Config) KappDeployRawOptions() []string {
 	gc.dataLock.RLock()
 	defer gc.dataLock.RUnlock()
 
+	kappOptions := make([]string, 0)
+
 	// Configure kapp to keep only 5 app changes as it seems that
 	// larger number of ConfigMaps negative affects other controllers on the cluster.
 	// Eventually kapp can be smart enough to keep minimal number of app changes.
 	// Set default first so that it can be overridden by user provided options.
-	return append([]string{"--app-changes-max-to-keep=5"}, gc.data.kappDeployRawOptions...)
+	// return append([]string{"--app-changes-max-to-keep=5"}, gc.data.kappDeployRawOptions...)
+	kappOptions = append(kappOptions, "--app-changes-max-to-keep=5")
+	kappOptions = append(kappOptions, "--apply-timeout=5m")
+	kappOptions = append(kappOptions, gc.data.kappDeployRawOptions...)
+	return kappOptions
 }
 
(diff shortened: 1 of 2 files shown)


prembhaskal:
I think test failure is flaky I am checking that.

pkg/config/config.go
```diff
@@ -164,11 +164,17 @@ func (gc *Config) KappDeployRawOptions() []string {
...
+	kappOptions := make([]string, 0)
+
 	// Configure kapp to keep only 5 app changes as it seems that
 	// larger number of ConfigMaps negative affects other controllers on the cluster.
 	// Eventually kapp can be smart enough to keep minimal number of app changes.
 	// Set default first so that it can be overridden by user provided options.
-	return append([]string{"--app-changes-max-to-keep=5"}, gc.data.kappDeployRawOptions...)
+	// return append([]string{"--app-changes-max-to-keep=5"}, gc.data.kappDeployRawOptions...)
+	kappOptions = append(kappOptions, "--app-changes-max-to-keep=5")
+	kappOptions = append(kappOptions, "--apply-timeout=5m")
```
rohitagg2020, line 175: I think you are planning to change the `wait-timeout` and not `apply-timeout`?

praveenrewar: The idea of changing apply-timeout is to avoid scenarios where we wait for a resource with the exists annotation for a very long time (_exists_ is an operation so the waiting for the resource to exists is accounted for in the apply-timeout)

neil-hickey (commented):
LGTM, I will defer to @praveenrewar / @rohitagg2020 about the specific kapp flag. But i'm good with the changes

praveenrewar (approved):
LGTM!
In kapp, the apply-timeout is only accounted for the time when a resource is being applied, which is milliseconds for most of the resources, so even if it's a very large package, we wouldn't hit 5 minutes. The only exception to this is when a resource has the `kapp.k14s.io/exists` annotation, in which case kapp would wait for the resource to _exists_, we wouldn't want an App to block kapp-controller queue due to any such resources, hence a reduced apply-timeout should reduce the time an App would spend time during reconciliation.

PR #1483, opened 2024-02-13
fix: update-kind-example-to-containerd_1.5

dmitrynovik:
Adjusted the example of running the `kapp-controller` in `kind` for the `containerd 1.5+`

The example supplied in this repository, [kapp-controller in kind](https://github.com/carvel-dev/kapp-controller/tree/develop/examples/local-kind-environment), **no longer works** with my `containerd://1.7.1`  - the pods are scheduled and created, but there are image pull errors as every pod tries to pull image over HTTPS instead of HTTP ( as in this [StackOverflow](https://stackoverflow.com/questions/74592879/error-deploying-image-to-kubernetes-pod-http-server-gave-http-response-to-htt) example).

I believe this change happened in [containerd v1.5](https://github.com/containerd/containerd/blob/main/docs/hosts.md#bypass-tls-verification-example) as each container needs a `hosts.toml` file.
This new requirement is also reflected in project kind: [create a local registry](https://kind.sigs.k8s.io/docs/user/local-registry/), note a new step which is missing in the 2 year old [kind-with-registry.sh](https://github.com/carvel-dev/kapp-controller/blob/develop/examples/local-kind-environment/scripts/kind-with-registry.sh) script:
```
# 3. Add the registry config to the nodes
#
# This is necessary because localhost resolves to loopback addresses that are
# network-namespace local.
# In other words: localhost in the container is not localhost on the host.
#
# We want a consistent name that works from both ends, so we tell containerd to
# alias localhost:${reg_port} to the registry container when pulling images
REGISTRY_DIR="/etc/containerd/certs.d/localhost:${reg_port}"
for node in $(kind get nodes); do
  docker exec "${node}" mkdir -p "${REGISTRY_DIR}"
  cat <<EOF | docker exec -i "${node}" cp /dev/stdin "${REGISTRY_DIR}/hosts.toml"
[host."http://${reg_name}:5000"]
EOF
done
```

Fixes #

NONE
-->
```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/examples/local-kind-environment/scripts/kind-with-registry.sh
+++ b/examples/local-kind-environment/scripts/kind-with-registry.sh
@@ -2,6 +2,8 @@
 set -o errexit
 
 # create registry container unless it already exists
+# http://kind-registry.local:5000
+# curl -s -X GET kind-registry.local:5000/v2/_catalog | jq .
 reg_name='kind-registry.local'
 reg_port='5000'
 if [ "$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)" != 'true' ]; then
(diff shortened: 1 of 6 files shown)


examples/local-kind-environment/scripts/kind-with-registry.sh
```diff
@@ -16,10 +16,26 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 containerdConfigPatches:
 - |-
-  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:${reg_port}"]
+  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."${reg_name}:${reg_port}"]
     endpoint = ["http://${reg_name}:5000"]
 EOF
 
+# 3. Add the registry config to the nodes
```
praveenrewar: nit: Remove `3. ` (or add numbers to the remaining steps).

praveenrewar (approved):
Thank you for making the contribution. Could you please sign your commits to make the DCO bot happy, we can merge the PR then.

examples/local-kind-environment/scripts/kind-with-registry.sh
dmitrynovik: Done.

praveenrewar:
@dmitrynovik I think the rebase didn't work so there are some commits coming up from develop. Could you please take a look at it? 
(You might have to do a reset before rebase and then rebase again)

PR #1598, opened 2024-07-18
Bump controller-runtime

praveenrewar:
- Bump controller-runtime
- Bump go
- Bump k8s libraries


rohitagg2020:
While bumping up the controller-runtime library (which also bump k8s.io/apiserver), it is observed that for k8s < 1.30, Create/Update/Delete doesn't work on Package and Packagemetadata resource while everything works perfectly for App CR, Package Repository and Package Install resource.

#### What makes Package and PackageMetadata different from other resources?
Package and PackageMetadata CRUD calls are handled via the Aggregated API Server while for others, we have reconcilers.

#### Why this behavior is observed?
While performing Create/Delete/Update, a validation function (e.g createValidation [here](https://github.com/carvel-dev/kapp-controller/blob/develop/pkg/apiserver/registry/datapackaging/package_crd_rest.go#L78)) executes. This function tries to initialise informer on v1.ValidatingAdmissionPolicy (https://github.com/kubernetes/apiserver/blob/v0.31.2/pkg/admission/plugin/policy/validating/plugin.go#L107). But as the v1.ValidatingAdmissionPolicy GVK doesn't exist on K8s < 1.30, validation fails and we get the following error: `Error from server (Forbidden): error when creating "packages/simple-app.corp.com/1.0.0.yml": packages.data.packaging.carvel.dev "simple-app.corp.com.1.0.0" is forbidden: not yet ready to handle request`. As the same GVK exist on K8s >= 1.30, we dont see any issues there.

#### Possible Way forward:
1. For K8s < 1.30, we do not run the validation functions.
	**Pros**: It should be easy to implement. Have to try out though.
	**Cons**: Any validating webhooks will not be executed during CUD of the Package/PackageMetadata.

2. While initialising the APIServer, we can see if there is a mechanism to skip the Validation of v1.ValidatingAdmissionPolicy and v1.ValidatingAdmissionPolicyBinding in case K8s < 1.30.
	**Pros**: The validating webhooks works for K8s version < 1.30.
	**Cons**: It looks time consuming task and also not sure if it is achievable.

PR #1623, opened 2024-09-10
Updated the golang to 1.22.6 the latest available stable release

yashsingh74:
The current version of Go used in kapp-controller is golang:[1.21.11](https://github.com/carvel-dev/kapp-controller/blob/develop/Dockerfile#L1), which is outdated and no longer aligned with the latest Kubernetes release.

Fixes #https://github.com/carvel-dev/kapp-controller/issues/1617


praveenrewar:
@yashsingh74 Bumping Go would require bumping some other libraries, including controller-runtime, there is a WIP [PR](https://github.com/carvel-dev/kapp-controller/pull/1598) for it, but it might take some time to get that in.

PR #1676, opened 2025-01-22
fix: stop calling redundant rand.Seed

mamachanko:
see: https://tip.golang.org/doc/go1.20

Signed-off-by: Max Brauer <mbrauer@vmware.com>


100mik (approved):

100mik:
Thanks for this!

PR #1751, opened 2025-08-19
Fix PKGI reconciliation taking old App status during upgrade

mdzhigarov:
## Problem
Fix race condition in PackageInstall reconciler where status was set based on App state before updating the App spec, leading to temporary inconsistencies.

## Solution
- Refactored `reconcileAppWithPackage` to accept a status updater closure
- Status is now set based on the App state AFTER any spec updates
- Eliminates race condition where PackageInstall shows stale status

## Changes
- Modified `reconcile()` method to create status updater closure
- Updated `reconcileAppWithPackage()` to accept and call the closure with correct App instance
- Added comprehensive tests demonstrating the fix

## Testing
- Added `Test_StatusUpdaterClosureUsesUpdatedAppState` - tests the race condition scenario
- Added `Test_StatusUpdaterClosureWithNoAppUpdate` - tests non-update scenario
- All existing tests continue to pass

Fixes: (https://vmw-jira.broadcom.net/browse/TNZ-54843)

### DIFF
--- a/pkg/packageinstall/packageinstall_test.go
+++ b/pkg/packageinstall/packageinstall_test.go
@@ -20,10 +20,12 @@ import (
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/version"
 	fakediscovery "k8s.io/client-go/discovery/fake"
 	"k8s.io/client-go/kubernetes/fake"
+	k8stesting "k8s.io/client-go/testing"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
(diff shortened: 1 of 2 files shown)


praveenrewar (commented):
Thanks for the PR Marin!
(Took a quick peek) This is the fix that I had in my mind as well, I will go through the tests in detail once the checks are green.
I think @cppforlife had some more thoughts regarding this so let's wait for him to add his comments.

cppforlife (approved):

cppforlife:
lgtm from me. ~@praveenrewar care to take a look?~

pkg/packageinstall/packageinstall_test.go
```diff
@@ -688,6 +690,290 @@ func Test_PlaceHolderSecretCreated_WhenPackageInstallUpdated(t *testing.T) {
 	assert.Equal(t, "instl-pkg-fetch-0", app.Spec.Fetch[0].ImgpkgBundle.SecretRef.Name)
 }
 
+// Test_StatusUpdaterClosureWithAppUpdateFromReconcileFailed tests the exact bug scenario:
+// 1. App exists in ReconcileFailed state (generation == observedGeneration)
+// 2. PackageInstall gets updated to reference a new package version
+// 3. This updates the App spec (and in real K8s would increment generation)
+// 4. Before the bug fix: PackageInstall would show status based on old App state
+// 5. After the bug fix: PackageInstall status reflects the updated App state
```
praveenrewar: nit: Could you remove 'bug' from here. Usually we would mention/refer bugs when they are still open.

praveenrewar (approved):
@mdzhigarov Changes looks good to me apart from the small (optional) nit. 
My only concern is that we don't have a proper test for this. It might be tricky to add a test for this but even if we could validate that the change is fixing the issue would be a good to have.

praveenrewar (approved):

PR #1766, opened 2025-10-12
Add retry logic for App CR updates to handle optimistic concurrency conflicts

mdzhigarov:
## Summary

This PR implements retry logic for App CR updates in the PackageInstall reconciler to handle optimistic concurrency control conflicts that can occur when multiple operations attempt to update the same App CR simultaneously.

### Core Implementation
- **Added `updateAppWithRetry()` function** in `pkg/packageinstall/packageinstall.go`
  - Implements retry logic with up to 5 attempts for App CR updates
  - Re-fetches App CR after each failed update to get the latest resourceVersion
  - Handles NotFound errors immediately without retries to avoid unnecessary delays
  - Uses flexible `updateFunc` signature that returns an App, allowing direct integration with `NewApp()`

### Integration Points
- **Updated two App update locations** in PackageInstall reconciler:
  - Line 233: App creation/update during normal reconciliation (uses `NewApp()` for transformation)
  - Line 408: App update during deletion reconciliation (applies specific field changes)

### Testing
- **Added comprehensive unit tests** in `pkg/packageinstall/packageinstall_test.go`:
  - `Test_UpdateAppWithRetry_HandlesConflictErrors`: Tests retry logic with simulated conflict errors
  - `Test_UpdateAppWithRetry_HandlesNotFoundError`: Verifies immediate return for NotFound errors
  - `Test_UpdateAppWithRetry_HandlesUpdateFunctionError`: Tests error propagation from update function

## Problem Solved

Fixes the "object has been modified; please apply your changes to the latest version and try again" error that occurs when:
- Multiple PackageInstall reconcilers attempt to update the same App CR
- App CR is updated by other controllers or operations between fetch and update
- High-frequency reconciliation triggers concurrent updates

## Testing

All existing tests pass, and new tests specifically validate:
- ✅ Retry logic correctly handles optimistic concurrency conflicts
- ✅ NotFound errors are handled specially (no retries)
- ✅ Update function errors are properly propagated
- ✅ Retry mechanism works as designed (5 attempts max, re-fetch between attempts)

```bash
go test ./pkg/packageinstall -v
# All tests pass including new retry logic tests
```

## Backward Compatibility

This change is fully backward compatible:
- No API changes
- No breaking changes to existing behavior
- Only adds resilience to existing update operations

### DIFF
--- a/pkg/packageinstall/packageinstall.go
+++ b/pkg/packageinstall/packageinstall.go
@@ -406,7 +406,6 @@ func (pi *PackageInstallCR) reconcileDelete(modelStatus *reconciler.Status) (rec
 
 	if !equality.Semantic.DeepEqual(existingApp, unchangeExistingApp) {
 		updatedApp, err := pi.updateAppWithRetry(existingApp, func(app *kcv1alpha1.App) (*kcv1alpha1.App, error) {
-			// Apply the changes we made to existingApp
 			app.Spec.ServiceAccountName = existingApp.Spec.ServiceAccountName
 			app.Spec.Cluster = existingApp.Spec.Cluster
 			app.Spec.NoopDelete = existingApp.Spec.NoopDelete
@@ -564,13 +563,11 @@ func (pi PackageInstallCR) createSecretForSecretgenController(iteration int) (st
 func (pi *PackageInstallCR) updateAppWithRetry(app *kcv1alpha1.App, updateFunc func(*kcv1alpha1.App) (*kcv1alpha1.App, error)) (*kcv1alpha1.App, error) {
 	var lastErr error
 	for i := 0; i < 5; i++ {
-		// Apply the update function to get the desired app
 		desiredApp, err := updateFunc(app)
 		if err != nil {
 			return nil, fmt.Errorf("update function failed: %s", err)
 		}
 
-		// Try to update
 		updatedApp, err := pi.kcclient.KappctrlV1alpha1().Apps(desiredApp.Namespace).Update(
 			context.Background(), desiredApp, metav1.UpdateOptions{})
 		if err == nil {
@@ -578,12 +575,10 @@ func (pi *PackageInstallCR) updateAppWithRetry(app *kcv1alpha1.App, updateFunc f
 		}
 		lastErr = err
 
-		// If it's a NotFound error, return it immediately (don't retry)
-		if errors.IsNotFound(err) {
+		if !errors.IsConflict(err) {
 			return nil, err
 		}
 
-		// If failure occurs, re-fetch the App using Get
 		app, err = pi.kcclient.KappctrlV1alpha1().Apps(app.Namespace).Get(
 			context.Background(), app.Name, metav1.GetOptions{})
 		if err != nil {
--- a/pkg/packageinstall/packageinstall.go
+++ b/pkg/packageinstall/packageinstall.go
@@ -230,8 +230,9 @@ func (pi *PackageInstallCR) reconcileAppWithPackage(existingApp *kcv1alpha1.App,
 	var appToCheckStatus *kcv1alpha1.App = existingApp
 
 	if !equality.Semantic.DeepEqual(desiredApp, existingApp) {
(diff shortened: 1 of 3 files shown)


pkg/packageinstall/packageinstall.go
```diff
@@ -549,3 +559,37 @@ func (pi PackageInstallCR) createSecretForSecretgenController(iteration int) (st
 	}
 	return secretName, nil
 }
+
+// updateAppWithRetry updates an App CR with retry logic to handle conflicts
+func (pi *PackageInstallCR) updateAppWithRetry(app *kcv1alpha1.App, updateFunc func(*kcv1alpha1.App) (*kcv1alpha1.App, error)) (*kcv1alpha1.App, error) {
```
praveenrewar, line 563: I think this function do the following things
- Get the latest app
- Get the desired app using the latest app and the update function
- Update the app
- If conflict, repeat

(So we should get the latest everytime before an update instead of getting it when a conflict occurs)

pkg/packageinstall/packageinstall.go
```diff
@@ -404,14 +405,23 @@ func (pi *PackageInstallCR) reconcileDelete(modelStatus *reconciler.Status) (rec
 	existingApp.Spec.DefaultNamespace = pi.model.Spec.DefaultNamespace
 
 	if !equality.Semantic.DeepEqual(existingApp, unchangeExistingApp) {
-		existingApp, err = pi.kcclient.KappctrlV1alpha1().Apps(existingApp.Namespace).Update(
-			context.Background(), existingApp, metav1.UpdateOptions{})
+		updatedApp, err := pi.updateAppWithRetry(existingApp, func(app *kcv1alpha1.App) (*kcv1alpha1.App, error) {
```
praveenrewar, line 408: The existing app is already the desired app at this stage, so the updateFunction probably won't be doing anything since we are not fetching the app again.

mdzhigarov: At this point the existing app is indeed equal to the desired app.
But the update func is necessary because it is still possible that update will fail. If that happens - we need to re-fetch the app and apply the update func.

pkg/packageinstall/packageinstall.go
mdzhigarov: The current flow is a bit more performant cuz it takes the desired app as input, attempts to update and only if failure occurs - a re-fetch is done:

- Apply update function on the provided app as input
- Update the app
- if conflict - re-fetch the app, repeat.

Do you think this is a problem?

praveenrewar: I think it's fine.

praveenrewar (dismissed):
LGTM!

pkg/packageinstall/packageinstall.go
```diff
@@ -549,3 +559,37 @@ func (pi PackageInstallCR) createSecretForSecretgenController(iteration int) (st
...
+		// Try to update
+		updatedApp, err := pi.kcclient.KappctrlV1alpha1().Apps(desiredApp.Namespace).Update(
+			context.Background(), desiredApp, metav1.UpdateOptions{})
+		if err == nil {
+			return updatedApp, nil
+		}
+		lastErr = err
+
+		// If it's a NotFound error, return it immediately (don't retry)
+		if errors.IsNotFound(err) {
```
praveenrewar: Do you think we should check if the error is due to a conflict and return otherwise?

mdzhigarov: I thought about it but at the end decided against it. Mainly because it's a bit hard to reproduce this issue in a real environment to make sure that indeed the api server returns http 409 Conflict in this case. But also because it may be beneficial to retry on other errors - i.e. intermitten timeout.

What do you think?

praveenrewar: My only concern is that there could be many reasons for the failure and it wouldn't be ideal to retry on all of them (and I am not sure if any of them can have any side effects)

mdzhigarov: Makes sense. Let's be on the safe side then and retry only on 409 conflict errors.

praveenrewar (approved):

PR #1840, opened 2026-06-26
[1839] Fix: Truncate status.deploy.stdout as it can exceed etcd's limit on large clusters, causing reconciliation delay

### LINKED ISSUE #1839: status.deploy.stdout can exceed etcd's 2 MiB limit on large clusters, causing reconciliation delays
**What steps did you take:**
[A clear and concise description steps that can be used to reproduce the problem.]

- Run kapp-controller on a cluster with 250+ nodes.
- Deploy an App CR that manages a large number of resources (e.g. DaemonSets or cluster-wide resources that produce one change entry per node in kapp's output).
- Observe the App enters a ReconcileFailed loop and takes ~40 minutes to eventually show Reconcile succeeded.

**What happened:**

When kapp deploys a large number of resources it produces verbose stdout — on a 250-node cluster this can reach several MiB. kapp-controller writes this verbatim into status.deploy.stdout and then calls the Kubernetes API to persist the status. etcd rejects objects larger than its hard 2 MiB gRPC message limit, causing the UpdateStatus call to fail.

The deployment itself has succeeded; only the status write fails. On the next reconcile cycle kapp produces a much smaller "no-op" diff that fits within the limit, so the status update eventually goes through — but only after repeated failed retries, resulting in a ~40-minute apparent delay.

**What did you expect:**

The status.deploy.stdout (and other output fields such as status.fetch.stdout, status.inspect.stdout, and status.usefulErrorMessage) should be bounded so that the status object never exceeds etcd's 2 MiB limit regardless of cluster size. When output is clipped the field should begin with a clear truncation notice (e.g. [output truncated]\n) so operators immediately know the field is incomplete, rather than silently truncating without any indication.

The issue is non-fatal — reconciliation will eventually succeed — but the delay is significant in production (observed: ~40 minutes) and can mask real problems if operators treat ReconcileFailed as a signal to investigate.
The fix is to truncate individual output fields to a reasonable limit (e.g. 1 MiB per field) before writing them into the status struct. Keeping the tail of the output is preferable because the most actionable content (final resource summary, error lines) always appears at the end of kapp output.
Other status string fields that may also grow large in failure scenarios: status.deploy.stderr, status.fetch.stderr, status.usefulErrorMessage

**Environment:**

- kapp Controller version (execute `kubectl get deployment -n kapp-controller kapp-controller -o yaml` and the annotation is `kbld.k14s.io/images`):any version
- Kubernetes version (use `kubectl version`) any version; effect is most pronounced on 250+ node clusters

---
Vote on this request

himsngh:

On large clusters (>~250 nodes), a kapp deploy can produce several MiB of change output (one line per resource per node). kapp-controller writes this verbatim into status.deploy.stdout and then calls UpdateStatus on the App CR. etcd rejects objects larger than its hard 2 MiB gRPC message limit, causing the status write to fail repeatedly. The deployment itself has already succeeded, only the status update is failing but the App stays in ReconcileFailed until a subsequent no-op reconcile produces a small enough diff to fit.

This PR fixes the issue by truncating individual status output fields to 1 MiB before they are written into the status struct. When truncation occurs the field is prefixed with [output truncated]\n so operators immediately know the field is incomplete rather than receiving silently clipped output. The tail of the output is kept (not the head) because the most actionable content, the final resource summary and any error lines, always appears at the end of kapp output.

Fields covered: status.deploy.stdout, status.deploy.stderr, status.fetch.stdout, status.fetch.stderr, status.inspect.stdout, status.inspect.stderr, status.usefulErrorMessage.

Fixes #1839 

```release-note
NA
```

##### Review Checklist:

- [-] Relevant docs in this repo added or updated
- [-] Relevant carvel.dev docs added or updated in a separate PR and there's
  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -40,6 +40,9 @@ type Hooks struct {
 type Opts struct {
 	DefaultSyncPeriod time.Duration
 	MinimumSyncPeriod time.Duration
+	// MaxStatusOutputBytes caps each stdout/stderr field written into App status.
+	// Zero means use defaultMaxStatusOutputBytes (1 MiB).
+	MaxStatusOutputBytes int
 }
 
 type App struct {
--- a/pkg/app/app_reconcile.go
+++ b/pkg/app/app_reconcile.go
@@ -127,9 +127,10 @@ func (a *App) reconcileFetchTemplateDeploy() exec.CmdRunResult {
 		var fetchResult exec.CmdRunResult
 		assetsPath, fetchResult = a.fetch(assetsPath)
 
+		fetchResult = fetchResult.WithTruncatedStrings(a.maxOutputBytes())
 		a.app.Status.Fetch = &v1alpha1.AppStatusFetch{
-			Stderr:    truncateOutput(fetchResult.Stderr, a.maxOutputBytes()),
-			Stdout:    truncateOutput(fetchResult.Stdout, a.maxOutputBytes()),
+			Stderr:    fetchResult.Stderr,
+			Stdout:    fetchResult.Stdout,
 			ExitCode:  fetchResult.ExitCode,
 			Error:     fetchResult.ErrorStr(),
 			StartedAt: a.app.Status.Fetch.StartedAt,
@@ -178,11 +179,11 @@ func (a *App) reconcileFetchTemplateDeploy() exec.CmdRunResult {
 }
 
 func (a *App) updateLastDeploy(result exec.CmdRunResult) exec.CmdRunResult {
-	result = result.WithFriendlyYAMLStrings()
+	result = result.WithFriendlyYAMLStrings().WithTruncatedStrings(a.maxOutputBytes())
 
 	a.app.Status.Deploy = &v1alpha1.AppStatusDeploy{
-		Stdout:           truncateOutput(result.Stdout, a.maxOutputBytes()),
-		Stderr:           truncateOutput(result.Stderr, a.maxOutputBytes()),
+		Stdout:           result.Stdout,
+		Stderr:           result.Stderr,
 		Finished:         result.Finished,
 		ExitCode:         result.ExitCode,
 		Error:            result.ErrorStr(),
@@ -242,12 +243,12 @@ func (a *App) resetLastDeployStartedAt() {
 }
 
 func (a *App) reconcileInspect() error {
-	inspectResult := a.inspect().WithFriendlyYAMLStrings()
+	inspectResult := a.inspect().WithFriendlyYAMLStrings().WithTruncatedStrings(a.maxOutputBytes())
 
 	if !inspectResult.IsEmpty() {
 		a.app.Status.Inspect = &v1alpha1.AppStatusInspect{
-			Stdout:    truncateOutput(inspectResult.Stdout, a.maxOutputBytes()),
-			Stderr:    truncateOutput(inspectResult.Stderr, a.maxOutputBytes()),
+			Stdout:    inspectResult.Stdout,
+			Stderr:    inspectResult.Stderr,
 			ExitCode:  inspectResult.ExitCode,
 			Error:     inspectResult.ErrorStr(),
 			UpdatedAt: metav1.NewTime(time.Now().UTC()),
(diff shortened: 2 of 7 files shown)


aroradaman (commented):
I see we already have `WithFriendlyYAMLStrings` which returns modified `CmdRunResult`. Can we follow the same pattern here?, may be add `WithTruncatedStrings(size)`. 


(ref: https://github.com/carvel-dev/kapp-controller/blob/be1faefd135d62d901a0ad4b4904b30c6c0dc7c3/pkg/exec/cmd_run_result.go#L53-L63)

himsngh:
> I see we already have `WithFriendlyYAMLStrings` which returns modified `CmdRunResult`. Can we follow the same pattern here?, may be add `WithTruncatedStrings(size)`.
> 
> (ref:
> 
> https://github.com/carvel-dev/kapp-controller/blob/be1faefd135d62d901a0ad4b4904b30c6c0dc7c3/pkg/exec/cmd_run_result.go#L53-L63
> 
> )

Updated!

aroradaman (commented):
## Review — testing only

Focusing on the test changes here (code changes look good). I reviewed the three test files against the actual source (`WithFriendlyYAMLStrings`, the `NewApp` signature, existing e2e conventions), not just the diff.

### Overall
Unit-level testing is genuinely strong; the e2e test is the weak link.

### What's good
- **`pkg/exec/cmd_run_result_test.go` is excellent.** Table-driven and covers the cases that matter for `truncateOutput`: under-limit, exactly-at-limit, empty, over-limit tail-preservation, and two UTF-8 boundary cases (continuation byte skipped, pure multibyte runes producing valid UTF-8). It also correctly asserts the marker is *not* counted toward `maxBytes`. This is the core logic and it's well locked down.
- **`app_status_truncate_test.go`** exercises the real methods (`updateLastDeploy`, `setUsefulErrorMessage`) through a clean helper, and the tail-preservation test correctly accounts for the `WithFriendlyYAMLStrings` → `strings.TrimSpace` interaction (`TrimSpace` strips the trailing `\n`, so the comment is accurate and `Contains` is robust either way).
- Using a configurable `Opts.MaxStatusOutputBytes` so tests run with a limit of `50` instead of generating multi-MiB strings — fast and deterministic.

### Gaps / issues (roughly by importance)

1. **Fetch and inspect truncation paths are untested.** The diff adds `.WithTruncatedStrings(...)` in three status paths — deploy, **fetch** (`reconcileFetchTemplateDeploy`), and **inspect** (`reconcileInspect`) — but the tests only cover deploy and `usefulErrorMessage`. A regression that dropped the `WithTruncatedStrings` call from the fetch or inspect path would pass the entire suite. Please add assertions on `Status.Fetch.Stdout/Stderr` and `Status.Inspect.Stdout/Stderr`.

2. **The e2e test never exercises truncation.** It deploys 30 ConfigMaps (~KB), so `strings.HasPrefix(stdout, marker)` is always false and the "bounded" assertion is trivially true. It's really a happy-path *capture* test that overlaps existing deploy e2e coverage — the name `..._CapturedAndBounded` oversells what it verifies. The comment acknowledges generating >1 MiB in-cluster is impractical, which is fair — but the truncation limit isn't runtime-configurable (only via `Opts` in code), so there's no way to drive this path from e2e today. Options: rename to reflect it's a capture test, or make the cap configurable via a controller flag so e2e can set a tiny limit and assert the marker appears. As-is its incremental value over existing tests is low.

3. **The e2e encodes an incorrect invariant.** The test asserts `len(stdout) < 1 MiB` with the comment "must never exceed the 1 MiB cap." But `truncateOutput` returns `TruncationMarker + last maxBytes`, so a truncated field is up to `1 MiB + len(marker)` (≈1 MiB + 19 bytes). The assertion only passes today because output is tiny; if truncation ever actually occurred, this assertion would fail. The real bound is `len(marker) + maxBytes`.

4. **Default-limit fallback branch has no coverage.** Every test passes an explicit `Opts.MaxStatusOutputBytes`, so the `== 0 → defaultMaxStatusOutputBytes` branch of `maxOutputBytes()` is never hit. A one-liner asserting `maxOutputBytes()` returns 1 MiB when `Opts` is zero closes it.

5. **e2e indexes `cr.Status.Conditions[0]` without a non-empty guard.** If conditions were empty this panics (index out of range) instead of failing cleanly — prefer `require.NotEmpty(...)` first. Also, existing e2e tests (`app_status_test.go`) assert the full condition (Type/Status/Message); asserting only `.Type` is slightly less thorough but consistent enough.

6. **Minor:** `Test_WithTruncatedStrings` only covers the both-fields-over-limit case. A mixed case (one field under the limit passed through verbatim, the other truncated) would better lock per-field behavior. Low priority since `truncateOutput`'s passthrough is already tested.

### Bottom line
Unit tests for the truncation primitive are thorough and correct. Before merge I'd want: (1) fetch/inspect truncation assertions, (3) the incorrect 1 MiB invariant fixed, and (4) the default-fallback covered. The e2e test (2) is the judgment call — it's currently a capture test wearing a truncation-test name.

aroradaman (approved):

PR #1842, opened 2026-07-08
fix(apiserver): probe SA permissions before loading admission-policy plugins

### LINKED ISSUE #1841: Aggregated apiserver rejects all writes with "not yet ready to handle request" on k8s 1.36+ when SA lacks RBAC for MutatingAdmissionPolicy
**What steps did you take:**
  1. Install a kapp-controller build that vendors `k8s.io/apiserver v0.36.0+` on a Kubernetes 1.36+ cluster (reproduced on kind 1.36 and EKS 1.36).
  2. Ensure the deployed ClusterRole for the kapp-controller ServiceAccount does NOT include `list` + `watch` on `mutatingadmissionpolicies` and `mutatingadmissionpolicybindings` in the
  `admissionregistration.k8s.io` group. This state arises naturally when: (a) an operator upgrades kapp-controller in-place without re-applying the shipped ClusterRole, (b) a distribution ships its own ClusterRole
  derived from an older release, or (c) an operator has deliberately narrowed RBAC.
  3. Create a `PackageRepository` targeting any OCI bundle:
     kubectl apply -f - <<EOF
     apiVersion: packaging.carvel.dev/v1alpha1
     kind: PackageRepository
     metadata:
       name: repro
       namespace: default
     spec:
       fetch:
         imgpkgBundle:
           image: <any-valid-package-bundle-image>
     EOF
  4. Observe the reconciler state:
     kubectl describe packagerepository/repro

**What happened:**

The PackageRepository fails to reconcile with:

```
Reconcile failed: kapp: Error: create packagemetadata/<name> (data.packaging.carvel.dev/v1alpha1) namespace: default:
    Creating resource packagemetadata/<name> (data.packaging.carvel.dev/v1alpha1) namespace: default:
      API server says: packagemetadatas.data.packaging.carvel.dev "<name>" is forbidden:
      not yet ready to handle request (reason: Forbidden)
```

Every mutating request into `data.packaging.carvel.dev/v1alpha1` fails the same way. Reads (`get`, `list`, `watch`) continue to work. The kapp-controller pod reports `Ready 2/2`. The aggregated APIService reports
`Status.Available=True`. Pod logs contain a high-frequency stream of `Failed to watch *v1.MutatingAdmissionPolicy: ... is forbidden: User "system:serviceaccount:kapp-controller:kapp-controller-sa" cannot list resource "mutatingadmissionpolicies"`.

**Root cause:** 

`k8s.io/apiserver v0.36` added the `MutatingAdmissionPolicy` admission plugin to `RecommendedPluginOrder` by default, and the plugin's feature gate went GA-on in Kubernetes 1.36. On 1.36+ clusters the plugin loads automatically; its `SharedInformer` tries to list+watch `MutatingAdmissionPolicy` and `MutatingAdmissionPolicyBinding`; if RBAC blocks that, `HasSynced()` never returns true and every mutating
 admission decision returns `admission.NewForbidden(a, fmt.Errorf("not yet ready to handle request"))` from `vendor/k8s.io/apiserver/pkg/admission/plugin/policy/generic/plugin.go`. `pkg/apiserver/apiserver.go`
already guards each admission-policy plugin behind a discovery-based `serverResourceExists` check, but discovery only proves the resource exists in the cluster — not that the ServiceAccount can watch it. On 1.36+
the resource always exists, so the guard passes and the plugin loads even when RBAC is insufficient.

The same failure mode can affect `ValidatingAdmissionPolicy` on k8s 1.30+ when its RBAC is out of sync.

**What did you expect:**
Loading an admission-policy plugin whose backing SharedInformer cannot start should not permanently break the aggregated apiserver. The apiserver should either:

1. Not load the plugin when the ServiceAccount lacks the RBAC required to run its informers, and emit a clear, actionable warning naming the specific resource and verb that would need to be granted; OR
2. Surface the failure loudly at startup (fail-fast or prominent error-level log) rather than passing readiness checks and silently rejecting every write.

At a minimum, `APIService.Status.Available=True` should not be reported when the admission chain is permanently in the "not yet ready" state.

Suggested fix: extend the existing plugin guard in `pkg/apiserver/apiserver.go` to also probe the ServiceAccount's permissions via `SelfSubjectAccessReview` before loading the plugin. Concretely, replace the two `serverResourceExists(...)` calls in the plugin-order block with a new helper `canUseAdmissionPolicyPlugin(discoveryClient, coreClient, policyResource, bindingResource, logger)` that returns true only when:

- Both resources are present under `admissionregistration.k8s.io/v1` in discovery, AND
- A `SelfSubjectAccessReview` for `list` and for `watch` on each of the two resources comes back with `Allowed=true`.

When the check fails due to missing RBAC, log a single warning naming the specific `(resource, verb)` pair the SA is missing, so operators can grep and fix. The `SelfSubjectAccessReview` probe uses
  `subjectaccessreviews/create`, which is already granted in the shipped ClusterRole — no new RBAC required for the probe itself.

  The proposed fix keeps the aggregated apiserver functional across every supported k8s version regardless of the operator's RBAC state, and turns a silent 403-forever loop into a single actionable log line at startup.

**Environment:**

  - kapp-controller version: any build vendoring `k8s.io/apiserver v0.36.0+` (including `develop` at HEAD). Releases vendoring older `k8s.io/apiserver` do not exhibit this bug because the `MutatingAdmissionPolicy`
  plugin does not exist in those library versions.
  - Kubernetes version:
    Client Version: v1.36
    Server Version: v1.36

---
Vote on this request

neel5481:
Fixes #1841

On Kubernetes 1.36+ the MutatingAdmissionPolicy admission plugin ships enabled by default. Its SharedInformer requires list+watch permission on mutatingadmissionpolicies and mutatingadmissionpolicybindings. If the deployed ClusterRole for the kapp-controller ServiceAccount lacks those permissions, the informer's cache never syncs, HasSynced() never returns true, and every mutating admission request into the aggregated apiserver fails permanently with 'not yet ready to handle request (reason: Forbidden)'.

Signed-off-by: Neel Patel

The existing serverResourceExists guard only confirms the API resource is present in cluster discovery, which is always true on k8s 1.36+ regardless of RBAC. This change extends the guard with a SelfSubjectAccessReview probe for both list and watch on the policy resource and its bindings resource before loading each admission-policy plugin. When the probe fails the plugin is skipped and an actionable warning is logged naming the missing (resource, verb) pair.

The probe uses subjectaccessreviews/create, which is already granted in the shipped ClusterRole, no new RBAC required.

Also applied to ValidatingAdmissionPolicy for symmetry; the same failure mode is possible on k8s 1.30+ if RBAC drifts.

Fixes #1841

```NONE

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/apiserver/apiserver.go
+++ b/pkg/apiserver/apiserver.go
@@ -20,6 +20,7 @@ import (
 	kcclient "carvel.dev/kapp-controller/pkg/client/clientset/versioned"
 	"github.com/carvel-dev/semver/v4"
 	"github.com/go-logr/logr"
+	authorizationv1 "k8s.io/api/authorization/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -151,7 +152,7 @@ func NewAPIServer(clientConfig *rest.Config, coreClient kubernetes.Interface, kc
 		return nil, fmt.Errorf("building aggregation client: %v", err)
 	}
 
-	config, caContentProvider, err := newServerConfig(aggClient, opts)
+	config, caContentProvider, err := newServerConfig(aggClient, coreClient, opts)
 	if err != nil {
 		return nil, err
 	}
@@ -235,7 +236,7 @@ func (as *APIServer) isReady() (bool, error) {
 	return false, nil
 }
 
-func newServerConfig(aggClient aggregatorclient.Interface, opts NewAPIServerOpts) (*genericapiserver.RecommendedConfig, *dynamiccertificates.DynamicFileCAContent, error) {
+func newServerConfig(aggClient aggregatorclient.Interface, coreClient kubernetes.Interface, opts NewAPIServerOpts) (*genericapiserver.RecommendedConfig, *dynamiccertificates.DynamicFileCAContent, error) {
 
 	recommendedOptions := genericoptions.NewRecommendedOptions("", Codecs.LegacyCodec(v1alpha1.SchemeGroupVersion))
 	recommendedOptions.Etcd = nil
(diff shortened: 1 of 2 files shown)


joaopapereira (changes requested):
I do not think that this approach is correct because in the end we will allow the user to surpass the RBAC configured. In my opinion we should go with something like:
1. Fail readiness / don’t mark APIService Available when the resource exists in discovery but SSAR says no list/watch (or when the plugin is required and can’t sync).
2. Or fail startup with a clear error: grant list/watch on mutatingadmissionpolicies / bindings (and validating equivalents), or set an explicit “disable admission-policy plugins” flag.
3. Optional escape hatch: only skip if an explicit flag opt-out exists; default = refuse to serve writes.

I think that option 2 would be the safest approach for kapp-controller, but if we want to ensure that the controller at least is able to keep reconciling the AppCR 1 is a good approach.

Also we should ensure that in the release notes that we tell advise for CLusterRoles to be reapplied whenever deploying a new version of kapp-controller

PR #1846, opened 2026-08-11
fix: avoid FIPS 140-only panic when parsing sops PGP private keys

sameerforge:
**Problem**
When kapp-controller is built with Go's native FIPS 140-3 module and run with strict enforcement (GODEBUG=fips140=only), using the sops template's PGP backend panics while loading the user's private key:
`panic: crypto/sha1: use of SHA-1 is not allowed in FIPS 140-only mode`

This happens on every reconcile of an `App` that uses `spec.template[].sops.pgp`, making the PGP decryption path entirely unusable under strict FIPS mode.

**Root cause**
`gpgKeyring.Write (pkg/template/sops.go)` calls `golang.org/x/crypto/openpgp.ReadArmoredKeyRing` to parse the user-supplied private key. Parsing an OpenPGP key packet unconditionally computes its RFC 4880 V4 key fingerprint using SHA-1, as part of just building the in-memory key structure — regardless of whether that fingerprint is ever used for anything security-sensitive. SHA-1 is not an approved algorithm under FIPS 140-3, so under strict enforcement this call panics rather than returning an error.

**Fix**
Wrap the `openpgp.ReadArmoredKeyRing` call in `crypto/fips140.WithoutEnforcement` (available since Go 1.26), which scopes the enforcement relaxation to just that one call rather than disabling FIPS enforcement for the process:

This is safe because the fingerprint computed here is used only as a local identifier while loading the key for later use by the external sops binary — it does not verify or establish trust in any signature, so relaxing enforcement for this narrow, non-security parse step does not weaken any FIPS security guarantee.

`WithoutEnforcement` is a no-op everywhere except `GODEBUG=fips140=only`: it short-circuits to a direct call in non-FIPS builds and under `fips140=on/off/unset`, so this change has no effect outside of strict FIPS mode.

**Testing**
- Added `TestGpgKeyringWrite_UnderFIPS140Only` in `pkg/template/sops_test.go`, which loads a private PGP key and writes it out via `gpgKeyring.Write`. Confirmed it panics on the pre-fix code under `GODEBUG=fips140=only` and passes with the fix applied.
- Verified the fix behaves identically to the pre-fix code under `GODEBUG=fips140=on/off/unset` and in non-FIPS builds — no behavioral change outside strict mode.
- Ran the existing `pkg/template` unit suite with and without `GODEBUG=fips140=only`; no other regressions.

### DIFF
--- a/pkg/template/sops.go
+++ b/pkg/template/sops.go
@@ -6,6 +6,7 @@ package template
 import (
 	"bytes"
 	"context"
+	"crypto/fips140"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


joaopapereira (approved):
Lgtm

PR #1850, opened 2026-08-25
Fix: kapp-controller silently drops errors when pushing CA certs/proxy settings to its sidecarexec container

### LINKED ISSUE #1849: kapp-controller silently drops errors when pushing CA certs/proxy settings to its sidecarexec container
**What steps did you take:**
[A clear and concise description steps that can be used to reproduce the problem.]

1. Deploy kapp-controller such that the sidecarexec sidecar container in the same Pod is slightly slower to start listening on its RPC socket than the main kapp-controller container (e.g. under load, on a busy/upgrading cluster, or simply due to normal container start-order non-determinism — Kubernetes gives no ordering guarantee between containers in the same Pod).
2. Configure the `kapp-controller-config` ConfigMap/Secret with `caCerts` and/or proxy settings (`httpProxy`, `httpsProxy`, `noProxy`).
3. Let the kapp-controller Pod start. On startup, `cmd/controller/run.go` calls the config `Reconciler.Reconcile()` once synchronously, specifically to push CA certs/proxy settings to the sidecarexec container "before any tool execution happens."
4. If the sidecarexec container's RPC socket isn't listening yet when this call happens, the RPC call to `OSConfig.ApplyCACerts`/`OSConfig.ApplyProxy` fails.

**What happened:**

`pkg/config/reconciler.go`'s `Reconcile()` logs the RPC failure but swallows it (`// continue on`) and always returns `nil`. Because of this:
- The startup check in `run.go` (`if err != nil { return fmt.Errorf(...) }`) never fires, even though CA certs/proxy were never actually applied to the sidecar.
- Since the config `Reconciler`'s watches (`AttachWatches`) only re-trigger `Reconcile` on subsequent changes to the `kapp-controller-config` ConfigMap/Secret, if the config doesn't change again, CA certs/proxy are **never retried** for the lifetime of that Pod.
- Any App/PackageInstall reconciliation that depends on those CA certs (e.g. fetching from an HTTPS git/helm/imgpkg source signed by a custom CA) can then fail or hang until it hits an unrelated timeout.

**What did you expect:**

A failure to apply CA certs or proxy settings to the sidecar should be treated as a retryable error:
- The initial synchronous startup reconcile should retry (with backoff/bounded timeout) rather than silently proceeding as if configuration succeeded.
- The regular watch-triggered `Reconcile()` should return the error so controller-runtime's standard requeue-with-backoff mechanism retries it, consistent with how every other reconciler in this codebase (`pkg/app`, `pkg/packageinstall`, `pkg/pkgrepository`) handles real (non-`NotFound`) errors.

Confirmed in a real cluster's kapp-controller logs — both the one-shot startup
reconcile and the very next watch-triggered reconcile (~1s later, triggered by
the initial informer sync of the config ConfigMap/Secret) failed with the same
underlying RPC dial error, showing the sidecarexec container was not yet
listening on its socket when either reconcile attempt ran:

    {"level":"error","logger":"kc.controller.config","msg":"Failed applying CA certificates",
     "request":{"name":""},
     "error":"Internal run comm: dial unix /etc/kappctrl-mem-tmp/sidecarexec.sock: connect: no such file or directory"}

    {"level":"error","logger":"kc.controller.config","msg":"Failed applying CA certificates",
     "request":{"name":"kapp-controller-config","namespace":"kapp-controller-system"},
     "error":"Internal run comm: dial unix /etc/kappctrl-mem-tmp/sidecarexec.sock: connect: no such file or directory"}

Both errors are logged and then discarded by `Reconcile()`, and since no
further config change occurred, no other retry was ever attempted for the
rest of that Pod's lifetime.

Root cause: `pkg/config/reconciler.go`'s `Reconcile()` calls `r.osConfig.ApplyCACerts(...)` and `r.osConfig.ApplyProxy(...)`, both of which (via `pkg/sidecarexec.OSConfigClient`) are `net/rpc` calls over a Unix domain socket to the sidecarexec container. The `reconnectingRPCClient` (`pkg/sidecarexec/client.go`) does a single dial attempt with no built-in retry — if the socket doesn't exist yet, the call fails immediately. `Reconcile()` currently logs this and returns `nil` for both calls, which silently defeats the explicit intent of the synchronous startup call in `cmd/controller/run.go` ("Reconcile once synchronously to ensure controller configuration ... is applied to sidecar before any tool execution happens").

**Environment:**

- v0.60.4 (reproducible on `develop` as well)

---
Vote on this request

himsngh:

Propagate config reconciler errors so sidecar-not-ready races self-heal

Reconcile() was swallowing errors from ApplyCACerts/ApplyProxy, which are RPC calls to the sidecarexec container. If that container's socket wasn't listening yet (a real container-start race, no ordering guarantee within the Pod), the failure was logged and discarded, so CA certs/proxy could go permanently unapplied for the Pod's lifetime with no retry.

Now Reconcile() returns the error so controller-runtime requeues watch- triggered reconciles with backoff, and the one-shot startup reconcile in run.go retries with a bounded backoff instead of silently continuing.

Fixes #1849 

```release-note
NONE
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/cmd/controller/run.go
+++ b/cmd/controller/run.go
@@ -28,6 +28,7 @@ import (
 	"carvel.dev/kapp-controller/pkg/reftracker"
 	"carvel.dev/kapp-controller/pkg/sidecarexec"
 	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp" // Initialize gcp client auth plugin
 	"k8s.io/component-base/cli/flag"
@@ -181,7 +182,19 @@ func Run(opts Options, runLog logr.Logger) error {
 
 		// Reconcile once synchronously to ensure controller configuration
 		// (e.g. proxy, CA certs) is applied to sidecar before any tool execution happens.
-		_, err = reconciler.Reconcile(context.TODO(), reconcile.Request{})
+		// The sidecarexec container may still be starting up when this process
+		// runs (there is no ordering guarantee between containers in the same
+		// Pod), so retry with backoff for a bounded amount of time instead of
+		// failing on the very first transient RPC error.
+		err = wait.PollUntilContextTimeout(context.TODO(), time.Second, time.Minute, true,
+			func(ctx context.Context) (bool, error) {
+				_, err := reconciler.Reconcile(ctx, reconcile.Request{})
+				if err != nil {
+					runLog.Error(err, "Reconciling config once at startup; will retry")
+					return false, nil
+				}
+				return true, nil
+			})
 		if err != nil {
 			return fmt.Errorf("Reconcile config reconciler once: %s", err)
 		}
--- a/pkg/config/reconciler.go
+++ b/pkg/config/reconciler.go
@@ -80,13 +80,13 @@ func (r *Reconciler) Reconcile(_ context.Context, request reconcile.Request) (re
 	err = r.osConfig.ApplyCACerts(r.config.CACerts())
 	if err != nil {
 		log.Error(err, "Failed applying CA certificates")
-		// continue on
+		return reconcile.Result{}, err
 	}
 
 	err = r.osConfig.ApplyProxy(r.config.ProxyOpts())
 	if err != nil {
 		log.Error(err, "Failed applying proxy opts")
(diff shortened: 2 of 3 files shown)


aroradaman (approved):
This is not a breaking change, we just wait and retry when rpc call fails.

PR #1851, opened 2026-09-03
Fix:[1852] Add TestK8APIEndpoints and run it across k8s v1.33-v1.37 in CI

### LINKED ISSUE #1852: Add CI coverage verifying kapp-controller's API endpoints across k8s v1.33–v1.37
kapp-controller exposes several Kubernetes API surfaces:
- CRDs: `kappctrl.k14s.io/v1alpha1` (`App`), `packaging.carvel.dev/v1alpha1` (`PackageInstall`, `PackageRepository`)
- an aggregated APIService: `data.packaging.carvel.dev/v1alpha1` (`Package`, `PackageMetadata`), served by `pkg/apiserver`.

The aggregated APIService is the fragile piece: it requires kube-apiserver to establish a TLS connection to kapp-controller's Service and mark it `Available`, and that proxying behavior has broken across Kubernetes releases before (e.g. [kubernetes/kubernetes#86666](https://github.com/kubernetes/kubernetes/issues/86666)).

Today, `.github/workflows/kind-action.yml` only tests against whatever k8s version `helm/kind-action@v1`'s default node image resolves to at run time (currently v1.35.0). That version isn't pinned and there's no dedicated check that API registration/discovery actually works — a regression on a specific k8s minor version would go unnoticed until a user hits it.

Describe the solution you'd like
1. A new e2e test, `TestK8APIEndpoints`, that after deploying kapp-controller to a cluster asserts:
- `apiservice v1alpha1.data.packaging.carvel.dev` reaches condition `Available: True`
- `packages` and `packagemetadatas` are listed under `GET /apis/data.packaging.carvel.dev/v1alpha1`
- `packageinstalls.packaging.carvel.dev` and `packagerepositories.packaging.carvel.dev` are in `kubectl api-resources`
- `apps.kappctrl.k14s.io` is in `kubectl api-resources`
2. A CI job that runs only this test as a matrix across `kindest/node` images `v1.33`, `v1.34`, `v1.35`, `v1.36` & `v1.37` in parallel with the existing full e2e suite (not replacing it, to avoid multiplying the full 60-minute suite's cost by 5).

sameerforge:

Adds `TestK8APIEndpoints`, an e2e test verifying that after deploying kapp-controller:
- the aggregated APIService `v1alpha1.data.packaging.carvel.dev` becomes `Available`
- `data.packaging.carvel.dev/v1alpha1` discovery exposes `packages` and `packagemetadatas`
- `packaging.carvel.dev` discovery exposes `packageinstalls` and `packagerepositories`
- `kappctrl.k14s.io` discovery exposes `apps`

This closes a gap where kapp-controller's CRD/aggregated-API registration had no dedicated test coverage, so a regression here (e.g. the aggregated APIService failing to become `Available` on a given apiserver version) could only be caught indirectly, if at all. This test runs on every PR via the existing "Kind Cluster E2E tests" `run-tests` job, alongside the rest of the e2e suite, against the default kind node image.

Adds a new `k8s-api-endpoints.yml` workflow, on a daily schedule plus `workflow_dispatch`, that runs `TestK8APIEndpoints` across the range of k8s minor versions kapp-controller supports. Compatibility drift from new k8s releases isn't a per-diff regression concern, so this runs separately from PRs rather than adding checks to every one, following the same schedule + `workflow_dispatch` idiom as `trivy-scan.yml`/`dependency-updater.yml`. The minors tested are derived from the `k8s.io/client-go` version pinned in `go.mod` (the last 4 released minors plus the next one), so the range shifts automatically as `dependency-updater.yml` bumps `go.mod` instead of being hand-maintained. The latest patch for each minor is resolved at run time from `kubernetes/kubernetes`'s tags (via `git ls-remote`, avoiding GitHub REST API rate limits). The job then loops over the resolved versions sequentially in a single job, creating and tearing down one kind cluster per version. A failure on one version doesn't stop the others: the loop runs through every version regardless, reports each failing k8s version via `::error::` annotations, and only fails the job once all versions have been attempted.

Sample CI run for this workflow : 
https://github.com/sameerforge/kapp-controller/actions/runs/34329006415/job/102392900710

Fixes #1852

```release-note
NONE
```

- The `k8s-api-endpoints` job installs `kind/kubectl` once via `helm/kind-action@v1` (`install_only: true`, pinned to `v0.33.0` because the action's current default, `v0.31.0`, can't parse the containerd config v4 format that newer `kindest/node` images ship — this was breaking `kind load docker-image` for k8s v1.34+), builds the kapp-controller image once, then loops: `kind create cluster` → `kind load docker-image` → `kapp deploy` → run `TestK8APIEndpoints` → `kind delete cluster`, per resolved version.
- `run-tests` (full e2e suite) is untouched and keeps testing against the default/latest kind node image.

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/.github/workflows/k8s-api-endpoints.yml
+++ b/.github/workflows/k8s-api-endpoints.yml
@@ -1,0 +1,101 @@
+name: K8s API Endpoints Compatibility
+
+on:
+  schedule:
+    - cron: '0 6 * * *'
+  workflow_dispatch:
+
+jobs:
+  k8s-api-endpoints:
+    name: TestK8APIEndpoints (all k8s versions)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+      - name: Install kind + kubectl
+        uses: helm/kind-action@v1
+        with:
+          install_only: true
+          version: v0.33.0
+      - name: Install Carvel Tools
+        run: ./hack/install-deps.sh
+      - name: Resolve k8s versions kapp-controller supports
+        run: |
+          set -e -x
+          # Derive the minors to test from the client-go version already
+          # pinned in go.mod instead of hand-maintaining a version list here
+          # - it shifts automatically as dependency-updater.yml bumps go.mod.
+          # Covers the last 4 released minors plus the next (not-yet-pinned)
+          # one, wider than client-go's own +/-1 minor skew guarantee, since
+          # kapp-controller's CRD/aggregated-API surface tends to stay
+          # compatible well beyond that window in practice.
+          client_go_minor=$(grep -oE 'k8s\.io/client-go v0\.[0-9]+\.[0-9]+' go.mod | grep -oE '[0-9]+\.[0-9]+$' | cut -d. -f1)
+          minors=(v1.$((client_go_minor - 3)) v1.$((client_go_minor - 2)) v1.$((client_go_minor - 1)) v1.${client_go_minor} v1.$((client_go_minor + 1)))
+
+          # kindest/node images are cut by the kind maintainers on their own
+          # cadence, not one-for-one with every kubernetes/kubernetes patch
+          # tag (e.g. kubernetes/kubernetes had v1.33.13 with no matching
+          # kindest/node image yet), so the only accurate source for "which
+          # node images actually exist" is the registry itself.
+          url="https://hub.docker.com/v2/repositories/kindest/node/tags?page_size=100"
+          all_tags=""
+          while [ -n "$url" ] && [ "$url" != "null" ]; do
+            page=$(curl -sf "$url")
+            all_tags="$all_tags
+          $(echo "$page" | jq -r '.results[].name')"
+            url=$(echo "$page" | jq -r '.next')
+          done
+
+          versions=()
+          for m in "${minors[@]}"; do
+            latest=$(echo "$all_tags" | grep -E "^${m}\.[0-9]+$" | sort -V | tail -1)
+            if [ -z "$latest" ]; then
+              echo "::error::No kubernetes/kubernetes tag found for $m"
+              exit 1
+            fi
+            versions+=("$latest")
+          done
+          echo "Resolved versions: ${versions[*]}"
+          {
+            echo "K8S_VERSIONS<<EOF"
+            printf '%s\n' "${versions[@]}"
+            echo "EOF"
+          } >> "$GITHUB_ENV"
+      - name: Build kapp-controller image
+        run: |
+          set -e -x
+          source ./hack/version-util.sh
+          ytt -f config/config -f config/values-schema.yml -f config-dev -v dev.version="$(get_kappctrl_ver)+develop" | kbld -f- > kbld.out 2> kbldmeta.out
+          echo "KC_IMAGE=$(cat kbldmeta.out | tail -n 1 | sed 's/.*final: kapp-controller -> \(.*\)$/\1/p' | tail -n 1)" >> "$GITHUB_ENV"
+      - name: Run TestK8APIEndpoints across k8s versions
+        run: |
+          set -x
+          mkdir -p tmp
+          failed=()
+          while IFS= read -r version; do
+            [ -z "$version" ] && continue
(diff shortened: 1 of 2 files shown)


.github/workflows/kind-action.yml
```diff
@@ -42,4 +42,86 @@ jobs:
...
+      - name: Install Carvel Tools
+        run: ./hack/install-deps.sh
+      - name: Resolve latest kindest/node patch per minor
+        run: |
+          set -e -x
+          # Fixed compatibility range (major.minor); patch versions are resolved
+          # to whatever kindest/node currently publishes as latest, so this list
+          # doesn't need to be bumped by hand every time a new patch ships.
+          minors=(v1.33 v1.34 v1.35 v1.36 v1.37)
+          url="https://hub.docker.com/v2/repositories/kindest/node/tags?page_size=100"
```
himsngh: test-gh.yml#L39-L47 already resolves a Kubernetes version dynamically via the GitHub releases API. Can we resolve the per-minor latest patches the same way instead of adding a new dependency on the Docker Hub tags API?

.github/workflows/kind-action.yml
```diff
@@ -42,4 +42,86 @@ jobs:
...
           mkdir tmp
-          KAPPCTRL_E2E_NAMESPACE=kappctrl-test eval './hack/test-e2e.sh'
\ No newline at end of file
+          KAPPCTRL_E2E_NAMESPACE=kappctrl-test eval './hack/test-e2e.sh'
+
+  k8s-api-endpoints:
+    name: TestK8APIEndpoints (all k8s versions)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
himsngh: Do we really need to run this in every PR? 

 This looks more like  a compatibility-drift (triggered by new k8s releases shipping), not a per-diff regression concern, and running the full 5-version sweep on every single PR seems like the wrong trigger for it.

This repo already has the idiom for this exact situation, trivy-scan.yml and dependency-updater.yml both run on schedule + workflow_dispatch rather than per-PR. Should k8s-api-endpoints follow that same pattern instead?

.github/workflows/kind-action.yml
```diff
@@ -42,4 +42,86 @@ jobs:
...
+          version: v0.33.0
+      - name: Install Carvel Tools
+        run: ./hack/install-deps.sh
+      - name: Resolve latest kindest/node patch per minor
+        run: |
+          set -e -x
+          # Fixed compatibility range (major.minor); patch versions are resolved
+          # to whatever kindest/node currently publishes as latest, so this list
+          # doesn't need to be bumped by hand every time a new patch ships.
+          minors=(v1.33 v1.34 v1.35 v1.36 v1.37)
```
himsngh: Why do we have this hardcode? Doesn't this mean it will only always check for these minors version? 

My suggestion would be to get the versions dynamically that kapp-controller supports and run the test.

.github/workflows/kind-action.yml
sameerforge: Switched to resolving via `git ls-remote --tags` against `kubernetes/kubernetes` (same GitHub-based approach as test-gh.yml), no more Docker Hub dependency.

.github/workflows/kind-action.yml
sameerforge: Moved this out of kind-action.yml into a new schedule + workflow_dispatch workflow (k8s-api-endpoints.yml), matching trivy-scan.yml/dependency-updater.yml.

.github/workflows/kind-action.yml
sameerforge: Fixed — the minors are now derived dynamically from the `k8s.io/client-go` version in `go.mod` (last 4 released minors + the next one), so the range shifts automatically as dependencies get bumped instead of being hand-maintained.

sameerforge:
Sample CI run for this workflow : 
https://github.com/sameerforge/kapp-controller/actions/runs/34329006415/job/102392900710

.github/workflows/kind-action.yml
sameerforge: Tried that, but hit a real failure in CI: kubernetes/kubernetes had tagged v1.33.13 with no matching kindest/node image published yet, so kind's own release cadence doesn't map 1:1 to k8s source tags — the registry is the only accurate source for which node images actually exist, so I've reverted to resolving via the Docker Hub tags API.

himsngh (approved):

PR #1860, opened 2026-09-16
Fix the duplicate reference to the `secretRef` field of the `PackageInstall` object within the `App` object

klzsysy:

If the `secretRef` field exists in the `packageinstall` object, the child `app` object will redundantly reference these secrets.

image
image

Fixes #

```release-note
The `app` object will no longer duplicate the `secretRef` field found in the `packageInstall` object.
```

It also fixed an issue where duplicate references occurred in the `app` object when `package` and `packageinstall` objects referenced the same `secret`.

Added relevant unit test cases.

##### Review Checklist:

  a link to that PR
  change

```docs

```


joaopapereira (changes requested):
Some of this already has been addressed and merge in #1862, do you mind rebasing your PR?

Some other changes:
- Dedup is incomplete: annotation-driven paths (ExtHelmPathsFromSecretNameAnnKey, ytt PathsFrom / ExtYttDataValuesOverlaysAnnKey) still append without checking existing names. Either extend or document as out of scope.
- We should add a line on the Release Note section of the PR saying that we are not allowing duplicated secrets and that we are only taking into account the first occurrence of each secret
- Prefer aligning with #1862’s test style rather than duplicating fixtures.

PR #1863, opened 2026-09-22
Fix TLS authority case sensitivity and kctrl error message grammar/formatting

### LINKED ISSUE #1864: Improve case-sensitivity handling for image references and TLS authority matching
## Problem

Case-sensitivity handling in image reference comparison and TLS authority matching needs refinement to comply with RFC 1035 and prevent incorrect duplicate detection.

Currently, image references are compared entirely case-insensitive when only the registry hostname should be case-insensitive per RFC 1035. Repository paths and tags can be case-sensitive (e.g., GHCR, ECR), which means distinct references like `docker.io/Foo/Bar:v1` and `docker.io/foo/bar:v1` are incorrectly treated as duplicates.

## Changes

- Implement proper case-sensitivity handling in PackageRepository add/update operations
- Fix TLS authority matching to correctly handle mixed-case hostnames
- Improve error messages and documentation consistency

sameerforge:

DNS/registry hostnames are case-insensitive per [RFC 1035](https://www.rfc-editor.org/rfc/rfc1035), but two comparisons in this codebase treated them as case-sensitive:

- `Config.ShouldSkipTLSForAuthority` (`pkg/config/config.go`) compared the admin-configured `dangerousSkipTLSVerify` allow-list against the host parsed from an `App`/`PackageRepository`'s fetch URL using `==`. Since the allow-list and the fetch URL are typed independently by different people (a cluster admin editing the `kapp-controller-config` Secret/ConfigMap vs. an app author writing a CR), a casing difference between the two (e.g. `Registry.Example.Com` vs. `registry.example.com`) silently left `DangerousSkipTLSVerify` `false`. The fetch then attempted real TLS verification against a host the admin explicitly intended to skip, failing with an `x509: certificate signed by unknown authority` error.
- `kctrl package repository add/update` (`cli/pkg/kctrl/cmd/package/repository/add_or_update.go`) compared the `--url` flag against the already-persisted `PackageRepository`'s stored image ref with `==`. A registry host typed with different casing than what's already stored missed the no-op/idempotency short-circuit, forcing an unnecessary `Update()` call instead of just tailing status.

Both comparisons now use `strings.EqualFold` instead of `==`. Added mixed-case regression cases to `Test_ShouldSkipTLSForAuthority` covering both the bare-host and host:port branches.

Additionally, this PR includes a set of grammar and formatting fixes to `kctrl` CLI error and help messages, found via a targeted error-message audit:

- `NamespaceNameFlag.resolveValue` (`cli/pkg/kctrl/cmd/core/namespace_flags.go`): fixed the malformed error text `"Expected to non-empty namespace name"` → `"Expected a non-empty namespace name"`.
- `SecureNamespaceFlags.CheckForDisallowedSharedNamespaces` (`cli/pkg/kctrl/cmd/core/secure_namespace_flags.go`): added missing spaces between concatenated string segments, which previously ran words together (e.g. `namespace (%s)(hint: ...`).
- `kctrl completion` (`cli/pkg/kctrl/cmd/completion.go`): the command's `Short`/`Long` help text and its `default` case error only listed `bash, zsh or fish` despite `powershell` being a supported and documented (`ValidArgs`, `Example`) shell; both now list all four shells, and the error message is capitalized consistently with other `kctrl` errors.
- `PackageValuesSchemaParser.walkOnValueSchemaProperties` (`cli/pkg/kctrl/cmd/package/available/values_schema.go`): fixed a missing space in the concatenated error string (`"...has unsupported"+" type %v..."`) and capitalized the message for consistency.
- Standardized the inconsistent adjective `"non empty"` to the correct compound adjective `"non-empty"` across 8 `kctrl` command files (`app/delete.go`, `app/get.go`, `app/kick.go`, `app/pause.go`, `app/status.go`, `package/repository/kick.go`, `package/installed/{status,delete,create_or_update,get,pause_or_kick}.go`).

None of these are functional changes — they only affect the literal text of user-facing CLI error/help strings.

Fixes #1864 

```release-note
Fixed two case-sensitivity bugs where a hostname/registry reference with different casing than a previously-recorded value was incorrectly treated as a mismatch: `dangerousSkipTLSVerify` allow-list entries now match fetch URL hosts regardless of case, and `kctrl package repository add/update` now correctly recognizes a no-op update when only the registry host's casing differs. Also fixed several grammar, spacing, and consistency issues in `kctrl` CLI error and help messages.
```


cli/pkg/kctrl/cmd/package/repository/add_or_update.go
```diff
@@ -196,7 +196,10 @@ func (o *AddOrUpdateOptions) Run(args []string) error {
 		return err
 	}
 
-	if o.URL == existingRepository.Spec.Fetch.ImgpkgBundle.Image &&
+	// registry hostnames are case-insensitive per RFC 1035, and o.URL (freshly typed by the
+	// operator) and the stored Image (persisted by a possibly earlier invocation) are typed
+	// independently, so compare without regard to case
+	if strings.EqualFold(o.URL, existingRepository.Spec.Fetch.ImgpkgBundle.Image) &&
```
joaopapereira: Only the registry host is case-insensitive; org/repo path and tag can be case-sensitive (e.g. GHCR). Folding the whole string can treat distinct refs as a no-op. Prefer parse with name.ParseReference (already imported) and fold only the registry, or compare RegistryStr() case-insensitively and the remainder exactly.

joaopapereira (changes requested):
Looks good in general 1 issues that I think we should correct.
Another thing is Release note is accurate; optional: file a tracking issue (body invites that).

cli/pkg/kctrl/cmd/package/repository/add_or_update.go
sameerforge: I've updated the implementation to use name.ParseReference() and now compare only the registry case-insensitively via Context().RegistryStr(), while comparing the repository path and tag case-sensitively. This prevents false matches when refs differ only in case.

sameerforge:
@joaopapereira Thanks for the review!  I've addressed the case-sensitivity issue with `proper name.ParseReference()` parsing, comparing only the registry hostname case-insensitively while treating the repository path and tag case-sensitively. Filed tracking issue #1864 and updated the PR description. All CI checks pass — Ready for your approval now.

joaopapereira (approved):
Lgtm

